PR #105, opened 2014-12-31
Avoid Alt ambiguity on base-4.8.0.0 and above

RyanGlScott:
Since `Data.Monoid` exports an `Alt` type in `base-4.8.0.0` and above, explicitly import the `Monoid` functions to avoid creating a naming conflict with HSE's `Alt`.

### DIFF
--- a/src/HSE/FreeVars.hs
+++ b/src/HSE/FreeVars.hs
@@ -3,7 +3,7 @@
 module HSE.FreeVars(FreeVars, freeVars, vars, varss, pvars, declBind) where
 
 import HSE.Type
-import Data.Monoid
+import Data.Monoid (Monoid(..))
 import qualified Data.Set as Set
 import Data.Set(Set)
 


ndmitchell:
Thanks for the patch. Have you actually built with GHC 7.10 (an RC thereof)? I know Monoid is mostly exported by the Prelude so that may have an impact.

RyanGlScott:
I built it on GHC 7.10-rc1. The problem wasn't that `Prelude` exports `Monoid`, but rather that `HSE.FreeVars` imports all of `Data.Monoid`.

ndmitchell:
I observe cpphs has a base version too restrictive so my automatic tests fail to install it with GHC HEAD. Pinging @malcolwallace.

ndmitchell:
@malcolmwallace (see above)

ndmitchell:
Thanks for the patch. I will make a new version once my dependencies are all suitably relaxed.

malcolmwallace:
Puzzled.  cpphs has dependency base >3 && <6, so version 4.8.0.0 should be acceptable.

malcolmwallace:
I think it turns out that, because polyparse-1.10 does not build with ghc-7.10.x due to the rearrangement of Functor/Applicative/Monad classes, cabal has reverted to a version of cpphs that does not depend on polyparse (winding back from cpphs-1.18.x to 1.13), and the cabal file for cpphs-1.13 does not meet current cabal standards, so the explicit dependencies are ignored, to poor effect.

malcolmwallace:
P.S. new version of polyparse-1.11 should mean that cpphs-1.18.6 is once again selected by cabal for the build plan, and all other deps should work again.

ndmitchell:
Awesome, thanks @malcolmwallace and @RyanGlScott for fixing all this. I've made a release HLint-1.9.15 which works with GHC 7.10.

PR #115, opened 2015-02-24
fromJust shouldn't exist in polished code

phadej:
Related to https://mail.haskell.org/pipermail/libraries/2015-February/025010.html

```
% cat testi.hs 
import Data.Maybe

foo :: Maybe Int -> Int
foo x = if isJust x then fromJust x else 0

bar :: Maybe Int -> Int
bar x = 1 + fromJust x

% cabal run testi.hs 
Preprocessing executable 'hlint' for hlint-1.9.16...
Running hlint...
testi.hs:4:9: Error: Use fromMaybe
Found:
  if isJust x then fromJust x else 0
Why not:
  fromMaybe 0 x

testi.hs:4:26: Error: Use fromMaybe
Found:
  fromJust
Why not:
  fromMaybe (error "Impossible! This should never happen.")

testi.hs:7:13: Error: Use fromMaybe
Found:
  fromJust
Why not:
  fromMaybe (error "Impossible! This should never happen.")

3 suggestions
```

### DIFF
--- a/data/Default.hs
+++ b/data/Default.hs
@@ -455,6 +455,10 @@ error "Use Foldable.forM_" = (case m of Nothing -> return (); Just x -> f x) ==>
 error "Use Foldable.forM_" = when (isJust m) (f (fromJust m)) ==> Data.Foldable.forM_ m f
     where _ = noQuickCheck
 
+-- PARTIALS
+
+error "fromJust is partial" = fromJust ==> fromMaybe (error "Impossible! This should never happen.")
+
 -- EVALUATE
 
 -- TODO: These should be moved in to HSE\Evaluate.hs and applied


watashi:
The implementation of `fromJust` is `fromMaybe (error "Maybe.fromJust: Nothing")`, which I think is better than `fromMaybe (error "Impossible! This should never happen.")`.

phadej:
@watashi the behind *Impossible! this should never happen.* is that is controllable by end-programmer, he could specify the message more precisely. Generic *Maybe.fromJust: Nothing* doesn't help, as it doesn't tell who called it.

Yet @ndmitchell discussed this on mailing list already, he didn't like this PR. Maybe if we could have separate rule databases, than this would be ok-ish.

ndmitchell:
You definitely don't want to suggest they replace `fromJust` with `fromMaybe (error "bad fromJust")` since after they've made the replacement there is no call to `fromJust` at all so it's very misleading. I'll come back to the multiple rule databases sometime in the near future, with any luck.

ndmitchell:
See https://github.com/ndmitchell/hlint/issues/420 - we can now, so this is now fine by me.

Issue #127, opened 2015-05-20
Should hlint ignore filenames starting with .?


bmjames:
When searching for files recursively, HLint ignores directories beginning with `.`, but finds temporary files created by my editor, which begin with `.`. This can result in spurious errors.

I can work around this by using a glob pattern in my shell, for example:

    $ hlint **/*.hs

When expanding the glob pattern, the shell finds all the real source files that HLint would by default, but ignores the temporary files.

However, sometimes one might want to invoke HLint without a shell that has the capability to expand a glob pattern.

Do you agree that HLint should ignore them by default?

ndmitchell:
Agreed, sounds reasonable. Patch welcome :)

ndmitchell:
Merged the patch, so now fixed.

PR #128, opened 2015-05-20
Avoid filenames beginning with dot

bmjames:
As discussed in #127


ndmitchell:
Thanks, looks good. It will be in the next release, but if you'd like a release immediately (in case that makes it easier for you) let me know.

PR #131, opened 2015-05-29
Complete line rules for (=<<) as dual of (>>=)

watashi:
`Dafault.hs` contains many lint rules for `(>>=)` but not for `(=<<)`, this commit completes the missing rules for `(=<<)`.

```
watashi % cat /tmp/T.hs                                                                                                                              git dual_bind ~/gao/hlint
module T where

h, h' :: IO Int -> Int Bool
h  x = x >>= return . (>3)
h' x = return . (>3) =<< x

watashi % hlint /tmp/T.hs                                                                                                                            git dual_bind ~/gao/hlint
/tmp/T.hs:4:8: Warning: Use liftM
Found:
  x >>= return . (> 3)
Why not:
  Control.Monad.liftM (> 3) x

1 suggestion

watashi % dist/build/hlint/hlint -d data /tmp/T.hs                                                                                               git dual_bind ~/gao/hlint
/tmp/T.hs:4:8: Warning: Use liftM
Found:
  x >>= return . (> 3)
Why not:
  Control.Monad.liftM (> 3) x

/tmp/T.hs:5:8: Warning: Use liftM
Found:
  return . (> 3) =<< x
Why not:
  Control.Monad.liftM (> 3) x

2 suggestions
```

### DIFF
--- a/data/Default.hs
+++ b/data/Default.hs
@@ -290,8 +290,11 @@ warn = fmap f $ x ==> f Control.Applicative.<$> x
 -- MONAD
 
 error "Monad law, left identity" = return a >>= f ==> f a where _ = noQuickCheck
+error "Monad law, left identity" = f =<< return a ==> f a where _ = noQuickCheck
 error "Monad law, right identity" = m >>= return ==> m where _ = noQuickCheck
+error "Monad law, right identity" = return =<< m ==> m where _ = noQuickCheck
 warn  = m >>= return . f ==> Control.Monad.liftM f m where _ = noQuickCheck -- cannot be fmap, because is in Functor not Monad
+warn  = return . f =<< m ==> Control.Monad.liftM f m where _ = noQuickCheck
 error = (if x then y else return ()) ==> Control.Monad.when x $ _noParen_ y where _ = not (isAtom y) && noQuickCheck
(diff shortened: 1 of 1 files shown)


ndmitchell:
Awesome, thanks a lot! I'll push these out with the next release of HLint, unless you particularly want a release with these in sooner.

watashi:
That's fine. We actually override the rule by `return . f =<< m ===> f <$> m` in customized hline/ignore file.

In there any plan for hlint to suggest `fmap` or `<$>` instead of `liftM` in future? As ghc 7.10 is released, `Functor` is a supperclass of `Monad` now according to Functor-Applicative-Monad Proposal.

Thanks a lot.

ndmitchell:
Yes, I'd be happy to do that now (specifically a ticket for me to do in future, or a patch from someone else doing it now). Taking a quick look, where I have patterns that match on `liftM` they should probably match on both `fmap` and `<$>` as well as `liftM`. For the `liftM` fusion rule, there should probably be `fmap` and `<$>` versions which stick to producing the same. Where I suggest something it should never be `liftM`. The only debate is if to suggest `liftM ==> fmap`, and I think my inclination is that's now a good idea too - I certainly never use `liftM`.

ndmitchell:
Let's take any further `liftM` related stuff to #132. The more I think about it, the more I like it - let's kill `liftM`.

PR #146, opened 2015-07-19
Necessary adjustments to automatically apply suggestions.

mpickering:
I will squash this once everything settles down but I think everything is nearly ready to merge. 

When you get the chance could you please review the changes and suggest any improvements you would like to see.

There was one part where I couldn't properly specify the refactoring. The unification algorithm does some rearranging in the middle to match `f . g $ x` as `f (g x)` but in doing this, it is possible to variables to unify with expressions which don't exist in the AST (namely `g x`). I tried a few things to solve this but couldn't get anything to stick so if you have any ideas about how to either move this rearrangement out of the unification step or another way to solve this problem it would be appreciated. 

Apart from that, I thought that the integration was quite clean for the most part. There are some additional places where some locations needed to be threaded around properly but most refactorings were easily specifiable with only a few primitives. The application library is also only about 700 LOC.


ndmitchell:
Looking through, it certainly seems highly plausible. My thoughts:

* Can you please merge in the head, since I broke the Travis tests (my fault) and that means the buildbot fails, now fixed in head.
* I support GHC 7.4, which means you can't use MultiwayIf - should be pretty easy to get rid of it.
* There's quite a lot of whitespace changes in there in a few files that would be otherwise unchanged - I've cherry picked them to make future reviewing easier. 
* Do the tests still pass? If so, it looks entirely plausible - although given the large amount of changes, I'll need to review the individual pieces more carefully.
* You have added NamedFieldPuns in a bunch of places that also have RecordWildCards - that's redundant.
* There are a few commented out fields:

    +--    , ideaPreciseSpan :: SrcSpan -- ^ The source code that needs to be replaced
    +--    , ideaSubst :: Maybe [(String, SrcSpan)] -- ^ The performed substitution
    +--    , ideaTemplate :: Maybe String -- ^ The template used to generate the output

Are they coming back? Are they important?
* SrcSpan seems quite memory high, I raised https://github.com/mpickering/refact/issues/1 to think about that (not a serious issue as far as HLint is concerned, but might be for other things).
* Can you add at least one or two tests of the flag, so I don't immediately break it without looking :). It doesn't need to be comprehensive, just something that hits a few cases. After its merged we can work towards more test coverage, or refactoring to make the changes less separate and thus requiring less testing.

ndmitchell:
"I will squash this once everything settles down" - do you mean squash into a single patch? Please don't feel the need to do that on my account. Having a true and accurate account of what happened, what you were thinking, and how your thoughts changed, is valuable. I realise it makes bisecting slightly more fiddly, but for something as simple as HLint that's not such a big deal. I'd rather a true history than one that is pretty but fake.

mpickering:
That's fine by me if you're happy with my quite ropey commit messages. Bisecting will be very difficult because of how the `refact` library has been changed alongside it so I don't expect any  older commits to compile at all.

Yes, the tests still pass. Most of the failures I had to fix were from `Hint.Match` which I had to modify to retain positional information when doing the unification.  

Please can you restart the travis build, the dependencies on `refact` were too tight which caused the older builds to fail. 

I can get to adding some tests tomorrow. Thank you for you suggestion about the representation. I'm not very familiar with these kinds of optimisations, it seems like a good idea.

ndmitchell:
Ropey commit messages are fine. Travis restarts itself whenever you update master, it's going now.

ndmitchell:
Oh, i see what you meant about Travis - restarted now.

ndmitchell:
OK, I suggest we think about the representation for SrcSpan (you might decide to do nothing, but if it does change, easier now than later) and once that's figured out, I do a final review, and then we merge.

mpickering:
I made the changes you suggested. If you would like any more changes made then please say!

ndmitchell:
Are you happy to merge as is? If so, I'll go through in detail with a view to merging today or tomorrow.

mpickering:
I just found a bug which needs to be fixed. There is also probably some refactoring I forgot about which I will go over tonight.

ndmitchell:
OK, great, let me know when it's done.

mpickering:
The issue was https://github.com/haskell-suite/haskell-src-exts/pull/242 . Funnily enough this patch also breaks hlint as the template files have declarations such as:

```
error = ...
error "label" = ...
```

`haskell-src-exts` rightly complains that there is an arity mismatch.

mpickering:
I had a quick look over, there are two parts which I think could definitely be better but unsure what to do with them.

1. https://github.com/ndmitchell/hlint/pull/146/files#diff-171b235db8767216ad3a8774aa18b250R83
2. https://github.com/ndmitchell/hlint/pull/146/files#diff-1072dd84743480ad2cfef6318282c4e1R118

It seems like for the first one it might be better to add a new method to `Brackets` which returns the type of the expression.. but this kind of reflection is what is already going on more directly.

The second case is a major case of copy & paste, I needed the functionality of descendBracket but to apply the changes to the template rather than the actual expression it was checking bracketing for. I'm sure there's a way to merge the two.

ndmitchell:
Neither of those look fatal - I'd be happy to have them fixed up afterwards (or when someone is next refactoring the code anyway). Are you happy for me to merge as is? If so, I'll review and merge tonight.

mpickering:
Yes I think everything is good.

ndmitchell:
OK, all merged. What's the next steps? Docs to describe how to put these things together?

mpickering:
At the moment `apply-refact` provides an executable called `refactor` which calls hlint as a subprocess with the appropriate flags. Hopefully once 7.10.2 gets released this week you can have a play with it and see if you think I need to add anything else.

PR #149, opened 2015-07-31
Integrate a --refactor option

mpickering:
Is this like what you had in mind? 

Using a temporary file rather than dealing with stdin later means that you can use a filter in vim to process the file to apply refactorings. This plays much nicer in my brief experience writing vimscript but I'm not sure it's a good idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,14 @@ The first suggestion is marked as an error, because using `concatMap` in prefere
 
 **Bug reports:** The suggested replacement should be equivalent - please report all incorrect suggestions not mentioned as known limitations.
 
+### Automatically Applying Hints
+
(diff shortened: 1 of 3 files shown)


mpickering:
Ah, I see now the non-zero exit's are when there are suggestions to apply.

ndmitchell:
I don't understand why you are writing stdin to a temp file. How can vim work with a temp file you delete before giving back to them?

mpickering:
I arranged it like this because I found it nicer in vim to work with filters which send the current buffer to stdin and  replace it with what comes back on stdout. I wrote stdin to a temp file so that apply-refact could inspect what was on stdin (and the behaviour is uniform whether the input came from a file or stdin). This way apply-refact reads the temp file and produces output on stdout (which vim picks up) before hlint cleans up the temporary file.

ndmitchell:
Ok, now I understand. I think I'd push that file creation down to just the refact path since temp files are always possibly less robust. Can you also use the temp file operations from the extra library since that's already an hlint dependency. With that, it looks good.

mpickering:
I just pushed the suggested changes. If you want to try it out then building hlint from this patch and apply-refact from [here](https://github.com/mpickering/apply-refact) should work well together.

ndmitchell:
Thanks, all looks good - I've merged.

PR #152, opened 2015-08-07
Propagate exit code from child process.

mpickering:
The previous change broke all my tooling integration!

### DIFF
--- a/src/HLint.hs
+++ b/src/HLint.hs
@@ -156,8 +156,8 @@ runHints cmd@CmdMain{..} flags = do
               -- writeFile "hlint.refact"
               let hints =  show $ map (\i -> (show i, ideaRefactoring i)) showideas
               runRefactoring path file hints cmdRefactorOptions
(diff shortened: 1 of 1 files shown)


ndmitchell:
Could we have runRefactoring return an ExitCode, then runHints actually exitWith? Having all the exit stuff in runHints means I have a chance of finding it in the future when I wonder "hmm, why is that not getting to the next line". Other than that, looks good.

mpickering:
Yep no problem.

ndmitchell:
Sorry, I hadn't spotted you had updated the branch! All merged now.

mpickering:
I have written some bindings for emacs, vim and atom. Do you think I should put a link in the README?

https://github.com/mpickering/hlint-refactor-mode
https://github.com/mpickering/hlint-refactor-atom
https://github.com/mpickering/hlint-refactor-vim

ndmitchell:
Yes, sounds great.

PR #155, opened 2015-08-12
Adapt to new haskell-src-exts API.

eigengrau:
The IVar namespace argument has moved to IAbs as of HSE b0527bf.

### DIFF
--- a/src/HSE/Scope.hs
+++ b/src/HSE/Scope.hs
@@ -97,8 +97,8 @@ possImport i (UnQual _ x) = not (importQualified i) && maybe True f (importSpecs
             where ms = map g xs
 
         g :: ImportSpec S -> Maybe Bool -- does this import cover the name x
-        g (IVar _ _ y) = Just $ x =~= y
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks for the patch. Were these the only changes required to get it working with the latest HSE? Generally the diff between versions is much larger.

Given the HSE release is coming out very soon, I might wait on this patch until the release is out and then merge it then, so the HEAD continues to work with the last HSE release and so that I can make a quick upgrade to the new release.

eigengrau:
Seems reasonable; apply when it seems prudent. I just had to patch this in because I needed the support for pattern synonym’s recently added in HSE’s head (6640237). Apart from this, the build against 6640237 went through fine. Maybe there are still some merges ahead for HSE before release.

HSE’s `VarA` has been refactored into a new constructor, `AppA`, and the `KindBang` constructor has been removed, but hlint doesn’t seem to be using either of these.

ndmitchell:
Closing because the commit is also a part of #184, thanks for the patch!

PR #170, opened 2015-08-30
Add hints 'not . (a ==) ==> (a /=)' etc

bergmark:


ndmitchell:
Good idea! I've implemented something more general, inspired by your patch. HLint already generalises `f (g x)` hints into `f . g` hints, so now I apply that same logic to `not (x == y)` hints, which automatically gives all the 4 hints you added. I've kept your tests, and they all pass with that change, and it probably also catches a few other hints.

bergmark:
Cool, thanks!

PR #177, opened 2015-10-18
Make --refactor-options=-s work

drvink:


ndmitchell:
There seem to be two separate changes here, both of which look entirely reasonable (in future, I'd gently prefer two unrelated changes in two commits, but I'm not super fussed). You are switching to using execution not via a shell, and separately making `-s` go through as an interactive process. The first seems clearly a good idea. The second relies on detecting `-s` as refactor detects it, which might be a bit fragile (I guess refactor also has a long name for it?). I wonder if there is some sequence of magic that works for both - what stops `runInteractiveRefactoring` working in non-interactive refactorings? If the only drawback is the the temp file, then I'd be inclined to say use the same pattern for both.

CC @mpickering who wrote this code.

ndmitchell:
Oh, and thanks for the commit, appreciated! And don't worry about the failing appveyor test - I'm still trying to figure out how to set up appveyor.

drvink:
You are correct about there being two (unrelated) changes--apologies for stuffing them in the same commit. And yes, it does appear that there's also a long option as well. I wasn't sure whether it would be preferable to avoid the temp file indirection if we could determine that the user asked for an interactive refactor, since if we always used the temp file, then we wouldn't have to care, but it does add (probably unnecessary) complexity, so I'll change it to just always use the temp file.

ndmitchell:
No worries at all about putting two things in one commit, I appreciate all changes.

For the `-s` thing and long arguments, I think it's probably even worse than that - I imagine `-vs` works to mean `-v -s` for short flags without arguments. Agreed that staying clear of temp files is generally desirable, but here, I don't think we have much choice to be robust.

mpickering:
It did used to use a temp file. I changed it in this commit - https://github.com/ndmitchell/hlint/commit/ce01fdb31a23326e269c77f95c4f38d224b0cd3a

drvink:
Oh, right, option packing and whatnot...yes, trying to parse the options forwarded to `refactor' would be a waste of time. Good point.

drvink:
OK, I split the unrelated changes into two commits and changed the `refactor` exec stuff as suggested.

ndmitchell:
Looks good to me. @mpickering, any final objections before I merge?

mpickering:
Nope, go ahead.

ndmitchell:
Thanks for the contribution @drvink!

PR #184, opened 2015-11-18
haskell-src-exts 1.17

mpickering:


ndmitchell:
Thanks! Both @eigengrau and @mpickering. I'll be making a release tonight.

ndmitchell:
Tomorrow morning as it turns out :)

ndmitchell:
Now released as hlint-1.9.23 - thanks for your help both of you!

PR #200, opened 2015-12-27
Lint self

cblp:
The cobbler always wears the worst shoes.

Fix problems found in own code.

### DIFF
--- a/src/HSE/All.hs
+++ b/src/HSE/All.hs
@@ -73,11 +73,11 @@ parseModuleEx flags file str = do
             ParseOk (x, cs) -> return $ Right (applyFixity fixity x, cs)
             ParseFailed sl msg -> do
                 -- figure out the best line number to grab context from, by reparsing
-                flags <- return $ parseFlagsNoLocations flags
-                ppstr2 <- runCpp (cppFlags flags) file str
-                pe <- return $ case parseFileContentsWithMode (mode flags) ppstr2 of
-                    ParseFailed sl2 _ -> context (srcLine sl2) ppstr2
-                    _ -> context (srcLine sl) ppstr
+                let flags' = parseFlagsNoLocations flags
+                ppstr2 <- runCpp (cppFlags flags') file str
+                let pe = case parseFileContentsWithMode (mode flags') ppstr2 of
+                        ParseFailed sl2 _ -> context (srcLine sl2) ppstr2
+                        _ -> context (srcLine sl) ppstr
                 Control.Exception.evaluate $ length pe -- if we fail to parse, we may be keeping the file handle alive
                 return $ Left $ ParseError sl msg pe
     where
--- a/src/Hint/Structure.hs
+++ b/src/Hint/Structure.hs
@@ -39,6 +39,7 @@ foo = let ~(x:xs) = y in z
 module Hint.Structure(structureHint) where
 
 import Hint.Type
+import Data.Function (on)
 import Data.List.Extra
 import Data.Tuple
 import Data.Maybe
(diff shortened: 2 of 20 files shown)


src/HSE/All.hs
```diff
@@ -73,11 +73,11 @@ parseModuleEx flags file str = do
             ParseOk (x, cs) -> return $ Right (applyFixity fixity x, cs)
             ParseFailed sl msg -> do
                 -- figure out the best line number to grab context from, by reparsing
-                flags <- return $ parseFlagsNoLocations flags
-                ppstr2 <- runCpp (cppFlags flags) file str
-                pe <- return $ case parseFileContentsWithMode (mode flags) ppstr2 of
-                    ParseFailed sl2 _ -> context (srcLine sl2) ppstr2
-                    _ -> context (srcLine sl) ppstr
+                let flags' = parseFlagsNoLocations flags
```
ndmitchell: I think this is a perfectly legitimate use of `return`, so I'll probably modify hlint not to suggest `let` in case of variable shadowing.

src/Hint/Structure.hs
```diff
@@ -39,6 +39,7 @@ foo = let ~(x:xs) = y in z
 module Hint.Structure(structureHint) where
 
 import Hint.Type
+import Data.Function (on)
```
ndmitchell: I tend not to like explicit import lists, they involve lots of import editing, especially for things where the module is in base.

src/Test/All.hs
```diff
@@ -8,6 +9,10 @@ import Data.List
 import System.Directory
 import System.FilePath
 
+#if __GLASGOW_HASKELL__ < 710
+import Data.Functor ((<$>))
+#endif
```
ndmitchell: Instead of this just `import Data.Functor` and on the next line do `import Prelude`. No CPP and it's warning free on all GHC versions.

ndmitchell:
Thanks for the patch - some of them were quite nasty, much nicer now.

I don't have time to merge now, so I've gone through and put line notes in. If you want to reroll the patch with those changes, feel free, but otherwise tomorrow when I have a bit more time I'll merge and do them myself - either works for me.

After doing this, I'll make sure I add a check to HLint that it doesn't have any HLint warnings - being warning free for HLint itself does seem reasonable and desirable.

ndmitchell:
All applied now, and the notes fixed. I've now also required zero unwarned errors, so it should stay lint free from now on.

src/Hint/Structure.hs
cblp: Implicit import lists decrease readability, is many cases you can't figure out where does a variable come from.

Implicit import lists may cause name conflicts if some library has added new name. Adding a new name is considered a non-breaking change in Hackage PVP and in Semver.

So, implicit imports are a bad style.

src/Test/All.hs
cblp: Hm, sounds like magic.

cblp: The code like `__GLASGOW_HASKELL__ < 710` is self-descriptive and will be dropped when you quit supporting previous versions.

`import Data.Functor; import Prelude` looks strange and will be polluting the code after versions < 7.10 support is dropped.

ndmitchell: Yes, both versions suck, but my opinion is that the version with CPP sucks more - with CPP there is code that is out of the reach of static checkers, lint systems etc, and that adds significant overhead running the preprocessor (on certain systems it can be on the order of 1s). I suspect the processor test would have to persist for ~7 years, so that's quite a long way in the future, but when I do drop GHC 7.8 support I'll probably go through and clean it up (or hopefully write a lint check to do so for everyone).

And I entirely agree, the way around it is total magic.

src/Hint/Structure.hs
ndmitchell: I disagree. IDE's are the thing that should resolve variables, not humans, and Hoogle usually gets it most of the time. For particularly obscure names I do sometimes use explicit imports, or where I am importing some functionality which seems to be disjoint from the rest of the module. But for things like `on`, most people should probably learn what they do and where they come from.

Adding a new name is considered a non-breaking change in PVP because PVP doesn't think these things through - I consider that a bug in PVP (or the whole concept flawed...). It's still a fairly rare instance though.

And the benefit of implicit imports is that I can code without continually adjusting the import list, which is a form of administration that distracts from the real job of coding. For me, the benefit in faster coding is worth the disadvantages you list. I appreciate a lot of people disagree with me here.

ndmitchell:
Since a few of these are down to my coding preferences, I have documented them here: https://github.com/ndmitchell/neil/blob/master/README.md#quirks - I intend to put some kind of link to that document somewhere at some point once I've finished it.

PR #202, opened 2016-01-06
Include refactorings is --json output.

xich:
This was useful for integrating hlint/refactor with arcanist, but seems generally useful to have.

### DIFF
--- a/src/Idea.hs
+++ b/src/Idea.hs
@@ -20,7 +20,7 @@ data Idea = Idea
     ,ideaFrom :: String -- ^ The contents of the source code the idea relates to.
     ,ideaTo :: Maybe String -- ^ The suggested replacement, or 'Nothing' for no replacement (e.g. on parse errors).
     ,ideaNote :: [Note] -- ^ Notes about the effect of applying the replacement.
-    , ideaRefactoring :: [Refactoring R.SrcSpan] -- ^ How to perform this idea
+    ,ideaRefactoring :: [Refactoring R.SrcSpan] -- ^ How to perform this idea
     }
     deriving (Eq,Ord)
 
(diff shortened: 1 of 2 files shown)


ndmitchell:
CC @mpickering.

Tests are failing because there are some sample outputs in https://github.com/ndmitchell/hlint/blob/master/tests/json.test, e.g. https://travis-ci.org/ndmitchell/hlint/jobs/100507478 - but that's easy to fix up.

My main question is if we should be trying to JSON encode the refactor type as well. I don't really want Refactor to depend on a JSON library (excessive dependencies), which might end up with everyone decoding the JSON themselves, and Read/Show is the "built-in" serialisation and protects me from API changes in Refactoring. I guess Read/Show is the best route forward, but I'd appreciate confirmation from @mpickering on that.

ndmitchell:
BTW, thanks for the patch, and agreed, it does seem sensible.

xich:
Yeah, I explicitly did not JSON encode the refactoring type because I don't want to parse/manipulate it elsewhere. I just wanted the blob that comes out of Show so I can pass it into the refactor command-line tool later if I want. (I'm building a web UI with an 'apply this' button, essentially.)

xich:
Also, I'll fix up the tests in the morning and update my pull request.

ndmitchell:
Sounds like an awesome project :) - as long as you've thought about Read/Show I'm happy with that decision. I'll hit the merge button once the tests are sorted out.

ndmitchell:
Thanks, all applied. Do you want a new release with this feature in it now, or should I not worry and just wait for the next release to come around?

xich:
Feel free to wait for the next release, no rush. Thanks for merging!

PR #204, opened 2016-01-07
Add error = concatMap id ==> concat

bkaestner:
Similarly to `map id ==> id`, `concatMap id` is just a `concat` in disguise.


ndmitchell:
Thanks! Merged and will be in the next release.

PR #205, opened 2016-01-09
Add hints for compare x y == EQ, compare x y /= EQ

bkaestner:
I don't recall to have seen `compare x y == EQ` in any code, but since all other `==`/`/=` and `Ordering` combinations are already there, I don't think it will cause any harm.


ndmitchell:
Thanks, good idea!

PR #229, opened 2016-03-17
Add hints for not using void with mapM_ and forM_

bergmark:


ndmitchell:
Thanks!

ndmitchell:
I note a related hint is `void (mapM f xs) ==> mapM_ f xs` - and it even has a significant performance boost too. @bergmark  - interested in submitting such a hint? If not, I probably will get round to it.

bergmark:
Sure, I'll do it later today!

PR #230, opened 2016-03-18
Add more hints for not using void with mapM/mapM_/forM/forM_

bergmark:


ndmitchell:
Thanks! For the future, note that there's no requirement to explicitly give the eta reduced version - HLint infers that itself. I've now mentioned that in the readme at https://github.com/ndmitchell/hlint/blob/master/README.md#adding-hints.

PR #232, opened 2016-03-25
Suggest <|> from Alternative instead of mplus from MonadPlus

watashi:
`Maybe` has instances for both `Alternative` and `MonadPlus`. And since AMP, `Alternative` has been a super class of `MonadPlus`. Let's suggest the more general `<|>` from `Alternative` instead of `mplus` from `MonadPlus`.


ndmitchell:
Makes sense, thanks a lot!

PR #234, opened 2016-03-29
Add variant of x `seq` x: "join seq"

anka-213:
I've seen this being used and causing problems.


ndmitchell:
Thanks! A very weird construct, and one I'd never have imagined, but certainly wrong and if it's causing problems in practice then one for the list.

Issue #160, opened 2015-08-18
Improve the documentation of the configuration file


ndmitchell:
From https://code.google.com/p/ndmitchell/issues/detail?id=577:

Could there be a configuration file .hlintrc for specifying warnings to ignore (or additional warnings to enforce)? For example, JSHint uses .jshintrc to style different JavaScript projects.

http://jshint.com

@ndmitchell replied:

You can already create a file HLint.hs with configuration in, including which rules to ignore, for example, create HLint.hs in the current directory containing:

    ignore "Eta reduce"

What were you hoping for in this bug?

1) Better documentation of that feature
2) The ability to also load in from ~
3) More standard Unix file name

User replied better documentation and a sample in the README would both be handy.

lwm:
I am working on some patches for [commercialhaskell/stack](https://github.com/commercialhaskell/stack) (see https://github.com/commercialhaskell/stack/pull/2606) and am trying to isolate a single type of suggestion to make the patches small.

In the case of #2606, I only wanted to report and refactor the 'Redundant bracket' suggestion. In this case, I had to use the `--ignore=Foo` for all the different type of errors I encountered. As you can imagine, I ended up with something like:

```
  hlint \
  --ignore="Use fmap" \
  --ignore="Redundant do" \
  --ignore="Reduce duplication" \
  ..
```

Is there some existing way to say 'only report this type of suggestion' to hlint right now? If not, how do you imagine this to be possible in the `Hlint.hs` file or possibly in a `.hlintrc` as mentioned in https://github.com/ndmitchell/hlint/issues/223.

ndmitchell:
There is not currently any way to do that - having `--only` seems quite reasonable. For the `.hlint` file I've replied in #223.

ndmitchell:
BTW, at the moment I currently do `--report`, which lets you filter by hint type, and then you can see just the ones which are an issue. I appreciate that often driving it from the command line is better.

PR #271, opened 2016-10-02
Add `--only` flag

lwm:
Something from my stash. Something for #160.

Related to https://github.com/haskellpeople/hlint/issues/12.


ndmitchell:
Thanks!

PR #272, opened 2016-10-02
Documents how to extend HLint by adding inline tests, resolved #260

### LINKED ISSUE #260: Document how to develop with tests
@tomjaguarpaw in #257 had no idea where to put the tests, or how to run them. A sad consequence of it not being guessable in any way. I should document it in the README.

daniel-j-h:
First try at some docs for how inline tests work. The behavior I wrote down here is what I figured out by poking HLint and some inline tests but more importantly reading `Annotations.hs`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@ HLint is a tool for suggesting possible improvements to Haskell code. These sugg
 * [Installing and running HLint](#installing-and-running-hlint)
 * [FAQ](#faq)
 * [Customizing the hints](#customizing-the-hints)
+* [Extending the HLint Code Base](#extending-the-hlint-code-base)
 
 ### Acknowledgements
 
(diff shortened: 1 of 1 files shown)


lwm:
Oh, I think we're missing a link to the TOC at the top?

ndmitchell:
-- is for comments, but very few tests use it, so no need to even mention it. <COMMENT> is the escaped version of a comment, so the test is literally working on comments. It's exceptionally rare so can be ignored for the docs.

Other than that, looks good

daniel-j-h:
Added TOC. @ndmitchell I think Github swallowed your second sequence of chars: care to elaborate? Okay with the `zipFrom` example?

ndmitchell:
-- is for comments, but very few tests use it, so no need to even mention it. [COMMENT] is the escaped version of a comment, so the test is literally working on comments. It's exceptionally rare so can be ignored for the docs.

Where I've used square brackets instead of angles. ZipFrom is fine as an example.

ndmitchell:
Thanks for the docs. Sorry it took so long to merge - just drifted off my radar a bit too much.

PR #275, opened 2016-10-05
Add Gitter badge

lwm:
Consider this the follow up from the event in Berlin ;) Thanks for making time for us! I hope to do some more events, so maybe it is useful to keep this chat room for anyone interested in contributing and needing some help. Sorry for whitespace noise, editor did it.


lwm:
Ah, a failing test case. I won't inline those badges then ;)

PR #276, opened 2016-10-14
Remove redundant nested parens

xich:
You can currently remove all the nested redundant parens with multiple
runs of hlint --refactor, but why not do them all at once?

### DIFF
--- a/src/Hint/Bracket.hs
+++ b/src/Hint/Bracket.hs
@@ -9,7 +9,7 @@ no = f (x x)
 yes = (foo) -- foo
 yes = (foo bar) -- @Suggestion foo bar
 yes = foo (bar) -- @Warning bar
-yes = foo ((x x)) -- @Warning (x x)
+yes = foo ((x x)) -- @Suggestion (x x)
 yes = (f x) ||| y -- @Suggestion f x ||| y
 yes = if (f x) then y else z -- @Suggestion if f x then y else z
 yes = if x then (f y) else z -- @Suggestion if x then f y else z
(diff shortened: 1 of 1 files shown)


src/Hint/Bracket.hs
```diff
@@ -102,9 +108,9 @@ bracket bad = f Nothing
 
         -- f (Maybe (index, parent, gen)) child
         f :: (Data (a S), Uniplate (a S), ExactP a, Pretty (a S), Brackets (a S)) => Maybe (Int,a S,a S -> a S) -> a S -> [Idea]
-        f Just{} o@(remParen -> Just x) | isAtom x = bracketError msg o x : g x
-        f Nothing o@(remParen -> Just x) | bad || isAtom x = (if isAtom x then bracketError else bracketWarning) msg o x : g x
-        f (Just (i,o,gen)) v@(remParen -> Just x) | not $ needBracket i o x =
+        f Just{} o@(remParens -> Just x) | isAtom x = bracketError msg o x : g x
+        f Nothing o@(remParens -> Just x) | bad || isAtom x = (if isAtom x then bracketError else bracketWarning) msg o x : g x
```
xich: This isAtom check caused the test on line 12 to flip from Warning to Suggestion. (Removing both sets of parens gives isAtom == false. Before we only removed one so isAtom was true.)

This seemed harmless, and I didn't really understand this logic, so I just changed the test.

ndmitchell:
Thanks! The logic was that if you put brackets around something that is already atomic that's just a terrible idea, whereas `(x y) z` is slightly more normal than `(x) y`. But whether when you are suggesting both in one step you could argue it either way.

Issue #279, opened 2016-11-10
Suggest replacing 'data' with 'newtype' when possible


23Skidoo:
When you have a definition like

```
data Foo = Foo Int
```

it's usually a good idea to replace it with

```
newtype Foo = Foo Int
```

even though the semantics is not 100% identical. I was surprised to find out that HLint doesn't already suggest this.

ndmitchell:
Good idea!

ndmitchell:
Done in #282.

23Skidoo:
Awesome, thanks!

PR #282, opened 2016-11-21
Suggest replacing 'data' with 'newtype' when possible

niteria:
This implements the idea from #279.

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -99,6 +99,7 @@ library
         Hint.Match
         Hint.Monad
         Hint.Naming
+        Hint.NewType
         Hint.Pattern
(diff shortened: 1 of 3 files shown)


ndmitchell:
This looks good. I'll take a proper look shortly.

ndmitchell:
Thanks! All merged. I changed it from warning to suggestion, and gave a note warning the user that it changes strictness properties, but minor tweaks - nice test cases and a good catch on existentials.

niteria:
Thanks!
To give credit where it's due, this is @DiegoNolan's patch, I'm only upstreaming it.

ndmitchell:
Thanks to both of you!

PR #308, opened 2017-02-26
Updated newtype hint to account for unboxed types

rahulmutt:
The program

```haskell
{-# LANGUAGE MagicHash #-}
import GHC.Prim

data X = X Int#
```

will give the suggestion that you should declare `X` as a newtype, which is incorrect. While the style of definition above is very rare in Haskell, it's quite common in Eta when declaring wrappers for Java objects so in bindings to Java libraries, you'll be flooded with the incorrect hint above.

This change assumes that all types ending in `#` are unboxed, which is a reasonable assumption make.

### DIFF
--- a/src/Hint/NewType.hs
+++ b/src/Hint/NewType.hs
@@ -20,6 +20,7 @@ data A = A {b :: !C} -- newtype A = A {b :: C}
 module Hint.NewType (newtypeHint) where
 
 import Hint.Type
+import Util
 
 newtypeHint :: DeclHint
 newtypeHint _ _ = newtypeHintDecl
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks for the patch! I actually tried your example and it still warned for `Int#` - my guess is you were missing the case for `TyCon` directly not wrapped in `TyApp`. I refactored the code to use a few utilities I already have in HLint, moved the HSE utility to HSE.Util (that's where I keep all my HSE utils) and did a few other tweaks - but it's pretty much following your logic.

rahulmutt:
Thanks! I was not aware of HSE.Util - I'll keep that in mind for future PRs. The refactored code looks very clean and nice.

PR #322, opened 2017-04-08
Fix default.yaml examples

cydparser:
The custom error example resulted in: 
```haskell
YamlParseException {yamlProblem = "did not find expected ',' or '}'", yamlContext = "while parsing a flow mapping", ...
```


ndmitchell:
Thanks! I raised #325 to do something smarter so I can auto-validate it.

Issue #340, opened 2017-05-14
»arguments« configuration setting is broken


quchen:
Having the `arguments` flag present with non-empty content in `.hlint.yaml` produces an error,
```
echo "main = id (pure ())" > Main.hs
echo "- arguments: [--color]" > .hlint.yaml
hlint .
```
```
hlint: .: openFile: inappropriate type (is a directory)
```

The issue seems independent of the flag in the list; `foo` instead of `--color` does not change the behaviour. Removing `.hlint.yaml` or emptying its contents produces the expected result,

```
./Main.hs:1:8: Warning: Evaluate
Found:
  id (pure ())
Why not:
  (pure ())

1 hint
```

ndmitchell:
Thanks for the report - interesting indeed. I'll investigate.

sergv:
It seems to me that the issue is caused by the following line: https://github.com/ndmitchell/hlint/blob/master/src/HLint.hs#L150. If yaml file contains `arguments` declaration, the `readAllSettings` function will try to parse command with extended list of arguments and return new command instead of old one that was specified to `runHints`. The issue is that old command has files already resolved, while newly parsed command has no such thing.

Consider my local example:

The old command than gets into `runHints`
```haskell
CmdMain
  { cmdFiles =
      [ "./Setup.hs"
      , "./doc-svg-gen/Main.hs"
      , "./doc-svg-gen/Setup.hs"
      , "./src/Language/Haskell/Brittany.hs"
      , "./src/Language/Haskell/Brittany/Backend.hs"
      , "./src/Language/Haskell/Brittany/BackendUtils.hs"
      , "./src/Language/Haskell/Brittany/Config.hs"
      , "./src/Language/Haskell/Brittany/ExactPrintUtils.hs"
      , "./src/Language/Haskell/Brittany/LayouterBasics.hs"
      , "./src/Language/Haskell/Brittany/Prelude.hs"
      , "./src/Language/Haskell/Brittany/Types.hs"
      , "./src/Language/Haskell/Brittany/Utils.hs"
      , "./src/Language/Haskell/Brittany/Config/Types.hs"
      , "./src/Language/Haskell/Brittany/Layouters/Decl.hs"
      , "./src/Language/Haskell/Brittany/Layouters/Expr.hs"
      , "./src/Language/Haskell/Brittany/Layouters/Pattern.hs"
      , "./src/Language/Haskell/Brittany/Layouters/Stmt.hs"
[...]

PR #341, opened 2017-05-15
Fix issue #340 - reading command line arguments from yaml config

sergv:
Fix for #340


ndmitchell:
FWIW I half fixed this issue on a train journey before seeing your answer. Your version is way nicer than mine! More efficient, better abstractions, nicer styling, all in all very nice. Many thanks!

PR #358, opened 2017-07-10
Warn on unused MagicHash (fixes #357)

### LINKED ISSUE #357: Warn on unused MagicHash
Should be easy to detect

mitchellwrosen:
Here you go sir, let me know if I've tested this sufficiently. Cheers!


ndmitchell:
Thanks! Does magic hash allow # just as a suffix? Or in the middle? Is a#b allowed?

mitchellwrosen:
Yup, just at the end. All identifiers, including type class names (!). Yes, `a#b` is allowed... though it means `a # b` ;)

ndmitchell:
Pull request looks textbook perfect - thanks a lot - merged with gratitude!

Issue #338, opened 2017-05-08
Anything imported hidden should be used


ndmitchell:
If you `import Foo hiding (bar)` then `bar` should be used in the document. Check that GHC doesn't give an error on this first.

ndmitchell:
Things not covered:

- `import Foo hiding (Bar)`
- `import Foo hiding (Bar(bar))`
- `import Foo hiding (Bar(Bar))`
- `import Foo hiding (pattern Bar)`
- `import Foo hiding (Bar(type Baz)`

ndmitchell:
See #404 - it seems this hint doesn't really work, which is a shame, but hey ho.

joneshf:
FWIW, I've found this hint to be incredibly useful. I wasn't bit by any of the cases that cause it to fail, but I'm curious if those cases could have been addressed with an `ANN` disabling the hint for the module.

If an `ANN` would work, would it be possible to have this hint again in a future version or is it too late?

ndmitchell:
@joneshf I really found this incredibly useful too. I really like HLint to by default (in at least 99% of cases), so I'm happy to turn this hint back on (and would love to do so), but it has to have a lot less false positives - not just recommending people use `ANN`. Currently the best idea I have is that if you have import hiding (foo), and foo doesn't occur anywhere in the document, then the hiding is redundant. It doesn't catch the record wildcards+warning case, but perhaps that's crossing the 99% threshold? Alternative ideas welcome.

CC @mitchellwrosen

mitchellwrosen:
The original code I wrote attempted to, given some manually hidden identifier, find all of the _uses_ of it. If there are none, complain that there's no reason to hide it.

However, as discovered in #404, there are more corner-cases than the original ones we thought of, which were basically just module re-exports:

- (1) RecordWildCards uses identifiers in `..`, which we can't know.
- (2) `main` is used in the `Main` module
- (3) Probably more :)

Apparently _shadowing_ an identifier was _not_ considered a "use". As in,

```haskell
import Foo hiding (blah)
blah = 5
```

The "unused warning" hint would fire here, complaining that there's no point in hiding `blah`, because it's not actually being used.

Perhaps that was wrong - if we _don't_ complain that `blah` is needlessly hidden in this example, that would cover both of the false positives that triggered `(1)` and `(2)` above.

joneshf:
@ndmitchell That makes sense. Thanks for the insight.

[...]

Issue #330, opened 2017-04-18
Warn about unused As imports


ndmitchell:
If someone does an `import Foo as Bar`, and the qualified name `Bar` is never used, then warn.

mitchellwrosen:
Just to clarify, this is not meant to catch code like the following, correct?

```haskell
import Foo as Bar
import Woo as Bar
x = Bar.y -- comes from Foo, not Woo
```

ndmitchell:
That seems impossible to detect with the information HLint knows.

PR #360, opened 2017-07-10
Find unused hidden expressions (addresses part of #338)

mitchellwrosen:
I took a stab at #338, but stopped only after identifying unused expressions, as they seem much simpler to spot than the other cases. For example, in `import Foo hiding (Bar)`, `Bar` could be a data constructor, type class, type synonym, type family, or perhaps something I'm forgetting.

Also, unfortunately wildcard syntax (`import Foo hiding Bar(..)`) prevents this particular hint as we don't know what that `..` is!

Anywho, this seemed like a good enough place to start. Please let me know if I've provided enough testing - cheers!

### DIFF
--- a/src/Hint/Import.hs
+++ b/src/Hint/Import.hs
@@ -33,6 +33,73 @@ import qualified List -- import qualified Data.List as List
 import Char(foo) -- import Data.Char(foo)
 import IO(foo)
 import IO as X -- import System.IO as X; import System.IO.Error as X; import Control.Exception  as X (bracket,bracket_)
+import A hiding (a) -- import A
+import A hiding (a, b); foo = a -- import A hiding (a)
+import A hiding (a, b); foo = A.a -- import A hiding (a)
+import A as B hiding (a) -- import A as B
+import A as B hiding (a, b); foo = a -- import A as B hiding (a)
+import A as B hiding (a, b); foo = B.a -- import A as B hiding (a)
+import qualified A hiding (a) -- import qualified A
+import qualified A hiding (a, b); foo = A.a -- import qualified A hiding (a)
+import qualified A as B hiding (a, b); foo = B.a -- import qualified A as B hiding (a)
+import A hiding ((+)) -- import A
+import A hiding ((+), (*)); foo = (+) -- import A hiding ((+))
+import A hiding ((+), (*)); foo = (+x) -- import A hiding ((+))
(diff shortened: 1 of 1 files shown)


ndmitchell:
Nice, this looks great. Once the tests pass I'll merge.

ndmitchell:
Btw, I wondered if you'd seen https://github.com/ndmitchell/hlint/issues/330 - looks like the same data you collected could enable that hint quite easily.

mitchellwrosen:
Oops, can you please hold off on mergin as I totally forgot to write this kind of test:

```
module Foo (show) where
import Prelude hiding (show) -- show is not used in code, but hiding it still matters because it's exported
show = ()
```

And no, I hadn't seen #330 but it is quite similar! Should be pretty easy to squash.

ndmitchell:
I note that this tripped up HLint of HLint:

```
src\HSE\Type.hs:11:1: Suggestion: Unnecessary hiding
Found:
  import Language.Haskell.Exts as HSE hiding (parse, loc, paren)
Why not:
  import Language.Haskell.Exts as HSE
```

As you note, a weird case we hadn't thought of.

mitchellwrosen:
Oof. That hint is wrong, since `HSE` is re-exported, and some other module that imports `HSE.Type` is not expecting `(parse, loc, paren)` to come with it.

mitchellwrosen:
This is almost ready, but I forgot about operators.

mitchellwrosen:
Ok, I think this is ready for review

ndmitchell:
Looks awesome - very thoroughly tested! Anything remaining for #338 to be closed? To me it looks like there's little we'd be able to do beyond this.

mitchellwrosen:
Yah - #338 is still valid! This patch only looks for unnecessary `hiding` expressions, not an unused `as` import.

ndmitchell:
#330 is tracking unused `as`, so #338 was just the `hiding`, which it looks like you've covered very thoroughly. I've closed #338.

mitchellwrosen:
Oops! I got my issues confused. This patch actually doesn't cover all of #338 because none of these are detected:

- `import Foo hiding (Bar)`
- `import Foo hiding (Bar(bar))`
- `import Foo hiding (Bar(Bar))`
- `import Foo hiding (pattern Bar)`
- `import Foo hiding (Bar(type Baz)`

etc... however, those could be made into new issues :)

ndmitchell:
Thanks, I put those notes in #338 and reopened it

PR #363, opened 2017-07-11
Add --git flag (fixes #337)

### LINKED ISSUE #337: Add --git flag to check files in git
Probably something like `git ls-files --cached --others --exclude-standard *.hs *.lhs`, and then run `hlint` on those.

mitchellwrosen:
Here you go! I wasn't exactly sure how to test this feature, but please let me know if you want an accompanying test. Thanks! #337


ndmitchell:
This code is way cleaner than I would have written - very nice and most appreciated!

PR #365, opened 2017-07-14
"only" -> "additionally"

mitchellwrosen:
Minor nit, but `--git` currently _appends_ all tracked files to the (possibly empty) files and directories manually passed to hlint. Should it instead just ignore them?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -2,7 +2,7 @@ Changelog for HLint
 
     #352, suggest maybe for fromMaybe d (f <$> x)
     #338, warn about things imported hidden but not used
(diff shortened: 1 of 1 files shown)


mitchellwrosen:
And now that I think about it, perhaps we should nub this list of files!

ndmitchell:
If you have passed files by hand, it seems wrong for `--git` to ignore them. I'm generally thinking `nub` might be the way to go?

ndmitchell:
Actually, `nub` on files is pretty nasty (slashes, casing, symlinks) and in practice people are unlikely to do `--git` and explicit files, so we can just not worry.

PR #376, opened 2017-08-16
New hint: Use fmap

strake:
Found:
```
  do x <- m_x
     return (f x)
```
Why not:
```
  do f <$> m_x
```

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -10,7 +10,7 @@
 - extensions:
   - default: false
   - name: [DeriveDataTypeable, GeneralizedNewtypeDeriving, NoMonomorphismRestriction, OverloadedStrings]
-  - name: [PatternGuards, Rank2Types, RecordWildCards, ScopedTypeVariables, ViewPatterns, TupleSections]
+  - name: [PatternGuards, Rank2Types, RecordWildCards, ScopedTypeVariables, ViewPatterns, TupleSections, LambdaCase]
   - name: [ExistentialQuantification, MultiParamTypeClasses]
   - name: [FlexibleContexts, FlexibleInstances]
   - {name: CPP, within: HsColour} # so it can be disabled to avoid GPL code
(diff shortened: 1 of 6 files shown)


ndmitchell:
Apologies that it's taken so long to get round to reviewing your code, it looks sensible, and the changes it caused in the hlint code base certainly make sense. My only concern is that in some cases it might cause too much code to get merged together - I'm wondering if there are some restrictions we could make on things like the size of `f` or the information content in `m_x`?

strake:
If we do so, we ought to make the limits configurable (including infinity). An alternative is to make it suggest rather than warn. I do like warning in at least some cases — what are your thoughts?

strake:
ping

ndmitchell:
I'm not a fan of configurable limits, partly because there aren't any in HLint so far, and partly because as soon as you have dials then the "search space" people have to explore to find configurations that suit them becomes huge.

After much (too much) deliberation I've merged it as is, and we can think about restrictions in the future if that seems useful.

ndmitchell:
Thanks for the patch, and sorry for the delay!

ndmitchell:
I've just pushed some patches dampening down this hint, as looking through my projects, it was causing unexpected and somewhat ugly hints. In particular, there were quite a few places I got things like:

```haskell
src\Debug.hs:120:37: Warning: Use <$>
Found:
  do newDecs <- traverse (appsFromDec tag) decs
     newExp <- appsFromExp tag exp
     return $ LetE newDecs newExp
Why not:
  do newDecs <- traverse (appsFromDec tag) decs
     LetE newDecs <$> appsFromExp tag exp
```

Here the move to using `<$>` makes things a lot less clear. As a first approximation, I've made it only apply when there are exactly two things in the `do` block. My general approach is when I find HLint has gone too far to dial it back to obvious cases, then slowly expand if it still seems good.

strake:
I've actually been working on a lint which would propose using `<*>` here:
```haskell
src\Debug.hs:120:37: Warning: Use <*>
Found:
  do newDecs <- traverse (appsFromDec tag) decs
     newExp <- appsFromExp tag exp
     return $ LetE newDecs newExp
Why not:
  do LetE <$> traverse (appsFromDec tag) decs <*> appsFromExp tag exp
```

PR #377, opened 2017-08-20
New hint: Use λ-case

strake:
Found:
```
  \ x ->
    case x of
        A -> a
        B -> b
```
Why not:
```
  \case
      A -> a
      B -> b
```

### DIFF
--- a/src/Hint/Lambda.hs
+++ b/src/Hint/Lambda.hs
@@ -67,6 +67,7 @@ yes = foo (\x -> Just x) -- @Warning Just
 foo = bar (\x -> (x `f`)) -- f
 baz = bar (\x -> (x +)) -- (+)
 yes = blah (\ x -> case x of A -> a; B -> b) -- \ case A -> a; B -> b
+no = blah (\ x -> case x of A -> a x; B -> b x)
 </TEST>
 -}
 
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks for these pull requests. I'll take the time to go through them, hopefully later in the week!

strake:
ping?

ndmitchell:
Pong. Afraid I'm snowed under, but it's on my pile very definitely!

ndmitchell:
Use of lambda case is not universally accepted yet - it's still a relatively new extension - but it is popular. I wonder if this hint should be off/ignored by default? Or maybe now it's easy enough to ignore hints that isn't a big deal?

src/Hint/Lambda.hs
```diff
@@ -142,6 +143,7 @@ lambdaExp p o@(Lambda _ pats x) | isLambda (fromParen x), null (universeBi pats
       munge ident p@(PWildCard _) = p
       munge ident p = PVar (ann p) (Ident (ann p) [ident])
       subts = ("body", toSS body) : zipWith (\x y -> ([x],y)) ['a'..'z'] (map toSS pats)
+lambdaExp p o@(Lambda _ [view -> PVar_ u] (Case _ (view -> Var_ v) alts)) | u == v = [suggestN "Use λ-case" o $ LCase an alts]
```
ndmitchell: This goes wrong if u is reused in one of the alts. That case needs guarding for.

ndmitchell: Can we stick to ascii both in the source and console messages, so lambda-case.

strake:
I'd like this on by default — λ-case was sorely wanted in Haskell — but it's ultimately your call.

strake:
ping

ndmitchell:
Sorry for the delay, I had a lot of other stuff going on... After some contemplation, I've decided you're right, merged with thanks!

PR #381, opened 2017-08-25
Delegate to hse-util package

pepeiborra:
The only objectionable bit in this PR (mostly) is the addition of an orphan Default instance for SrcSpanInfo in HSE.Util; the alternative would be to newtype it.


src/HSE/Util.hs
```diff
@@ -296,6 +297,8 @@ an = toSrcInfo nullSrcLoc [] nullSrcLoc
 dropAnn :: Functor f => f s -> f ()
 dropAnn = void
 
+instance Default SrcSpanInfo where
+  def = an
```
ndmitchell: Any reason this instance shouldn't live in haskell-src-exts-util?

ndmitchell: Also, where is the `def` value used? Does one of the HSE-util things have it as required context?

pepeiborra: It is used by the rebracketing code to insert `Paren` constructors. 

You are probably right that the instance could live in HSE-util. I never know with orphan instances, but in this case there is only one sensible default value and it would make HSE-util immediately easier to use.

ndmitchell: Yep, I think that's the thing that should be pushed up to haskell-src-exts first - it's obvious useful. Having it in HSE-util just makes the transition easier.

ndmitchell:
The Appveyor failure is because HSE-util isn't yet in nightly - that's not an issue. The Travis failure is because it won't build in GHC 7.8 or below, because of a missing `import Data.Monoid` - see https://travis-ci.org/ndmitchell/hlint/jobs/268372812. Any thoughts about making this work further back? I'm not too fussed, but following the same GHC version tolerance as HSE itself seems like a reasonable default.

pepeiborra:
I see no inconvenient with adding the missing module import and making HSE-util more backward compatible. Let’s see how it goes after I upload a new version of HSE-util and amend this PR.

ndmitchell:
As a tip, if you have unused import warnings, you should do `import Data.Monoid` first then `import Prelude` after. That way you are free of import warnings across all compilers - it's what I do in HLint.

pepeiborra:
Done. I've pushed a new version of HSE-util with the orphan instance and the missing import. It will be in Stackage nightly soon as well, as my PR to build-constraints was merged this morning.

ndmitchell:
Seems like `unknown flag in  {-# OPTIONS_GHC #-} pragma: -Wno-orphans` happens with GHC 7.10 and below.

ndmitchell:
I imagine `-fno-orphans` or similar would be compatible further back.

ndmitchell:
Thanks!

PR #383, opened 2017-08-26
Allow hlint to function as a CC engine

pbrisbin:
- Add `--json-cc` flag to output in in accordance with the CC SPEC[1].
- Add `cc/` directory with the machinery required to run as a codeclimate
  engine. The only bit that couldn't be isolated in this directory is
   the .dockerignore which must live in the directory you're building
   from. 
- Add a .codeclimate.yml so we can test the engine on ourselves.

[1]: https://github.com/codeclimate/spec/blob/master/SPEC.md

This should replace pbrisbin/codeclimate-hlint and address the requested features around configurability. This version of the engine runs the hlint binary normally, so passing additional flags, accessing project-local configuration files, etc, should all Just Work.

Try it out:

    cc/build
    codeclimate analyze

See https://github.com/codeclimate/codeclimate#installation

This is meant as a jumping-off point: please let me know if there's a better home for `CC.hs`, a better name for the `--json-cc` flag, or a better place to put the `cc/` Docker machinery.

One other question: I notice `stack.yml` is git-ignored. That file will have to exist for the docker image to build, and code climate maintainers will be doing the image build out of a fresh checkout, so I'm not sure if we should commit one, or put one in `cc/` that we can use as part of the `cc/build` process?

### DIFF
--- a/cc/build
+++ b/cc/build
@@ -1,0 +1,6 @@
+#!/bin/sh
+set -e
+
+docker build \
+  --file cc/Dockerfile \
+  --tag codeclimate/codeclimate-hlint .
--- a/cc/engine.hs
+++ b/cc/engine.hs
@@ -63,7 +63,8 @@ main = do
     c <- readConfig "/config.json"
 
     callProcess "hlint" $
-        [ "--datadir", "/opt/hlint"
+        [ "lint"
+        , "--datadir", "/opt/hlint"
         , "--json-cc"
         , "--no-exit-code"
         ]
--- a/cc/engine.hs
+++ b/cc/engine.hs
@@ -1,0 +1,40 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Main where
+
+import Data.Aeson
+import Data.List (isSuffixOf)
+import System.Process (callProcess)
+import System.IO (hPutStrLn, stderr)
+
+import qualified Data.ByteString.Lazy.Char8 as C8
+
+data Config = Config
+    { cIncludePaths :: [FilePath]
+    , cExtraFlags :: [String]
+    }
+
+instance FromJSON Config where
+    parseJSON = withObject "Config" $ \o -> Config
+        <$> o .: "include_paths"
+        <*> pure [] -- TODO: optional config.flags
+
+main :: IO ()
+main = do
+    ec <- eitherDecode <$> C8.readFile "/config.json"
+    options <- case ec of
+        Left ex -> do
+            hPutStrLn stderr $ "Ignoring invalid /config.json: " ++ ex
+            return ["."]
+        Right c -> return $ cExtraFlags c ++ filter include (cIncludePaths c)
+
+    callProcess "hlint" $
+        ["--datadir"
+        , "/opt/hlint"
+        , "--json-cc"
+        , "--no-exit-code"
+        ] ++ options
+
+  where
+    -- Return true for directories as well, so they are included
+    include :: FilePath -> Bool
+    include p = any (`isSuffixOf` p) ["/", ".hs", ".lhs"]
--- a/cc/engine.hs
+++ b/cc/engine.hs
@@ -15,6 +15,42 @@ data Config = Config
     }
 
 instance FromJSON Config where
+    -- Given a source tree:
+    --
+    -- > ./
+    -- >   a.hs
+    -- >   b.hs
+    -- >   c/
+    -- >     d.hs
+    --
+    -- And .codeclimate.yml:
+    --
+    -- > engines:
+    -- >   hlint:
+    -- >     enabled: true
+    -- >     config:
+    -- >       flags:
+    -- >         - --foo
+    -- >         - --bar
+    -- >
+    -- > exclude_paths:
+    -- >   - b.hs
+    --
+    -- We will find a /config.json like:
+    --
+    -- > {
+    -- >   "include_paths": [
+    -- >     "a.hs",
+    -- >     "c/"
+    -- >   ],
+    -- >   "config": {
+    -- >     "flags": [
+    -- >       "--foo",
+    -- >       "--bar",
+    -- >     ]
+    -- >   }
+    -- > }
+    --
     parseJSON = withObject "Config" $ \o -> Config
         <$> o .: "include_paths"
         <*> do
--- a/cc/engine.hs
+++ b/cc/engine.hs
@@ -32,7 +32,7 @@ main = do
         Right c -> return $ cExtraFlags c ++ filter include (cIncludePaths c)
 
     callProcess "hlint" $
-        ["--datadir"
+        [ "--datadir"
         , "/opt/hlint"
         , "--json-cc"
         , "--no-exit-code"
--- a/cc/engine.hs
+++ b/cc/engine.hs
@@ -3,6 +3,7 @@ module Main where
 
 import Data.Aeson
 import Data.List (isSuffixOf)
+import Data.Maybe (fromMaybe)
 import System.Process (callProcess)
 import System.IO (hPutStrLn, stderr)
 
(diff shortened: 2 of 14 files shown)


ndmitchell:
Thanks for the code! I'll review properly in the next few days.

For `stack.yaml` I already test that `stack init` generates a working `stack.yaml`, which is why I don't have one in the directory. Could the code climate scripts just call `stack init`? I'm generally not a fan of putting dates in a repo that need incrementing on some schedule.

pbrisbin:
> Could the code climate scripts just call stack init?

That's fine with me, except it doesn't seem to work:

```
Resolver 'lts-9.2' does not have all the packages to match your requirements.
    haskell-src-exts-util not found
        - hlint requires >=0.2.1.2
```

Everything is fine if I make a `stack.yaml` that specifies the `extra-deps`.

AFAICT, this affects master hlint overall, not just my attempts at building.

pbrisbin:
Oh by the way, if you'd prefer, it'd be reasonable to reduce this PR to only the `--json-cc` addition. The machinery under `cc/` could rename in my `codeclimate-hlint` project. Please let me know if that'd be a better first step.

ndmitchell:
Given currently a stack is required, we should add it - at least for now. I will aim to remove it in future though. I'm happy to review all at once.

cc/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
ndmitchell: Is build the convention? I tend to prefer build.sh so people know it's a shell script. However, I'm not super fussed.

cc/engine.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Main where
```
ndmitchell: Please add an explicit export list of 'main'

cc/engine.hs
```diff
@@ -0,0 +1,83 @@
...
+import Data.Maybe (fromMaybe)
+import System.Process (callProcess)
+import System.IO (hPutStrLn, stderr)
+
+import qualified Data.ByteString.Lazy.Char8 as C8
+
+data Config = Config
+    { cIncludePaths :: [FilePath]
+    , cExtraFlags :: [String]
+    }
```
ndmitchell: Is this a config format we are defining? Any reason not to reuse something that already exists? Or is this a code climate format?

cc/engine.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE OverloadedStrings #-}
```
ndmitchell: I'd prefer Engine with a leading capital

hlint.cabal
```diff
@@ -126,3 +133,19 @@ executable hlint
     ghc-options:        -rtsopts
     if flag(threaded)
         ghc-options:    -threaded
+
+executable cc-engine
```
ndmitchell: Could we roll this executable into Hlint - eg have "Hlint cc" do the necessary magic?

ndmitchell: Assuming we do need a separate executable, would it make more sense to have cc/hlint-cc.cabal? I don't see that many benefits to keeping it in the main one.

src/CC.hs
```diff
@@ -0,0 +1,138 @@
...
+        , T.pack to
+        , "```"
+        ]
+
+    categories _ = ["Style"]
+
+    points Ignore = 0
+    points Suggestion = basePoints
+    points Warning = 5 * basePoints
+    points Error = 10 * basePoints
```
ndmitchell: It seems like these points are markers of difficulty rather than importance - so are these numbers sensible?

src/CC.hs
```diff
@@ -0,0 +1,138 @@
...
+printIssue :: Issue -> IO ()
+printIssue = C8.putStrLn . (<> "\0") . encode
+
+-- | Convert an hlint @'Idea'@ to a datatype more easily serialized for CC
+fromIdea :: Idea -> Issue
+fromIdea Idea{..} = Issue
+    { issueType = "issue"
+    , issueCheckName = "HLint/" <> T.pack (camelize ideaHint)
+    , issueDescription = T.pack ideaHint
+    , issueContent = content ideaFrom ideaTo
```
ndmitchell: You are dropping ideaNote which should definitely end up in the content - those are warnings about when refactorings have subtle side conditions.

src/CC.hs
```diff
@@ -0,0 +1,138 @@
...
+import qualified Data.Text as T
+import qualified Data.ByteString.Lazy.Char8 as C8
+
+import Idea (Idea(..), Severity(..))
+
+data Issue = Issue
+    { issueType :: Text
+    , issueCheckName :: Text
+    , issueDescription :: Text
+    , issueContent :: Text -- TODO: Markdown type?
```
ndmitchell: I don't think a markdown type gives enough benefits here.

src/CC.hs
```diff
@@ -0,0 +1,138 @@
...
+    , issueCategories :: [Text]
+    , issueLocation :: Location
+    , issueRemediationPoints :: Int
+    }
+
+data Location = Location FilePath Position Position
```
ndmitchell: I wonder if defining srcSpanJSON makes sense, so you avoid the orphan instance and avoid a separate location type. Would avoid multiple conversions and let you put SrcSpan in Issue?

src/CmdLine.hs
```diff
@@ -174,6 +175,7 @@ mode = cmdArgsMode $ modes
         ,cmdCppSimple = nam_ "cpp-simple" &= help "Use a simple CPP (strip # lines)"
         ,cmdCppAnsi = nam_ "cpp-ansi" &= help "Use CPP in ANSI compatibility mode"
         ,cmdJson = nam_ "json" &= help "Display hint data as JSON"
+        ,cmdJsonCC = nam_ "json-cc" &= help "Display hint data as Code Climate Issues JSON"
```
ndmitchell: Should the flag just be --code-climate? The important thing is it's in code climate format. The fact code climate format is JSON is just a detail.

cc/build
pbrisbin: I'm happy to name it whatever you'd like. My personally style is to never use extensions with executables. Callers shouldn't know or care, and we should be free to change, the implementation language without changing the name. Would you like me to rename it `build.sh`?

cc/engine.hs
pbrisbin: This is (our materialization of) the defined `/config.json` format we operate with as a CC Engine. I tried to explain how it works a little bit in the comment down with the `ToJSON` instance.

hlint.cabal
pbrisbin: I thought it was cleaner to keep a separate executable, and only build it on a flag. IMO, users (and developers) of `hlint` directly shouldn't really be exposed to or have to think about the CC cruft daty-to-day. Had I thought of it, I would've done `cc/engine.cabal`, so I'm in favor of that.

That said, if you prefer `hlint cc` just let me know, I see now reason that wouldn't work.

src/CC.hs
pbrisbin: I would love you input on that. The comment down with `basePoints` describes how it's meant to work, using that could you recommend rough numbers for the various severities?

src/CC.hs
pbrisbin: I didn't realize that; could you offer a rough template of what content should look like, taking that into account?

src/CC.hs
pbrisbin: You're suggesting making a `ToJSON SrcSpan` and using it directly here? To me it feels wrong to make *the* JSON instance for anything conform CC's definition, so I'd lean towards a separate type anyway (even if just a `newtype` to allow a specific instance). Additionally, `SrcSpan -> Location` also handles relativizing the file-path (albeit naively at the moment). Taken together, these make the `Location` type wortwhile IMO.

Those are my reasons, but again it's totally your call if you'd like to do something else.

src/CmdLine.hs
pbrisbin: `--code-climate` works for me. `--cc` would also be OK with me, I'm not sure if there's risk in the brevity, WDYT?

pbrisbin:
@ndmitchell just a friendly bump on this. Any other changes you'd like to see?

ndmitchell:
Merged! Sorry it took so long - life went crazy :)

pbrisbin:
No worries, thank you! I'll touch base with the CC folks next week and get this new version into their system.

ndmitchell:
I just released 2.0.10, so you have a released version to build upon.

Issue #393, opened 2017-10-21
Suggest "fmap f m" from "m >>= pure . f" and "pure . f =<< m"


contivero:
Hlint currently suggests that when using `return` instead of `pure`, but there is no equivalent hint for `pure`. I personally prefer `f <$> m`, but since `fmap` is what's currently being suggested, I stuck with that.

ndmitchell:
Makes a lot of sense. FWIW, recommended fmap since it was in the Prelude but <$> wasn't. Now that is no longer the case I'd welcome a PR to suggest <$> in return and pure cases.

PR #394, opened 2017-10-22
#393, Suggest f <$> m for m >>= pure . f and similar expressions

contivero:
Adds the pure equivalents of the m >>= return . f ==> fmap f m
hints, and changes the 4 of them to use <$> instead of fmap.


contivero:
My bad, tried `stack test` instead of `stack exec hlint test` (why are the tests part of the executable btw?), so I missed adding / updating the tests.

I'm editing the `<TEST> </TEST>` part, and the `serialise.test` file which are the ones failing.

ndmitchell:
Thanks! I generally find the Cabal test mechanism to be pretty poor - it requires additional compiles, poor module sharing, and isn't very powerful. That said, exposing the tests through a test suite as well is probably not a terrible idea so it works as people expect.

PR #397, opened 2017-10-22
Should close #395, or at least the part agreed on.

### LINKED ISSUE #395: Suggest x $> y for x *> pure y, and similar expressions.
The suggestions would be:

| LHS          |   RHS |
| ----------------- |:----------:|
| `x *> pure y`    | `x $> y` |
| `x >> return y` | `x $> y` |
| `return x << y` | `x <$ y` |
| `pure x <* y`    | `x <$ y` |

Although I guess the following are also valid:

| LHS          |   RHS |
| ----------------- |:----------:|
| `x >> pure y`  | `x $> y` |
| `x *> return y` | `x $> y` |
| `return x <* y` | `x <$ y` |
| `pure x << y`  | `x <$ y` |

If you are ok with them, I'll work on a PR.

contivero:
I only added the suggestions listed in #395 for `*>` (not those for `>>`).

I took the liberty of modifying the previous line in CHANGES.txt, as the suggestion given is `f <$> m` instead of `fmap f m`. I also added the new changes to the file.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,8 @@
 Changelog for HLint
 
-    #393, suggest fmap f m from m >>= pure . f and pure . f =<
+    #395, Suggest x $> y from x *> pure y and x *> return y
+          Suggest x <$ y from pure x <* y and return x <* y
+    #393, Suggest f <$> m from m >>= pure . f and pure . f =<<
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks! As a general rule, for very simple replacements, I'm happy for you to add a test or not - either is fine.

PR #408, opened 2017-11-26
Use the same config file logic in argsSettings as in hlint

alanz:
This correctly picks up the local config file, when used in `haskell-ide-engine`.

See https://github.com/haskell/haskell-ide-engine/issues/363


ndmitchell:
Thanks! I take it a new release would be handy?

alanz:
Yes, but not urgent as hie builds with git deps all over the place still. This is just one more.

ndmitchell:
Released as 2.0.11.

PR #409, opened 2017-11-27
Didn't know that module-level annotations need to be placed after imp…

saurabhnanda:
…ort statements.


ndmitchell:
Great information to include! However, typographical conventions mean “after” should be italic not bold - happy to merge and tweak myself or get a revised PR.

saurabhnanda:
> However, typographical conventions mean “after” should be italic not bold

Done.

ndmitchell:
Thanks!

PR #419, opened 2017-12-19
show a couple of very small code examples as code

kadoban:
This improves readability of these notes, as 'a' and 'and' otherwise scan as just normal English words, making the sentences hard to read.


ndmitchell:
That is significantly clearer! Thanks

PR #423, opened 2018-01-10
New hint: Use bind

strake:
Found:
  do x <- bar
     f x
Why not:
  do bar >>= f

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -45,6 +45,7 @@ flag gpl
 
 library
     default-language:   Haskell2010
+    default-extensions: LambdaCase
     build-depends:
         base == 4.*, process, filepath, directory, containers,
         unordered-containers, yaml, vector, text, bytestring,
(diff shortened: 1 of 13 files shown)


ndmitchell:
Thanks for the patch. I find this quite a lot less readable than the alternative formulation - it feels like putting each action on a separate line is the natural thing to do, and the `do` syntax makes that clear. My inclination is to say this is one step too far and reject it.

strake:
Would you accept it if disabled by default? (Is this possible for built-in lints?) I find this lint quite useful in my work.

ndmitchell:
I'm afraid, reviewing the Monad file, it's got a bit out of hand - its just too complex, and off by default hints at the moment aren't really feasible. I think the sensible thing to do is decline this PR for now, and in the future it might become more feasible. (Perhaps by expressing these hints are more like normal HLint rules)

Issue #428, opened 2018-01-25
Lens related hlint rules.


yairchu:
After `hlint` [supports `lens` well](https://github.com/ndmitchell/hlint/issues/416), it would be nice to add add these rule:

    warn = a .~ Just b ==> a ?~ b
    warn = a & Control.Lens.mapped %~ b ==> a <&> b
    warn = a & Control.Lens.mapped . b %~ c ==> a <&> b %~ c

And probably more rules will follow too :)

mpickering:
Or I would prefer,

```
warn = a .~ Just b ==> set a (Just b)
```

and others would prefer different things. 

I don't think this domain specific hint belongs in hlint.

yairchu:
@mpickering 

There are two topics here:

* Preference of names over operators or vice versa. This is a debate where I don't expect people to reach a consensus.
* Preference of existing combinators such as the rule existing for [`concatMap`](https://github.com/ndmitchell/hlint/blob/b56acbe0c42a73e9d1bb07e423891e4f0d75cd47/data/hlint.yaml#L101)

My suggested hint is of the latter, using an existing operator instead of a more complicated way to use another operator. Not about changing from names to operators which I wouldn't suggest.

ndmitchell:
There's also a case of deciding what the "core" lens concepts are - I know what `.~` does, but had no idea on `?~`...

I'm not averse to some such rules though.

yairchu:
> I know what `.~` does, but had no idea on `?~`...

One could say it's slightly similar situation with `concatMap` - that a user writing `concat (map f xs)` may have been unaware on the existence of `concatMap` and now learns about it via `hlint`. Granted, `concatMap`'s name is pretty self explanatory, but lens also tries to have some logic behind its choice of operators too and in such a situation that you get a hint to use `?~` you actually will learn about it and understand well what it does.

ndmitchell:
Yeah, after I wrote this I did wonder if that's exactly the kind of hint HLint should have...

ndmitchell:
I'll let someone who uses lens more than I do come up with the list of suitable hints, but the three above look like plausible ones. PR's welcome.

ndmitchell:
Done by the PR in #430.

PR #430, opened 2018-01-29
#428, add first rules for lens

yairchu:
These are some rules for stuff that occurred in my own code and could probably be helpful to others.


ndmitchell:
Thanks, looks awesome!

PR #431, opened 2018-01-30
Fix typo in previous commit for #428

yairchu:
Sorry accidentally I messed up the previous PR by keeping changes in my staging area and not committing them..

So I guess perhaps the next PR should be test cases for these.


ndmitchell:
Thanks! I added one test case after the last commit to check the lens set was turned on. I don't think yet cases help in general because the definition is so direct - odds are a test case will have the same bugs anyway!

Issue #438, opened 2018-02-07
Add fixity for (Data.Monoid.<>)


yairchu:
`<>` is a popular operator that's lacking the fixity definition for it (`infixr 6 <>`). This can cause rules to misfire.

I'm not sure if the right place to add it is a PR to [haskell-src-exts](https://github.com/haskell-suite/haskell-src-exts/blob/master/src/Language/Haskell/Exts/Fixity.hs#L162) or in hlint.

yairchu:
On it :)

yairchu:
Created a PR there - https://github.com/haskell-suite/haskell-src-exts/pull/400

Got some more rules for lens but for me some of them misfired without this fixity. I guess I'll contribute those once this fixity is fixed?

ndmitchell:
Happy to take a PR adding them in a new section "base not yet in HSE" - otherwise you'll have to wait for a new HSE release, which can take a while.

ndmitchell:
Fixed by #439.

PR #439, opened 2018-02-07
#438 add missing base fixities

yairchu:


ndmitchell:
Thanks, I also added `<$` and `$>`.

PR #442, opened 2018-02-12
maybe mempty ==> foldMap

yairchu:


ndmitchell:
I'm unsure as I generally find the foldable/traversable harder to read, but I'm very on the fence, so I just Tweeted it out to try and get more opinions: https://twitter.com/ndm_haskell/status/963106790018777089

ocharles:
This is a tricky one. There are cases where sometimes it's better and sometimes it's unclear what the underlying monoid is... which can be important to know when reading code. I'm probably +0.6 on this, but there's no emoji for that.

ChShersh:
>  There are cases where sometimes it's better and sometimes it's unclear what the underlying monoid is...

I agree. `foldMap` is not really shorter than `maybe mempty`. And it's extremely clear without context what `maybe mempty` does.

yairchu:
>  sometimes it's unclear what the underlying monoid is... which can be important to know when reading code

In that case you wouldn't use `maybe mempty` but rather `maybe specializedEmpty`. This hides the foldable rather than the monoid..

yairchu:
So iiuc one important issue is whether `lens` should avoid suggesting combinators of less specific types. It might also potentially in some cases remove necessary information for the type inference.

In that case if such rules shouldn't be included, perhaps they could still be included in a separate group that's not enabled by default?

yairchu:
Is `generalise-for-conciseness` a good name? Can one even have hyphens in group names?

yairchu:
Btw I noticed that this rule:

`- warn: {lhs: case m of Nothing -> return (); Just x -> f x, rhs: Data.Foldable.forM_ m f}`

Is a "generalisation for conciseness" too. That is - in some cases after applying the suggestion code may no longer type-check (nothing would force the `Maybe` as the `Foldable`).

ndmitchell:
I don't think `generalise-for-conciseness` is a good name to use - I think the fact it's a generalisation is just one reason why people don't like it - it's also a big generalisation and a small conciseness, which is why people are concerned. That said, we have nothing better, so in it goes.

In contrast, the `forM_` are much more concise and thus can be a default-on hint.

PR #447, opened 2018-02-26
make .ghci work in unix environments

yairchu:
When trying to use ghcid within VS code on hlint, ghcid failed to work due to the hlint's `.ghci` file containing paths with Windows backslashes. Iiuc, forward slashes nowadays work in all platforms, hence this patch replaces the backslashes with normal slashes. Haven't tested it in Windows, have tested with on macOS.

### DIFF
--- a/.ghci
+++ b/.ghci
@@ -1,21 +1,21 @@
 :set -fwarn-incomplete-patterns
 :set -fwarn-unused-binds -fwarn-unused-imports -fwarn-orphans
 :set -isrc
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks! One issue is that the `:!foo` lines on Windows have to have `\` for their first component (the executable name), or the second component if that is something like `ptime`. I've also stripped the `.exe`'s which are never needed and make things slightly more portable. I've pushed that as a follow up.

If you are going to be using the commands that fire off executables we should be able to replace `:!foo\bar` with `:!cd foo && bar`, as for `cd` on Windows the directory name is allowed `/`.

Issue #434, opened 2018-02-02
Support equivalence of functions


yairchu:
Take this rules for example:

    - warn: {lhs: fmap f (fmap g x), rhs: fmap (f . g) x, name: Functor law}
    - warn: {lhs: f <$> g <$> x, rhs: f . g <$> x, name: Functor law}
    - warn: {lhs: fmap id, rhs: id, name: Functor law}
    - warn: {lhs: id <$> x, rhs: x, name: Functor law}

These are repetitions due to `fmap` and `<$>` being the same things and hlint having rules for both. An alternative solution would be to add a equivalence declarations in the imports section:

    - fmap = (<$>)
    - return = pure

And then have less rules:

    - warn: {lhs: f <$> g <$> x, rhs: f . g <$> x, name: Functor law}
    - warn: {lhs: id <$> x, rhs: x, name: Functor law}

When the rule is applied to `fmap f (fmap g x)`, `hlint` will see that `fmap` was used so it will also suggest `fmap` rather than `<$>` in its suggestion.

Cases such as `fmap f (g <$> x)` will now also be recognised. For mixed usages like these it's fine if hlint just suggests `f . g <$> x` as in the rules because there isn't any clear user preference between the two.

An extra feature on top would be to also support flipped equivalences such as:

    - (=<<) = flip (>>=)

And then rules such as:

    - warn: {lhs: return a >>= f, rhs: f a, name: "Monad law, left identity"}
    - warn: {lhs: f =<< return a, rhs: f a, name: "Monad law, left identity"}
    - warn: {lhs: m >>= return, rhs: m, name: "Monad law, right identity"}
    - warn: {lhs: return =<< m, rhs: m, name: "Monad law, right identity"}

Would be reduced to:

    - warn: {lhs: return a >>= f, rhs: f a, name: "Monad law, left identity"}
    - warn: {lhs: m >>= return, rhs: m, name: "Monad law, right identity"}

ndmitchell:
[...]

PR #451, opened 2018-03-05
add --with-group command line option

### LINKED ISSUE #444: Better `--help` for enabling extra hint groups.
The `--help` text has nothing about enabling or disabling hint groups.

The `-w` option seems to be for individual hints only.

yairchu:
Also resolves #444 

Not sure about the logic in `CmdLine.cmdHintFiles` - which decides to look in files according to existence of `--with` options. Should that also depend on `--with-group` options?


ndmitchell:
Can you add a test in `tests/`? This looks like exactly the kind of functionality I'm going to break in a refactoring without realising it!

Not sure about the `--with` vs `--with-group`, but suggest we go with what we've got for now. I think #434 might impact whatever we do here anyway...

yairchu:
Sure, now with tests! :)

ndmitchell:
Thanks!

PR #453, opened 2018-03-08
Also flag pure for Use <$>

mfine:
I often use `pure` in place of `return` in places - having the fmap rule pick those up as well would be nice. This patch is probably not the right way to handle that - a Applicative module or applicativeFmap function?


ndmitchell:
Thanks! Good spot, and good idea. I merged your patch, then built on top of it to define `isReturn` so everywhere I used to look for `return` I now accept `pure`.

mfine:
Fantastic! Extending it to the other "return" areas is great. Running against my projects it was able to pick up all these new cases. Thanks so much!

PR #455, opened 2018-03-13
add some hints for odd uses of sequence

yairchu:
these situations occur when a list is changed to have one element, such as here:
https://github.com/lamdu/lamdu/commit/27d87a6aa7569fa55607176e56127606024987c6#diff-ecf05a9fd506bf563383f04e1652a41aL258

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -379,6 +379,7 @@
     - warn: {lhs: mapM_ f (map g x), rhs: mapM_ (f . g) x}
     - warn: {lhs: mapM id, rhs: sequence}
     - warn: {lhs: mapM_ id, rhs: sequence_}
+    - warn: {lhs: sequence (return a), rhs: return <$> a}
 
     # APPLICATIVE / TRAVERSABLE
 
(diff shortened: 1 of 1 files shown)


data/hlint.yaml
```diff
@@ -379,6 +379,8 @@
     - warn: {lhs: mapM_ f (map g x), rhs: mapM_ (f . g) x}
     - warn: {lhs: mapM id, rhs: sequence}
     - warn: {lhs: mapM_ id, rhs: sequence_}
+    - hint: {lhs: "sequence [a]", rhs: "(:[]) <$> a"}
```
ndmitchell: Is the use of functor and `(:[])` really simpler? There seems a lot more "syntactic noise" in the RHS.

yairchu: It's true that it's noisy syntacticly, but it is conceptually simpler.

Imagine a case where someone had `mconcat <$> sequence [a, b]` and later edited it to `mconcat <$> sequence [a]`. After seeing the suggestion they may notice it should actually be just `a` now.

quasicomputational: The generalisation from `Applicative` to `Functor` is definitely nice and could also lead to wins elsewhere, but the RHS is definitely ugly. Maybe part of a new set of hints, for other uglifying generalisations?

yairchu: Maybe this is appropriate for the "generalise" group?

yairchu:
Moved the `sequence [a]` rules to the `generalise` group and generalised it to use `pure` rather than `(:[])`. It is more succinct but does replace the list with any `Applicative`.

The other rules generalise only in reducing constraints (i.e require `Functor` rather than `Traversable`) but that shouldn't cause any unresolved type variables so those are not in the special group.

ndmitchell:
Thanks!

PR #462, opened 2018-03-30
Ignores home directory when it isn't present.

blackgnezdo:
Reordered the imports alphabetically while at it.

### DIFF
--- a/src/CmdLine.hs
+++ b/src/CmdLine.hs
@@ -10,21 +10,22 @@ module CmdLine(
 import Control.Monad.Extra
 import Data.Char
 import Data.List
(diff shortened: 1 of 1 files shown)


blackgnezdo:
The easiest way to reproduce the failure is to run `env - hlint test` which quickly starts complaining without this patch about:
GOT : HOME: getHomeDirectory:getEnv: does not exist (no environment variable)

The fixed version goes much farther through the test suite eventually breaking with
hlint: tests/json.test: hGetContents: invalid argument (invalid byte sequence)

Still, hlint is functional with this patch when packaged up correctly even without HOME.

ndmitchell:
Thanks!

PR #477, opened 2018-05-03
Allow `unit_` prefix to bypass camelCase suggestion

league:
You were allowing `case_` and `prop_` to bypass camelCase suggestion.

But tasty-discover deprecates `case_` in favor of `unit_` for HUnit tests: https://github.com/lwm/tasty-discover#write-tests


ndmitchell:
Thanks! Good idea.

Issue #472, opened 2018-04-22
Need to write LANGUAGE pragma?


pythonissam:
Suppose we have the following code:

```haskell
dayTimeField :: Field Handler UTCTime
dayTimeField =
  Field
  { fieldParse =
      \rawVals _fileVals -> do
        currentTime <- liftIO getCurrentTime
        case map unpack rawVals of
          [[], _] -> return $ Right $ Just currentTime
          [_, []] -> return $ Right $ Just currentTime
          [day, time] -> return $ Right $ Just $ stringToUTCTime (day ++ time)
          _ -> parseHelper (\_ -> Left MsgInvalidDay) [] []
  , fieldView =
      \idAttr nameAttr otherAttrs _eResult _isReq ->
        [whamlet| $newline never
            <div>Date
            <input id=#{idAttr} name=#{nameAttr} *{otherAttrs} type=date>
            <div>Time
            <input id=#{idAttr}-confirm name=#{nameAttr} *{otherAttrs} type=time>
        |]
  , fieldEnctype = UrlEncoded
  }
```

`hlint` throws a parse error if we don't put `{-# LANGUAGE QuasiQuotes #}` at the top of this file:

```haskell
A.hs:16:24: Error: Parse error
Found:
            [whamlet| $newline never
                <div>Date
  >             <input id=#{idAttr} name=#{nameAttr} *{otherAttrs} type=date>
                <div>Time
                <input id=#{idAttr}-confirm name=#{nameAttr} *{otherAttrs} type=time>


1 hint
```

Even if we specify `QuasiQuotes` to `default-extensions` field in the cabal file, it throws the same error. Isn't there any wordaround other than putting LANGUAGE pragma into every file in the project?

ndmitchell:
You can put it into a `.hlint.yaml` file as `- arguments: [-XQuasiQuotes]`, which should mean every folder can find it. Does that solve the problem? I note that isn't very well documented, so definitely needs adding to the README.

pythonissam:
Thank you! It worked perfectly!

PR #481, opened 2018-05-08
update README

pythonissam:
I updated the document. If it is unnatural, please correct.


pythonissam:
ref: https://github.com/ndmitchell/hlint/issues/472

ndmitchell:
Thanks! Looks great.

PR #485, opened 2018-05-08
Fix code climate build, broken by removal of stack.yaml

eborden:
The Code Climate build relied on the existence of the `stack.yaml` file.
When that was removed this build broke. I've reconfigured the Dockerfile
to initialize `stack.yaml` itself.

There is probably a way to get this tested in CI. I'm not sure if you have
opinions there.

cc: @dblandin @larkinscott


ndmitchell:
Thanks! Hopefully this should break very rarely, so testing won't be necessary.

PR #495, opened 2018-06-05
sortBy . comparing → sortOn

dzhus:
Evening

I discovered `sortOn` yesterday and rushed to see if it's in hlint yet. 7e775ed was made 16 hours prior but it doesn't appear to suggest «[the Schwartzian transform](http://hackage.haskell.org/package/base-4.11.1.0/docs/Data-List.html#v:sortOn)» itself. I'm not sure if I missed anything but it's added in this PR. Applied to the source of hlint itself as well.


ndmitchell:
Thanks! Good catch.

PR #498, opened 2018-06-09
fix "remove it." to "remove it?"

justinwoo:
This has been bothering me for a while, so I made a PR for it.

### DIFF
--- a/src/CC.hs
+++ b/src/CC.hs
@@ -89,7 +89,7 @@ fromIdea Idea{..} = Issue
         , T.pack from
         , "```"
         , ""
-        , "remove it?"
+        , "remove it."
         ]
 
     content from (Just to) = T.unlines
@@ -99,7 +99,7 @@ fromIdea Idea{..} = Issue
         , T.pack from
         , "```"
         , ""
-        , "Why not"
+        , "You might want"
         , ""
         , "```"
         , T.pack to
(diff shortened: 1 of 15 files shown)


ndmitchell:
Thanks for highlighting the issue, I can see why it rankles you - we should definitely fix it.

However, I wonder if that is the right fix? What I'm trying to express is:

```
Found:
  old code
Why not:
  new code
```

But where there is no `new code`. Is the `Why not` in the normal case, without ?, annoying? It seems weird to have question marks for one but not the other. Is `Alternative:` or `Suggestion:` better? But when does `Suggestion` overlap with the warning level suggestion?

justinwoo:
Seems like there's quite a lot of overlapping places, so one approach I've tried is to remove "Why not" altogether and replace it with "You might want", as both `You might want:` and `You might want to remove it.` work as non-interrogatives

justinwoo:
Pushed up the changes. Removal suggestion is longer now but should be fine.

src/CC.hs
```diff
@@ -99,7 +99,7 @@ fromIdea Idea{..} = Issue
         , T.pack from
         , "```"
         , ""
-        , "Why not"
+        , "You might want"
```
ndmitchell: Did this one get forgotten?

justinwoo:
I think "perhaps you should" is the common form in US and elsewhere, whereas the first is less familiar.

ndmitchell:
Makes sense - thanks for the cleanup!

PR #500, opened 2018-06-15
hint for redundant return after when/unless

yairchu:
Recently hlint warns at redundant returns after `traverse_` and others, this commit adds `when` and `unless`.


ndmitchell:
Thanks! I also note `void` is a function returning unit, so included that in your list. Nice ideas!

PR #513, opened 2018-07-31
HTML report: Update the location bar on load and update

chreekat:
Makes it easy to copy/save/share links to particular hints or files!

See a preview at http://databrary.github.io/databrary/hlint.html.

Downsides:
1. The sidebar jumps around, since the url hash pieces are the same as anchor ids. This could be fixed with a modicum more sophistication, but it might also be considered a feature.
2. Setting the location hash to '' leaves an ugly '#' appended to the url. I don't want to work around it, though, until I understand why that is how the DOM API works.

### DIFF
--- a/data/report_template.html
+++ b/data/report_template.html
@@ -29,19 +29,31 @@ function insertRule(s)
 
 // SHOW/HIDE LOGIC //
 
-var last = "";
-
-function show(id)
+/**
+ * @param {bool} [initialise=false] - Used to signal we are loading the page, so
+ the current location doesn't match our presumed state.
+ */
+function show(id, initialise=false)
 {
-    if (id == last) return;
-    if (id == "")
+    var last;
+    if (initialise)
+    {
+        last = "";
+    }
+    else
+    {
+        last = document.location.hash.slice(1);
+    }
+
+    if (id === last) return;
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks! I won't be able to review this for about a week, but will do so then - the idea is good so will give it a try then.

ndmitchell:
Thanks for the patch! Nice work. I also improved it to not jump around and remove the ugly hash by using `history.replaceState`. For info, I thought of two further improvements, which might be valuable if you are doing work and using the report (I tend not to use it myself):

1. Use `hash` on the name to derive an id (with a backup of what you have now if there are clashes). Advantage is that given a link it's likely to remain stable even if it's regenerated.
2. Use folder structure for files, so everything below `src` is one level, then `src/Model`. Would also let you see all errors/warnings in a folder at a time.

However, no requirement to do them, just ideas based on how you seem to be using the feature.

chreekat:
Nice! I agree it might be useful to make names more stable, though that change might wait for the next project I inherit that has 500+ hints to consider. :)

ndmitchell:
Fair enough - I raised https://github.com/ndmitchell/hlint/issues/515 so the ideas don't get forgotten.

PR #520, opened 2018-08-27
Fix #75: Windows 10: Enable ANSI-capable terminals

### LINKED ISSUE #75: Enable color on all platforms
Best done by using the `ansi-terminal` library.

mpilgrem:
This proposed pull request suggests that, on Windows, `main` should try to enable virtual terminal (VT) processing on `stdout`. On Windows 10, the native (ConHost) terminals (Command Prompt and PowerShell) are ANSI-capable but not ANSI-enabled by default; applications have to enable the ANSI-capability.

Some C pre-processor directives are used, to add the Windows-specific code to `Main.hs`; I did not know how to avoid any CPP. `if os(windows)` is used in `hlint.cabal`.

(The addition of `stack.yaml` and most of the changes to `hlint.cabal` seemed necessary in order to build `hlint` (using `stack build`), in order to test the other changes. (`stack init` would not work for me.)

The example below shows: (top) `hlint --color` 'as is' (`hlint --version` is `HLint v2.1.10, (C) Neil Mitchell 2006-2018`) and the problem of 'raw' 'ANSI' codes being sent to the terminal; and (bottom) `hlint --color` after these changes (executed via the `stack exec` function).

image

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,7 @@ test_script:
 - set HASKELL_DEPENDENCIES=apply-refact
 - ps: Invoke-Expression (Invoke-WebRequest 'https://raw.githubusercontent.com/ndmitchell/neil/master/appveyor.ps1')
 # - set UNIPLATE_VERBOSE=-1 # Temporary
-- stack exec -- hlint test
+- stack exec -- hlint --test
 - stack exec -- hlint src
 # Test the documented way of running it in appveyor
 - ps: Invoke-Command ([Scriptblock]::Create((Invoke-WebRequest 'https://raw.githubusercontent.com/ndmitchell/hlint/master/misc/appveyor.ps1').Content)) -ArgumentList @('--version')
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -65,7 +65,7 @@ library
         cpphs >= 1.20.1,
         cmdargs >= 0.10,
         uniplate >= 1.5,
-        ansi-terminal >= 0.6.2,
+        ansi-terminal >= 0.8.1,
         extra >= 1.7.3,
         refact >= 0.3,
         aeson >= 1.1.2.0,
--- a/src/CmdLine.hs
+++ b/src/CmdLine.hs
@@ -20,7 +20,7 @@ import Language.Haskell.GhclibParserEx.GHC.Driver.Session as GhclibParserEx
 import DynFlags hiding (verbosity)
 
 import Language.Preprocessor.Cpphs
-import System.Console.ANSI(hSupportsANSI)
+import System.Console.ANSI(hSupportsANSIWithoutEmulation)
 import System.Console.CmdArgs.Explicit(helpText, HelpFormat(..))
 import System.Console.CmdArgs.Implicit
 import System.Directory.Extra
@@ -29,7 +29,6 @@ import System.Exit
 import System.FilePath
 import System.IO
 import System.IO.Error
-import System.Info.Extra
 import System.Process
 import System.FilePattern
 
@@ -89,7 +88,7 @@ data ColorMode
 
 
 instance Default ColorMode where
-  def = if isWindows then Never else Auto
+  def = Auto
 
 
 data Cmd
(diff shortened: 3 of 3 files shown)


mpilgrem:
I have revised my pull request, as I realised my previous mistake with `stack.yaml` and `hlint.cabal`.

ndmitchell:
Thanks! Colors on Windows would be great.

I was expecting such a PR to move to using the ansi-terminal library instead, any reason for preferring to try and enable ANSI in the console?

Any chance the console enabling thing can be moved to ansi-terminal itself? It seems like its general purpose code that shouldn't really live in HLint only.

I see you move the Main file from src to app - any reason for that?

ndmitchell:
1. The `hscolour` package I depend on is GPL, and that causes issues with some distributions of HLint alreaady, to the extend that there is a flag to remove the GPL dependency. It also doesn't do that much interesting - it syntax highlights the Haskell - but looking at what GHC does in error messages syntax highlighting the code seems a lot less useful than coloring the contextual HLint pieces. As a result, I wonder about dropping hscolour, and either hand-rolling something trivial to syntax highlight (which could use ansi-terminal), or not having syntax highlighting but using color in other ways (which could also use ansi-terminal).

2. Big fan of that! I don't want to maintain the code above, and I might want to reuse it.

3. If the code moved to ansi-terminal, or we followed one of the paths above, that wouldn't be necessary, which is my preference if it's feasible.

And I agree that enabling color by default is the right way forward once we have it working! See what ghcid does for my current thinking on enabling color, https://github.com/ndmitchell/ghcid/blob/master/src/Ghcid.hs#L193

mpilgrem:
@ndmitchell, I've revisited this pull request after a hiatus, rebasing on the current master, correcting a compiler warning elsewhere, and turning colour off on those tests that are confused by colour in the output. CI now succeeds.

mpilgrem:
I have revisited this pull request, rebasing on the current master.

ndmitchell:
Thanks for these changes, and sorry for the (disastrously) long time its taken for me to give them proper consideration. Looks good! My only question is why is the `--color=never` required on the tests? Those tests are run piping the stdout to a buffer. So why does it detect that a pipe buffer can accept color?  Should we be calling `hIsTerminalDevice` https://hackage.haskell.org/package/base-4.14.0.0/docs/System-IO.html#v:hIsTerminalDevice? If so, would that eliminate the need for `--color=never`?

mpilgrem:
On Windows, `hSupportsANSIWithoutEmulation` makes use of all three states: `Just False` - definitely not; `Just True` - definitely yes; and `Nothing` - can't tell, one way or the other. 'Can't tell' could be a handle connected to a file or a handle connected to a non-native terminal (other than `mintty`) implemented using redirection. On Windows, `hIsTerminalDevice` can't identify a terminal implemented using redirection as being a terminal. (`mintty` is a popular example of a terminal implemented using redirection, but fortunately the `mintty` package has a fix for that.) 

My pull request errs on the side of `Nothing` implying something that can support ANSI (files can be viewed as either supporting or not supporting ANSI, depending on what you intend to do with the file content), so: `pure $ Just False /= supportsANSI`.

An alternative would be to err on the side of 'Nothing' implying that ANSI is unsupported (and, consequently, treating files as not supporting ANSI - that is, ANSI codes are not wanted in output to files, by default), so: `pure $ Just True == supportsANSI`.

Personally, I do not feel strongly about one alternative over the other.

ndmitchell:
I think I'd prefer to err on the side of caution to start with. I imagine `hlint src > file.txt` detects as Nothing? And if so, we definitely don't want escape codes or similar showing up in `file.txt`. If you do Nothing can we remove all the `--color=never` flags in the tests?

ndmitchell:
Thanks for fixing the Appveyor too! Truth be told, I didn't notice it was broken, since the Windows builds have been segfaulting enough that I no longer care. Great to have your improvements merged after so long!

mpilgrem:
I know what you mean. However, GHC 8.8.4 (resolver `lts-16.13`) works on Windows 10 version 2004. I can build `hlint` on Windows with that resolver.

ndmitchell:
Yep, I'm using it on windows too as my primary system. But there are a bunch of segfaults that have been identified and break the CI pretty regularly.

PR #535, opened 2018-09-26
add more rules to the lens group

yairchu:
(these rules are for stuff that I had stumbled with in my own code and I believe they could help others)

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -635,6 +635,7 @@
     rules:
     - warn: {lhs: "(a ^. b) ^. c", rhs: "a ^. (b . c)"}
     - warn: {lhs: "(a ^. b) ^? c", rhs: "a ^? (b . c)"}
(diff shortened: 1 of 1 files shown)


ndmitchell:
Not a fan of the zipWith ones since they're changing from the base library to the lens library. I personally prefer zipFrom to that pattern. The rest look good though, assuming they are correct (I don't know enough about lens to be sure - so please double check them yourself)

yairchu:
I I guess those would better fit a `lens-zealots` group :) Which could have more rules, like:

    - hint: {lhs: "Data.Set.member a", rhs: "(^. Control.Lens.contains a)"}

ndmitchell:
A lens zealots group off by default is fine by me.

yairchu:
I was half kidding with the `lens-zealots` group :)

ndmitchell:
Thanks, looks good!

Issue #546, opened 2018-11-03
Suggest `x <$ y` instead of `const x <$> y`


foxbenjaminfox:
More than once I've found myself writing something along the lines of `const x <$> y`, and only later realizing that I can replace that with `x <$ y`. This is definitely something that hlint could suggest.

The hint for this would be
```yaml
- hint: {lhs: const x <$> y, rhs: x Data.Functor.<$ y}
```
There probably should also be the equivalent hints for `pure x <$> y` and `return x <$> y`.

Perhaps for completeness's sake we'd also want to suggest `x $> y` instead of `x <&> const y`, and likewise for `x <&> pure y` and `x <&> return y`.

If you're OK with these, I'll open a PR. I'm not so sure if the flipped versions comes up often enough to be worthwhile, so if you don't think they're worth it I can leave them out.

ndmitchell:
`<$` is in the Prelude, so no need to qualify even. Is the `pure`/`return` equivalent to const in these cases, being a `->` monad? Are pure/return there common, if they are just const? Can the `pure`/`return` there be in a different monad, such that the simplification is no longer valid? (My gut feel is yes, but maybe the application after stops that being the case?)

I'm happy with these, even the flipped versions. Query on if the pure/return playing the role of const, but only on that aspect.

foxbenjaminfox:
About `<$` being in the Prelude: I was basing the qualification based on [your comment](https://github.com/ndmitchell/hlint/issues/395#issuecomment-338491540) on #395, so I'm confused, because `<$` was already in the Prelude at that point (since GHC 7.10 in 2015, according to [this Trac issue](https://ghc.haskell.org/trac/ghc/ticket/10113)).

About `pure`/`return`: Yes, they're playing the role of `const` here, and only `const`. The first argument of `<$>` (or of `fmap`) is a function, so if the first argument to `<$>` is `pure x`, then `pure x` has to be a function. `pure x` has the type `Application f => f a`, and the only way that type can be a function is if `f` is specialized to `(->) r` (for some `r`).

foxbenjaminfox:
I just realized that I was confused between `<$` and `$>`: it turns out that while `<$` is in the Prelude, for some reason `$>` isn't. So that's why `$>` needed qualification in #395, whereas `<$` doesn't. So never mind about that then.

ndmitchell:
Do people really use pure/return instead of const? I know they could, but does it happen in practice?

foxbenjaminfox:
[...]

PR #547, opened 2018-11-03
 Suggest `x <$ y` instead of `const x <$> y` (#546)

foxbenjaminfox:
This pull request adds the hints I proposed in #546.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,7 @@
 Changelog for HLint (* = breaking change)
 
+    #546, suggest `x $> y` for `const x <$> y`, `pure x <$> y`, and `return x <$> y`
+    #546, suggest `x <$ y` for `x <&> const y`, `x <&> pure y`, and `x <&> return y`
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks for the PR. This fails with:

```
INPUT: yes = pure alice <$> [1, 2]-- alice <$ [1, 2]
OUTPUT: C:\projects\hlint\.stack-work\install\9b4c8a4d\share\x86_64-windows-ghc-8.6.1\hlint-2.1.10\hlint.yaml:1:7: Suggestion: Use <$
Found:
  pure alice <$> [1, 2]
Perhaps:
  alice <$ [1, 2]
WANTED: <failure>
```

Is it possible you need a space before the `--`? I wrote the test generator, and certainly all my other examples have the space there.

foxbenjaminfox:
Yes, that was a typo, which I've now fixed.

For some reason, that failure didn't show up when I ran `stack exec hlint test` (as suggested by the README.) I'm not sure why that is.

PR #566, opened 2018-12-10
Update link for 'Dhall to HLint' blog post

ChShersh:
Should work without `.html` 👍 

P.S. Thanks for mentioning our blog post in README!


ndmitchell:
Thanks!

PR #575, opened 2018-12-27
add fixities for operators from the constraints and lattices packages

yairchu:
I've had a hint misfire due to not knowing `\/`'s fixity. Haven't had trouble with `\\` yet but I'm guessing that's bound to happen to someone :)


ndmitchell:
The `\\` operator is already in the base library, so I'd rather not override it to be a different fixity. Happy to take the lattice ones though.

yairchu:
Good point. I removed that one from the PR

ndmitchell:
Thanks!

PR #576, opened 2019-01-08
List Restyled in Integrations section of README

pbrisbin:
:wave: Hey Neil, I recently added and HLint "restyler" to my Restyled side-project and would love to see it listed here as an Integration. Totally open to change the language however you'd like. Thanks!


ndmitchell:
Awesome! Looks like a cool project too. Great to see, and delighted to include it.

Issue #579, opened 2019-01-11
Redundant LANGUAGE pragmas which are implied by others


yairchu:
`KindSignatures` is implied by `PolyKinds` and by `TypeFamilies`. Furthermore, often GHC will tell you to enable it and then tell you to enable another one which implies it, which often creates redundant language pragmas. I suggest `hlint` should hint to remove it.

ndmitchell:
Solved by #580

PR #580, opened 2019-01-11
Hints to remove language extensions implied by other enabled extensions

yairchu:
PR for https://github.com/ndmitchell/hlint/issues/579
It's no longer true that uses of TH disable extension hints, as this new hint is still valid when TH is used.
I don't know if there's a list of other implied language pragmas. If there are, it would be easy to add them.

`allExts` is calculated in `extensionsHint` to avoid recalculating it in calls to `impliedExt`/`minimalExtensions`.

### DIFF
--- a/src/Hint/Extensions.hs
+++ b/src/Hint/Extensions.hs
@@ -143,41 +143,44 @@ extensionsHint _ x = [rawIdea Warning "Unused LANGUAGE pragma" (srcInfoSpan sl)
           (warnings old new) [refact]
     | o@(LanguagePragma sl exts) <- modulePragmas x
     , let old = map (parseExtension . prettyPrint) exts
-    , let new = minimalExtensions allExts x old
+    , let new = (if usedTh then old else minimalExtensions x old) \\ impliedExts
     , let newPragma = if null new then "" else prettyPrint $ LanguagePragma sl $ map (toNamed . prettyExtension) new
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks for the patch!

I'm still wondering what I think about this. Is it a good idea? If you write `MonoLocalBinds` and `GADTs` did you mean it, or do you always want `MonoLocalBinds` gone? My concern is that a lot of people have no idea which extensions imply which. Are you writing those extensions to shut the compiler up, or to document your intent? On the reverse, if you get your intent wrong, you don't get an error, so the odds of peoples intent being correct are low.

I opened it up to Twitter: https://twitter.com/ndm_haskell/status/1086017990578135040 (result is non-binding but curious to get a finger in the air poll)

If we do go ahead, it would be very useful to turn it into a more data driven approach. Something like:

```
implies :: [(Extension, Extension)]
```

Which just records that A implies B. From that we can iterate, build maps and potentially use the data in other ways we haven't yet explored.

ndmitchell:
Twitter spoke and said yes 93%, so make it data driven and I'll merge.

yairchu:
Made it data driven now. Also, made the `used TemplateHaskell` check happen at most once again (see commit)

yairchu:
Awesome! It's much better when it is explained :)

Issue #590, opened 2019-01-29
improve readability of "Use fewer LANGUAGE pragmas"?


Javran:
Hi there! Just want to open a discussion about the  "Use fewer LANGUAGE pragmas" message:

It currently looks like:
```
Warning: Use fewer LANGUAGE pragmas
Found:
  {-# LANGUAGE LambdaCase, TypeApplications, NamedFieldPuns,
    LambdaCase, ScopedTypeVariables, DataKinds #-}
Perhaps:
  {-# LANGUAGE LambdaCase, TypeApplications, NamedFieldPuns,
    ScopedTypeVariables, DataKinds #-}

```
This might be easy for some refactoring tools to process, but I feel as a human being I care more about "which one is spurious" and unfortunately this info is hard to see from the output - I have to read through two lists to spot the difference (not to mention the situation becomes worse when you have a long list of LANGUAGE pragmas), any idea about how this can be improved?

ndmitchell:
Thanks to @yairchu it now says which extensions are unused as notes. Hopefully that makes it clear.

Javran:
Thanks! Really appreciate it!

PR #592, opened 2019-01-31
mention which extensions are not used (#590)

yairchu:
Adds notes such as:

    Note: Extension LambdaCase is not used


ndmitchell:
Awesome, thanks!

PR #600, opened 2019-02-24
Add lens hints for 'Either'

alexfmpe:


ndmitchell:
The other lens hints all deliberately go from a Lens operator, to a better lens operator, but don't tell people to use Lens who aren't already using it. As such, they are enabled by default. I'd be happy with a lens2 hints group which was disabled by default containing such hints, but don't want to push people at lens unless they opt in.

ndmitchell:
Thanks, being in its own group sounds good.

PR #601, opened 2019-02-24
Add 'mfilter' hint for 'Maybe'

alexfmpe:


data/hlint.yaml
```diff
@@ -473,6 +473,7 @@
     - hint: {lhs: case m of Nothing -> Nothing; Just x -> x, rhs: Control.Monad.join m}
     - hint: {lhs: maybe Nothing id, rhs: join}
     - hint: {lhs: maybe (f x) (f . g), rhs: f . maybe x g, note: IncreasesLaziness, name: Too strict maybe}
+    - warn: {lhs: \x -> if f x then Just x else Nothing, rhs: mfilter f . Just}
```
ndmitchell:
```suggestion
    - warn: {lhs: if f x then Just x else Nothing, rhs: mfilter f (Just x)}
```

Eta-reducing the hints tends to make them match less, not more.

alexfmpe:
> I appreciate that this avoids mentioning x twice

Another advantage in my book is that it signals the intent of syncing the payload and the target of the predicate. You can't accidentally change one (to something of the same type) and forget the other one.

> the use of a relatively rare function like mfilter makes it harder for me to understand than the original

I agree with this argument in the vast majority of cases. I'm kind of on the fence for *hlint in particular*, since I see learning equivalences as being one of the greatest benefits of hlint.

> I'm inclined to say not one of the standard set of hints.
Sure. Looks like a fit for `generalise-for-conciseness` ?

ndmitchell:
Can't argue with it generalise for conciseness - thanks!

PR #606, opened 2019-03-07
Arguments passed from cli are last, taking precedence. Fixes #605

### LINKED ISSUE #605: Command line --color=never argument doesn't override setting in `.hlint.yaml`
```
% hlint --version
HLint v2.1.15, (C) Neil Mitchell 2006-2019
```

phadej:


ndmitchell:
Thanks, makes a lot of sense. Do you want a release with this included? Or shall I wait (potentially a month or so) and do it with the next batch of changes?

phadej:
I'm ok with using locally build version for now.

PR #609, opened 2019-03-16
Add smells to hlint

benweitzman:
This PR adds detection of "code smells" to hlint, inspired by some of the functionality in `swiftlint`.

I've started by adding 4 smells (none of them are on by default):

* Long function bodies
* Functions with many arguments
* Type level lists with many elements (especially useful for programming in some of the free monad effect styles)
* Many imports 

All of these are things that can indicate that it's time to do a little cleanup + reorganization. 

Each smell is parameterized by an integer limit (e.g. function bodies should be < 20 lines long)

### DIFF
--- a/src/Config/Type.hs
+++ b/src/Config/Type.hs
@@ -106,11 +106,12 @@ data Restrict = Restrict
     ,restrictWithin :: [(String, String)]
     } deriving Show
 
-data SmellType = SmellLongFunctions | SmellLongTypeLists deriving (Show,Eq,Ord)
+data SmellType = SmellLongFunctions | SmellLongTypeLists | SmellManyArgFunctions deriving (Show,Eq,Ord)
 
 getSmellType :: String -> Maybe SmellType
 getSmellType "long functions" = Just SmellLongFunctions
 getSmellType "long type lists" = Just SmellLongTypeLists
+getSmellType "many arg functions" = Just SmellManyArgFunctions
 getSmellType _ = Nothing
 
 data Smell = Smell
--- a/src/Config/Type.hs
+++ b/src/Config/Type.hs
@@ -106,10 +106,11 @@ data Restrict = Restrict
     ,restrictWithin :: [(String, String)]
     } deriving Show
 
-data SmellType = SmellLongFunctions deriving (Show,Eq,Ord)
+data SmellType = SmellLongFunctions | SmellLongTypeLists deriving (Show,Eq,Ord)
 
 getSmellType :: String -> Maybe SmellType
 getSmellType "long functions" = Just SmellLongFunctions
+getSmellType "long type lists" = Just SmellLongTypeLists
 getSmellType _ = Nothing
 
 data Smell = Smell
--- a/src/Config/Type.hs
+++ b/src/Config/Type.hs
@@ -1,8 +1,8 @@
 
 module Config.Type(
     Severity(..), Classify(..), HintRule(..), Note(..), Setting(..),
-    Restrict(..), RestrictType(..),
-    defaultHintName, isUnifyVar, showNotes, getSeverity, getRestrictType
+    Restrict(..), RestrictType(..), Smell(..), SmellType(..),
+    defaultHintName, isUnifyVar, showNotes, getSeverity, getRestrictType, getSmellType
     ) where
 
 import HSE.All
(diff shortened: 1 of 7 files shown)


src/Hint/Smell.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE RecordWildCards #-}
+module Hint.Smell where
```
ndmitchell: Can we get an explicit export list

src/Hint/Smell.hs
```diff
@@ -0,0 +1,83 @@
...
+subTypes (TyInfix _ a _ b) = [a, b]
+subTypes (TyKind _ a _) = [a]
+subTypes (TyPromoted _ (PromotedList _ _ x)) = x
+subTypes (TyPromoted _ (PromotedTuple _ x)) = x
+subTypes (TyEquals _ a b) = [a, b]
+subTypes (TyBang _ _ _ a) = [a]
+subTypes _ = []
+
+unrollType :: Type l -> [Type l]
+unrollType t = t : (subTypes t >>= unrollType)
```
ndmitchell: I'd rather use concatMap here as it makes it clearer.
```suggestion
unrollType t = t : concatMap unrollType (subTypes t)
```

src/Config/Type.hs
```diff
@@ -106,11 +106,27 @@ data Restrict = Restrict
...
+    {smellType :: SmellType
+    ,smellLimit :: Int
+    } deriving Show
+
 data Setting
     = SettingClassify Classify
     | SettingMatchExp HintRule
     | SettingRestrict Restrict
     | SettingArgument String -- ^ Extra command-line argument
+    | SettingSmell Smell
```
ndmitchell: I'd be tempted to just inline `Smell` here, so `SettingSmell SemllType Int`. I'm not sure how much the `Smell` data type adds.

src/Hint/All.hs
```diff
@@ -59,6 +60,9 @@ builtin x = case x of
     HintComment    -> comm commentHint
     HintNewType    -> decl newtypeHint
     HintRestrict   -> mempty{hintModule=restrictHint}
+    HintSmell      -> mempty{hintDecl=smellHint
```
ndmitchell: On one line should be sufficient, and then its easier to read the list of hints

benweitzman:
I looked into adding a `@smell` annotation with additional settings, and it looks like it's going to be tricky since the part of the code that deals with annotations happens after from applying the rules. 

How do you feel about this syntax:

```haskell
{-
<TEST> [{smell: { type: many arg functions, limit: 2 }}]
f :: Int -> Int -> Int -> Int -> Int; f = undefined --
</TEST>
-}
```

You can apply any config (in addition to what is defined in the project `.hlint.yaml`) inline for the entire test group.

ndmitchell:
Perfectly happy with that syntax of its easier.

benweitzman:
Great, I have that working now, will start adding some tests. 

Re `unrollType`, the purpose of this is to catch things that are hidden in higher order functions. 

For example, if you say you don't want functions with  3 or args, then you don't want something like 

`f :: (a -> b -> c -> d) -> e`

`f` only has one arg, but that arg is a function with `3` args. 

Unrolling the types + sub types into a list was just what I felt was the easiest way to work with the type structure defined in `haskell-src-exts`, since that structure is not very friendly to traversing. 


Re long type lists:

My organization uses the `freer-simple` effect system, which represents effects in a type-level list:

(e.g `Eff '[Reader String, State Bool] Integer` ) 

The smell here arises when you have a computation that is using a lot of effects. It becomes hard to reason about, hard to test, and it pollutes the effect list of every other functions that uses the computation .

Counting + limiting the number of items in the list can indicate that it's time to make a higher level effect that masks some of the lower lever ones, making it easier to reason about, easier to test, etc. 

This smell only checks in type annotations, so if you want to have a long type list you can put it into a type alias.

ndmitchell:
Have you seen the `universe` functions used elsewhere in HLint? They automatically unroll all the way and are much more amenable to these kinds of problems.

benweitzman:
I think I've addressed all of the technical issues. I changed the long function smell to take where clauses into account (added more tests for this too). 

I've written a draft blog post, would you mind looking it over? https://medium.com/@ben_80237/continuous-improvement-with-hlint-code-smells-e490886558a1 Would appreciate your thoughts. 

Do you have any thoughts on how you want to include documentation? I didn't see a natural place to add anything to in the README.

ndmitchell:
Code looks awesome. One final remark - when we have two smell declarations for the same type, we take the first. I'd be tempted to say the minimum, as that's the effect of running with both smell declarations separately and combining the result. 

Blog post looks great.

For user manual I suggest a new subsection before "Adding Hints". Feel free to link to the blog post too. It doesn't need to be too long.

ndmitchell:
I'm happy to merge now and do the docs after, or save it and do it together. Your call.

ndmitchell:
This is now out in HLint 2.1.16 so please put the blog post up whenever you are ready!

PR #612, opened 2019-03-25
New hint: replace 'maybe Nothing f' with 'f =<<'.

23Skidoo:
What the title says.


ndmitchell:
I had to think about this for a while - it's more polymorphic which I'm not a huge fan of. But I decided to include it in the end. Thanks!

PR #614, opened 2019-03-31
Adds refactorings for camelCase and some list suggestions

SarthakKumar1997:
Adds refactorings for the "Use camelCase" suggestion and for four list suggestions in `src/Hint/List.hs`.

### DIFF
--- a/src/Hint/List.hs
+++ b/src/Hint/List.hs
@@ -40,7 +40,8 @@ import Hint.Type
 import Data.List.Extra
 import Data.Maybe
 import Prelude
-import Refact.Types
+import Refact.Types hiding (SrcSpan)
+import qualified Refact.Types as R
 
 
 listHint :: DeclHint
@@ -55,19 +56,24 @@ listDecl x =
 
 listComp :: Exp_ -> [Idea]
 listComp o@(ListComp a e xs)
-    | "False" `elem` cons = [suggest "Short-circuited list comprehension" o o' [Replace Expr (toSS o) [] (prettyPrint o')]]
-    | "True" `elem` cons = [suggest "Redundant True guards" o o2 [Replace Expr (toSS o) [] (prettyPrint o2)]]
-    | let ys = moveGuardsForward xs, xs /= ys = [suggest "Move guards forward" o (ListComp a e ys) [Replace Expr (toSS o) [] (prettyPrint (ListComp a e ys))]]
+    | "False" `elem` cons = [suggest "Short-circuited list comprehension" o o' (suggestExpr o o')]
+    | "True" `elem` cons = [suggest "Redundant True guards" o o2 (suggestExpr o o2)]
+    | xs /= ys = [suggest "Move guards forward" o o3 (suggestExpr o o3)]
     where
+        ys = moveGuardsForward xs
         o' = List an []
         o2 = ListComp a e $ filter ((/= Just "True") . qualCon) xs
+        o3 = ListComp a e ys
         cons = mapMaybe qualCon xs
         qualCon (QualStmt _ (Qualifier _ (Con _ x))) = Just $ fromNamed x
         qualCon _ = Nothing
 listComp o@(view -> App2 mp f (ListComp a e xs)) | mp ~= "map" =
-    [suggest "Move map inside list comprehension" o o2 [Replace Expr (toSS o) [] (prettyPrint o2)]]
+    [suggest "Move map inside list comprehension" o o2 (suggestExpr o o2)]
     where o2 = ListComp a (App an (paren f) (paren e)) xs
 listComp _ = []
+
+suggestExpr :: Exp_ -> Exp_ -> [Refactoring R.SrcSpan]
+suggestExpr o o2 = [Replace Expr (toSS o) [] (prettyPrint o2)]
 
 -- Move all the list comp guards as far forward as they can go
 moveGuardsForward :: [QualStmt S] -> [QualStmt S]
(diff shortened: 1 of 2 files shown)


SarthakKumar1997:
Warning: DO NOT MERGE THIS COMMIT. See issue #615

SarthakKumar1997:
This commit can now be merged.

src/Hint/Naming.hs
```diff
@@ -104,6 +104,6 @@ suggestName x
 
 
 replaceNames :: Data a => [(String,String)] -> a -> a
-replaceNames rep = descendBi f
+replaceNames rep = transformBi f
```
ndmitchell: I am unconvinced this changes anything. Before it looked at all the Ident's within the data type. Now it also looks at all the idents inside indents. Does it have any change for you?

src/Hint/Naming.hs
```diff
@@ -49,9 +49,9 @@ namingHint :: DeclHint
 namingHint _ modu = naming $ Set.fromList $ concatMap getNames $ moduleDecls modu
 
 naming :: Set.Set String -> Decl_ -> [Idea]
-naming seen x = [suggestN "Use camelCase" x2 (replaceNames res x2) | not $ null res]
+naming seen x = [suggest "Use camelCase" x x2 [Replace Bind (toSS x) [] (prettyPrint x2)] | not $ null res]
     where res = [(n,y) | n <- nubOrd $ getNames x, Just y <- [suggestName n], not $ y `Set.member` seen]
-          x2 = shorten x
+          x2 = replaceNames res x
```
ndmitchell: The reason for calling `shorten` here was if you are suggesting to change the name of a function that is 100 lines long then all you see is the body of the function, to `shorten` tries to tidy up the result. Perhaps it should be `shorten $ replaceNames res x`?

src/Hint/Naming.hs
```diff
@@ -49,9 +49,9 @@ namingHint :: DeclHint
 namingHint _ modu = naming $ Set.fromList $ concatMap getNames $ moduleDecls modu
 
 naming :: Set.Set String -> Decl_ -> [Idea]
-naming seen x = [suggestN "Use camelCase" x2 (replaceNames res x2) | not $ null res]
+naming seen x = [suggest "Use camelCase" x x2 [Replace Bind (toSS x) [] (prettyPrint x2)] | not $ null res]
```
ndmitchell: Before I think this suggesting replacing top-level names. Now I think it might suggest all recursive names? That seems like too much, since having internal names be whatever you want seems reasonable. Or am I misreading?

src/Hint/Naming.hs
```diff
@@ -49,9 +49,9 @@ namingHint :: DeclHint
 namingHint _ modu = naming $ Set.fromList $ concatMap getNames $ moduleDecls modu
 
 naming :: Set.Set String -> Decl_ -> [Idea]
-naming seen x = [suggestN "Use camelCase" x2 (replaceNames res x2) | not $ null res]
```
ndmitchell: In CI this file gives an error about Replace and Bind not being in scope. How come? Did you forget to commit some imports?

src/Hint/List.hs
```diff
@@ -55,16 +55,17 @@ listDecl x =
...
+    | let ys = moveGuardsForward xs, xs /= ys = [suggest "Move guards forward" o (ListComp a e ys) [Replace Expr (toSS o) [] (prettyPrint (ListComp a e ys))]]
     where
+        o' = List an []
         o2 = ListComp a e $ filter ((/= Just "True") . qualCon) xs
         cons = mapMaybe qualCon xs
         qualCon (QualStmt _ (Qualifier _ (Con _ x))) = Just $ fromNamed x
         qualCon _ = Nothing
 listComp o@(view -> App2 mp f (ListComp a e xs)) | mp ~= "map" =
-    [suggest "Move map inside list comprehension" o o2 []]
+    [suggest "Move map inside list comprehension" o o2 [Replace Expr (toSS o) [] (prettyPrint o2)]]
```
ndmitchell: I wonder if a helper `suggestExpr` which takes two expressions and generates `[Replace Expr (toSS o) [] (prettyPrint o2)]` might be valuable?

src/Hint/Naming.hs
SarthakKumar1997: Yes. Good spot :) will recommit with the new import.

src/Hint/Naming.hs
SarthakKumar1997: Ah, forgot to rewind this in the second commit, which changes the refactoring from Decl to Bind. Will do so in the new commit.

src/Hint/Naming.hs
SarthakKumar1997: Sounds good - have shorted `x` as well as `replaceNames res x` in the new commit.

src/Hint/List.hs
SarthakKumar1997: Makes sense - it simplifies four different refactorings! Have done this :)

SarthakKumar1997: I've put this in the List.hs file for now, but perhaps it might be useful as as a system wide helper function. Up to you...

src/Hint/Naming.hs
SarthakKumar1997: Apologies, I'm not sure what you mean exactly - I'm guessing this changes because of `descendBi`?

ndmitchell:
Thanks, looks great now!

Issue #623, opened 2019-04-16
Operator wildcards


yairchu:
hlint treats single letter variable names as wildcards in its rule. Would be nice if something like a single letter operator could also be treated as a wildcard. Then one could add a rule such as ```{lhs: \x -> x `o` a, rhs: (`o` a)}``` instead of the currently hard-coded functionality of the `Avoid lambda` rule.

ndmitchell:
We could make `?` (and any iterated sequences of `?`) the universal wildcard operator. I'd be happy with a patch doing that.

yairchu:
Iterated sequences of `?`, like [`??`](https://www.stackage.org/haddock/lts-13.17/lens-4.17/Control-Lens-Lens.html#v:-63--63-) are and could be actual operators though..

yairchu:
[This commit](https://github.com/yairchu/hlint/commit/5095b37be88ecb4ef0011aa7ee0ef6a75389f631) adds this functionality, but I feel kind of bad about the `(toNamed (fromNamed op))` parts that I used there. Perhaps you have suggestions?

ndmitchell:
Only way to avoid the toNamed/fromNamed would be to have a separate Subst for this part which records the variable substitutions separate from the operator substitutions - but that seems like overkill. Overal, looks quite reasonable.

PR #626, opened 2019-04-17
Operator wildcards

yairchu:
Fix https://github.com/ndmitchell/hlint/issues/623

### DIFF
--- a/data/Test.hs
+++ b/data/Test.hs
@@ -43,6 +43,8 @@ error = zip [1..length x] x ==> zipFrom 1 x
 
 error = before a ==> after a
 
+warn "noop" = a ? 0 ==> a
+
 {-
 <TEST>
 main = readFile "foo" >>= putStr            \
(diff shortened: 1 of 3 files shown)


ndmitchell:
Looks good. Can I get:

* A comment that says that subst on ??? style variables is guaranteed to be a Var
* A test. I think `data/Test.hs` might be a reasonable place to put it?

yairchu:
Added a test.

As `toNamed` is invoked in 6 places here, where would you place the comment?

ndmitchell:
I'd put the comment on the type Subst itself

yairchu:
Changed the code because CI noted that `hlint` proposes improvements, and in that I also added the comment in an extracted function that now does the `toNamed . fromNamed`.

ndmitchell:
Thanks, all merged! Would a release help you?

yairchu:
There's no rush, I'm just using the modified version :)

Do you think that it would be a good idea now to replace the operator sections code in `niceLambdaR` with more succinct rules in `hlint.yaml`?

ndmitchell:
The `niceLambdaR` stuff gets used in a few places, as a helper in other rules, so I don't think replacing it would work.

ndmitchell:
FWIW I released a new version anyway

PR #629, opened 2019-04-25
Only unlit .lhs files

shayne-fletcher-da:
This isn't the whole story with respect to the using of `unlit` but a quick patch to restrict its use to only files with suffix '.lhs'.

### DIFF
--- a/src/GHC/Util.hs
+++ b/src/GHC/Util.hs
@@ -59,5 +59,5 @@ parseFileGhcLib filename str =
   Lexer.unP Parser.parseModule parseState
   where
     location = mkRealSrcLoc (mkFastString filename) 1 1
-    buffer = stringToStringBuffer (unlit filename str)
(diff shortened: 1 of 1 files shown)


src/GHC/Util.hs
```diff
@@ -59,5 +59,5 @@ parseFileGhcLib filename str =
   Lexer.unP Parser.parseModule parseState
   where
     location = mkRealSrcLoc (mkFastString filename) 1 1
-    buffer = stringToStringBuffer (unlit filename str)
+    buffer = stringToStringBuffer (if ".lhs" `isSuffixOf` filename then unlit filename str else str)
```
ndmitchell:
```suggestion
    buffer = stringToStringBuffer (if takeExtension filename == ".lhs" then unlit filename str else str)
```

ndmitchell:
Looks good, but suggest applying my tweak and I'll merge.

ndmitchell:
Changed my mind - merged as is and I'll make the suggested edit on my side.

PR #630, opened 2019-05-02
Some small preparation towards eliminating the SDoc field from ParseError

shayne-fletcher-da:
In this PR, we lift a new function `failOpParseModuleEx`, lifted out of `parseModuleEx`. The intent is to then split this function into two cases, preferring to construct `ParseError` values from GHC `ParseFailed _` values when they exist else fall back on the standard HSE implementation. Doing that will no doubt involve updating test results (error strings will change). In any case, at that point we'll then eliminate the `SDoc` field from `ParseError` (which will remain unused up to its elimination). Tests pass and no new behaviors are introduced.

### DIFF
--- a/src/HSE/All.hs
+++ b/src/HSE/All.hs
@@ -160,6 +160,31 @@ data ParsedModuleResults = ParsedModuleResults {
   , pm_ghclib :: Maybe (Located (HsSyn.HsModule HsSyn.GhcPs)) -- ghc-lib-parser result.
 }
 
+-- | Utility called from 'parseModuleEx' and 'failOpModuleEx'.
+mkMode :: ParseFlags -> String -> ParseMode
+mkMode flags file = (hseFlags flags){parseFilename = file,fixities = Nothing }
+
+-- | Error handler called on HSE parse failure.
+failOpParseModuleEx :: String
+                   -> ParseFlags
+                   -> FilePath
+                   -> String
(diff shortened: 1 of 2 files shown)


src/HSE/All.hs
```diff
@@ -172,27 +197,18 @@ parseModuleEx flags file str = timedIO "Parse" file $ do
...
               return $ Right (ParsedModuleResults (applyFixity fixity x, cs) (Just mod))
-            (ParseFailed sl msg, PFailed ps) ->
-              runParseFileContentsWithMode ppstr sl msg flags file str (pprErrMsgBagWithLoc $ snd $ getMessages ps dynFlags)
             (ParseOk (x, cs), PFailed _) ->
               return $ Right (ParsedModuleResults (applyFixity fixity x, cs) Nothing)
-            (ParseFailed sl msg, POk _ _) ->
-              runParseFileContentsWithMode ppstr sl msg flags file str []
+            (hseFail @ (ParseFailed sl msg), ghcFail @ (PFailed ps)) ->
+              failOpParseModuleEx ppstr flags file str hseFail (Just ghcFail)
+            (hseFail @ (ParseFailed sl msg), POk _ _) ->
```
ndmitchell: Generally for @ patterns you don't put any space around them, by convention

ndmitchell:
Thanks! I did a clean up pass after, let me know if I screwed up anything you were hoping for.

PR #635, opened 2019-05-14
Eliminate the recently added SDoc list field from ParseError

shayne-fletcher-da:
In this iteration, the recently added `[SDoc]` field in `ParseError` is removed. Now, if a GHC parse state value is available from a `PFailed s` then the error handling code uses it to produce the `ParseError`. Despite this somewhat fundamental change, the test suite continues to pass in entirety. 

Manual testing seems to indicate that either the old way or the new way, the error message harvested dealing with the failure doesn't actually get reported to the user.

### DIFF
--- a/src/Apply.hs
+++ b/src/Apply.hs
@@ -80,7 +80,7 @@ parseModuleApply flags s file src = do
     res <- parseModuleEx (parseFlagsAddFixities [x | Infix x <- s] flags) file src
     case res of
         Right (ParsedModuleResults (m, c) _)  -> return $ Right (m,c)
-        Left (ParseError sl msg ctxt _) ->
+        Left (ParseError sl msg ctxt) ->
             return $ Left $ classify [x | SettingClassify x <- s] $ rawIdeaN Error "Parse error" (mkSrcSpan sl sl) ctxt Nothing []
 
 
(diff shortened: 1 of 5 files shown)


src/Apply.hs
```diff
@@ -80,7 +80,7 @@ parseModuleApply flags s file src = do
     res <- parseModuleEx (parseFlagsAddFixities [x | Infix x <- s] flags) file src
     case res of
         Right (ParsedModuleResults (m, c) _)  -> return $ Right (m,c)
-        Left (ParseError sl msg ctxt _) ->
+        Left (ParseError sl msg ctxt) ->
             return $ Left $ classify [x | SettingClassify x <- s] $ rawIdeaN Error "Parse error" (mkSrcSpan sl sl) ctxt Nothing []
```
shayne-fletcher-da: Is this where the user error message is formed? Sure enough the `msg` is not used here and the result is consistent with what I observe testing ('`Parse error`').

ndmitchell:
The reason for not using the error message is that we want users to be able to match on it easily, and HLint hint ignores must textually match perfectly, so easier to match the word "Parse error". Also, the lexical errors from HSE/GHC are typically not fantastic.

shayne-fletcher-da:
We discussed in some detail the state of the error handling GHC implementation and are inclined to leave off here as "good enough" for the moment.

PR #638, opened 2019-05-16
Upgrade to ghc-lib-0.20190516

shayne-fletcher-da:
In this upgrade, `ghc-lib-parser` employs syntactic obfuscation of its cbits which will help avoid conflicts (e.g. duplicate symbol definitions) when `hlint` is linked into executables aggregating other libraries containing GHC sources. Additionally, the parser was produced by happy 1.19.10 with the `--coerece` option so parse execution times should be expected to pick up by about 10%.


ndmitchell:
Thanks Shayne!

PR #643, opened 2019-05-24
suggest null in cases equivalent to length is zero

### LINKED ISSUE #642: More warnings on trivial length comparisons?
There currently appear to be warnings on all the following cases:
* `length x == 0`
* `length x /= 0`
* `length x >= 0`
* `length x > 0`
* `length x >= 1`

... suggesting calls of `null`, `not . null`, or simply `True` accordingly.

But we find our students also writing expressions like:
* `length x < 1`
* `length x <= 0`

for which currently no warnings are given.

Open for a pull request adding such warnings? Or not considered desirable?

jvoigtlaender:
closes #642


ndmitchell:
Thanks!

PR #650, opened 2019-05-31
add another hint suggesting use of null, in a new group: teaching

jvoigtlaender:
There already was a suggestion for the case:
```haskell
if x == [] then ...
```
but not for the complementary case:
```haskell
if x /= [] then ...
```

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -732,6 +732,14 @@
     rules:
     - warn: {lhs: a $ b $ c, rhs: a . b $ c}
 
+- group:
+    name: teaching
+    enabled: false
+    imports:
+    - package base
+    rules:
+    - hint: {lhs: "x /= []", rhs: not (null x), name: Use null}
+
 
 # <TEST>
 # yes = concat . map f -- concatMap f


ndmitchell:
My concern here is that /= is a lot shorter than not null - I think this particular omission was actually intentional.

jvoigtlaender:
Okay, though I am not really thinking about shorter vs. longer in this case.

My point for using `null` over `==` on list emptiness checking is that `==` or `/=` introduce an unnecessary `Eq` constraint.

Concretely, my students work with lists of pictures (a la Gloss/CodeWorld), but also learn about polymorphic functions on lists. So when they write functions, I want them to make them as polymorphic as possible. If they have to deal with a list of pictures and check whether it is empty, then doing that with `== []` or `/= []` means that they inadvertantly introduce an `Eq` constraint where none would be needed if they used `null` or `not . null` instead. That can be problematic if they work on a type (like `[Picture]`) that has no `Eq`, or when they are meant to abstract from a type that does have `Eq`, like `[Int]`, to something polymorphic.

But granted, it is not necessarily `hlint`'s job to steer them in the right direction here. It would just be nice.

ndmitchell:
Yep, I think for a teaching course its quite reasonable to demand `null` for students, but no in general. One option would be to create a teaching section of hints in the `.yaml` file, much like the lens ones.

jvoigtlaender:
I have added a new group "teaching", which is not enabled by default, and moved the `x /= [] ==> not (null x)` hint there.

ndmitchell:
Thanks, looks great!

PR #651, opened 2019-05-31
Check for NamedFieldPuns extension

yairchu:
According to haskell-src-exts documentation NamedFieldPuns replaced RecordPuns (which is a deprecated extension name):

http://hackage.haskell.org/package/haskell-src-exts-1.21.0/docs/Language-Haskell-Exts-Extension.html#v:RecordPuns

So the check for RecordPun should also be used for NamedFieldPuns


ndmitchell:
Awesome, many thanks!

ndmitchell:
PS. It has already caught 2 issues in my own repos :)

Issue #652, opened 2019-06-02
symmetric versions of (some) hints


jvoigtlaender:
Surrounding `Eq` and `Ord` operators, there are some hints that exist in symmetric forms, e.g.:
```yaml
    - warn: {lhs: x == True, rhs: x, name: Redundant ==}
    - hint: {lhs: x == False, rhs: not x, name: Redundant ==}
    - warn: {lhs: True == a, rhs: a, name: Redundant ==}
    - hint: {lhs: False == a, rhs: not a, name: Redundant ==}
    - warn: {lhs: x == Nothing , rhs:  isNothing x}
    - warn: {lhs: Nothing == x , rhs:  isNothing x}
    - warn: {lhs: not (a >  b), rhs: a <= b, note: incorrect if either value is NaN}
    - warn: {lhs: not (a >= b), rhs: a <  b, note: incorrect if either value is NaN}
    - warn: {lhs: not (a <  b), rhs: a >= b, note: incorrect if either value is NaN}
    - warn: {lhs: not (a <= b), rhs: a >  b, note: incorrect if either value is NaN}
```
And some that don't, e.g.:
```yaml
    - warn: {lhs: compare x y /= GT, rhs: x <= y}
    - warn: {lhs: compare x y == LT, rhs: x < y}
    - warn: {lhs: compare x y /= LT, rhs: x >= y}
    - warn: {lhs: compare x y == GT, rhs: x > y}
    - warn: {lhs: compare x y == EQ, rhs: x == y}
    - warn: {lhs: compare x y /= EQ, rhs: x /= y}
    - hint: {lhs: length x == 0, rhs: null x, note: IncreasesLaziness}
```
For example, there is currently no rule matching on `0 == length x`.

I do find that students write stuff like `0 == length x`, for which I would like to give the `Use null` warning. Experienced programmers may be more likely to tend towards writing expressions like `function argument == value` instead of `value == function argument` in conditionals, which might explain why `length x == 0` is present in the hint database, but not the converse.

Anyway, short of introducing a general mechanism for specifying hints like
```yaml
    - warn: {lhs: x == True, rhs: x, name: Redundant ==}
    - hint: {lhs: x == False, rhs: not x, name: Redundant ==}
    - warn: {lhs: x == Nothing , rhs:  isNothing x}
    - warn: {lhs: not (a >  b), rhs: a <= b, note: incorrect if either value is NaN}
[...]

PR #653, opened 2019-06-03
symmetric versions of (some) hints

jvoigtlaender:
closes https://github.com/ndmitchell/hlint/issues/652


data/hlint.yaml
```diff
@@ -739,6 +739,17 @@
     - package base
     rules:
     - hint: {lhs: "x /= []", rhs: not (null x), name: Use null}
+    - hint: {lhs: "[] /= x", rhs: not (null x), name: Use null}
```
ndmitchell: This one should be in teaching since the normal version is also in teaching. I'm happy with all the others being in the normal section.

jvoigtlaender: Okay, I've kept this one there, and moved all others next to their respective original hints.

ndmitchell:
Thanks!

Issue #659, opened 2019-06-06
"Use tuple-section" not working?


jvoigtlaender:
I seem to be unable to trigger these warnings:
https://github.com/ndmitchell/hlint/blob/213b9bb678393469bbb6b27d9ebd41042b5e85f0/data/hlint.yaml#L267
https://github.com/ndmitchell/hlint/blob/213b9bb678393469bbb6b27d9ebd41042b5e85f0/data/hlint.yaml#L268
With:
```haskell
f = flip (,) 0

g = (,) 1
```
I get:
```
> hlint --language=TupleSections .\Test.hs
No hints
```

ndmitchell:
The `--language=TupleSections` should have no impact at all. For me, I get a hint for the first one, but not the second. However, the second demands `((,) 1)` since the idea is that should be `(1,)` but `(,) 1 2` should not become a tuple section. However, having just tried that expression, it isn't matched, so that's a bug in the bracket matching.

As to why it doesn't detect the first, are you using the most recent version? Any custom `.hlint.yaml` or ignores anywhere?

jvoigtlaender:
Ah, I wasn't using the most recent version, or rather the version on my laptop wasn't the same as the one used on the system that receives and checks student submissions...

So I am now seeing what you see, thanks.

The part about brackets being relevant in the matching is good to know. I was going to propose two additional hints related to tuple sections. And I now guess they have to be written like this (no _noParen_ needed on the right-hand sides, contrary to the existing tuple-section hints):
```yaml
    - warn: {lhs: "\\y -> (x,y)", rhs: "(x,)", name: Use tuple-section, note: RequiresExtension TupleSections}
    - warn: {lhs: "\\y -> (y,x)", rhs: "(,x)", name: Use tuple-section, note: RequiresExtension TupleSections}
```

ndmitchell:
Brackets aren't usually relevant, which I'm sure is why I've got them wrong in the matching engine - it usually sees through them.

Those hints look great!

ndmitchell:
I found the issues with brackets at the root, so patch pushed.

jvoigtlaender:
Cool!

PR #660, opened 2019-06-06
two additional hints for use of tuple-sections

jvoigtlaender:
As proposed in https://github.com/ndmitchell/hlint/issues/659#issuecomment-499468439.


jvoigtlaender:
The Travis failures appeared to be caused by the fact that one of the tests in the `hlint.yaml` file itself contained an occurrence of `\y -> (x,y)`. That, I fixed with the additional commit.

The AppVeyor failure, on the other hand, seems to have something to do with cabal erroring out due to dependency problems. Not something to do with this pull request specifically.

ndmitchell:
Thanks. Yep, I need to make a new release to silence the Appveyor warnings.

PR #662, opened 2019-06-06
Add failing test for brackets / slice ($x)

robx:
More of a bug report than a real PR.

hlint complains about redundant brackets in the slice `($x)`. It seems there's something about `$x` that causes it to be treated as an atom, presumably due to Template Haskell?


ndmitchell:
Thanks for the test cases. For annoying reasons it is hard to stop `($x)` triggering and keep `(($x))` triggering - but I figure the latter is fairly rare, so its OK to miss some valid bracket removals. I've pushed a patch fixing them.

PR #666, opened 2019-06-09
Add some hints to use sequenceA_ and traverse_

josephcsible:


ndmitchell:
Thanks for the contribution. I find these a bit harder to read, and prefer the more specific version. However, I appreciate that neither choice is universally held. How about moving these to the generalise section of hints in the .yaml file, then people can opt in if they want.

josephcsible:
I think generalize-for-conciseness is a better fit for these than generalize. Do you agree?

ndmitchell:
Agreed, that does seem the right place - thanks.

ndmitchell:
Thanks, yep, I agree!

PR #675, opened 2019-06-15
Complain about redundant fmaps on Eithers and Maybes

josephcsible:


data/hlint.yaml
```diff
@@ -498,12 +498,23 @@
...
+    - warn: {lhs: mapMaybe f . fmap g, rhs: mapMaybe (f . g)}
+
 
     # EITHER
 
     - warn: {lhs: "[a | Left a <- a]", rhs: lefts a}
     - warn: {lhs: "[a | Right a <- a]", rhs: rights a}
     - warn: {lhs: either Left (Right . f), rhs: fmap f}
+    - warn: {lhs: either f g . fmap h, rhs: either f (g . h)}
+    - warn: {lhs: isLeft . fmap f, rhs: isLeft, name: Redundant fmap}
```
ndmitchell: I would suggest `isLeft (fmap f x)` as if you write it in that format HLint will generate the `.` hint anyway, plus it will match the normal thing.

josephcsible:
I also added `maybe x f (fmap g y)` to `maybe x (f . g) y`.

ndmitchell:
Thanks!

PR #676, opened 2019-06-15
Only fire "Apply De Morgan law" when it's helpful

josephcsible:
If the "Apply De Morgan law" hints don't immediately result in a double negative, then they'll ultimately make the code more complicated instead of making it simpler. Make sure that at least one inner bit contains a `not` before firing them. This prevents adding complexity in the end, while still preserving the use-case in #668.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -776,6 +776,91 @@
     - hint: {lhs: "not (x || y)", rhs: "not x && not y", name: Apply De Morgan law}
     - hint: {lhs: "not (x && y)", rhs: "not x || not y", name: Apply De Morgan law}
 
+- group:
+    name: demorgan
+    enabled: true
+    imports:
+    - package base
+    rules:
+
+    # not
+
+    - hint: {lhs: "not (not x || y)", rhs: "x && not y", name: Use De Morgan law to eliminate a not}
+    - hint: {lhs: "not (x || not y)", rhs: "not x && y", name: Use De Morgan law to eliminate a not}
+    - hint: {lhs: "not (not x && y)", rhs: "x || not y", name: Use De Morgan law to eliminate a not}
+    - hint: {lhs: "not (x && not y)", rhs: "not x || y", name: Use De Morgan law to eliminate a not}
+
+    # ==, /=
+
+    - hint: {lhs: "not (a == b || y)", rhs: "a /= b && not y", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+    - hint: {lhs: "not (x || a == b)", rhs: "not x && a /= b", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+    - hint: {lhs: "not (a == b && y)", rhs: "a /= b || not y", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+    - hint: {lhs: "not (x && a == b)", rhs: "not x || a /= b", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+
+    - hint: {lhs: "not (a /= b || y)", rhs: "a == b && not y", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+    - hint: {lhs: "not (x || a /= b)", rhs: "not x && a == b", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+    - hint: {lhs: "not (a /= b && y)", rhs: "a == b || not y", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+    - hint: {lhs: "not (x && a /= b)", rhs: "not x || a == b", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+
+    # >, <=
+
+    - hint: {lhs: "not (a > b || y)", rhs: "a <= b && not y", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+    - hint: {lhs: "not (x || a > b)", rhs: "not x && a <= b", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+    - hint: {lhs: "not (a > b && y)", rhs: "a <= b || not y", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+    - hint: {lhs: "not (x && a > b)", rhs: "not x || a <= b", name: Use De Morgan law to eliminate a not, note: incorrect if either value being compared is NaN}
+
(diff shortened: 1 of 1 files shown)


ndmitchell:
For the RHS, please eliminate the `not (not x)` - no need to step through one by one. With that, I'd be happy for these to be general hints.

However, the motivation in #668 by @jvoigtlaender included things like `not (a /= b || c)` - which can still be collapsed using de Morgan, but in a more subtle way, with an indirection. I'm inclined to leave that in the teaching set, although defer to @jvoigtlaender (who is actually using it for teaching).

jvoigtlaender:
Indeed I‘m in favor of keeping the current two laws intact in the teaching section, precisely because the premise that De Morgan is only useful when it results in a double negation is wrong, e.g. „not (x < y && z == 5)“.

Another thought, for when the special cases with at least two not-occurrences are added to the general section: What will the behavior then be for expressions like „not (not x && not y)“? Will one get several competing hints? Or should there be yet another specialized version replacing that expression by „x || y“?

Maybe such expressions are not very likely to occur. But there could be competition between newly added special cases of De Morgan in the general section and the less constrained cases of De Morgan in the teaching section.

josephcsible:
This kind of makes me wish that hlint had the same level of power as GHC's rewrite rules. In particular, list fusion works by rewriting things like `map` into `build`s and `foldr`s and then looking for further rewrites on those. In particular, if no further rewrites are found, a later rewrite rule turns the `build`s and `foldr`s back into just `map`.

In general, I think the LHS of these rules are how code should look. I wish we had a way to tell hlint "rewrite the LHS to the RHS, see if you get any new hints when looking at that, and only if so should you emit this hint". A step further would be that if it saw something like `not a && not b`, and neither `not a` nor `not b` could be rewritten, then it could run just the reverse case of this and suggest `not (a || b)`.

ndmitchell:
While you could write the HLint hints to be multiple applications in one step, and then add some kind of search, you then need to have heuristics for what "looks better" etc. It's also more confusing for users. I'm very much in favour of keeping the current model, since it's served us pretty well.

I think the conclusion on this patch is that for beginners always applying demorgans is the right thing to do, so we won't change the teaching hints. I'm not averse to adding these hints in general, and then the teaching hints will need to get a bit more specific or accept overlap (which I can see is a bit vile for teaching).

jvoigtlaender:
One easy solution, at my end, to the overlap problem would be if the "teaching version" and the "general version" of these hints would use different rule names. The versions in teaching are called "Apply De Morgan law". If the versions proposed in this patch, and potentially added to the general group of hints, would be called something like "Use De Morgan law to eliminate a not", I could simply decide to not use them, thus avoiding overlap.

(Context: Students submit solutions to an online tool, get immediate feedback. One part of that feedback consists of hlint suggestions. Which ones are even active is decided when configuring tasks in the online tool, by explicitly listing, per exercise task, which hints are relevant according to the students' knowledge etc.)

jvoigtlaender:
FYI, though a bit off-topic, this is roughly the list of hints we typically use (from one of the last tasks in this semester, which is almost over, so we have largely settled on a reusable configuration after gaining some experience):
```yaml
- Apply De Morgan law
- Avoid lambda
- Avoid reverse
- Eta reduce
- Evaluate
- Fuse concatMap/map
- Fuse foldr/map
- Fuse mapMaybe/map
- Length always non-negative
- Move brackets to avoid $
- Redundant $
- Redundant /=
- Redundant ==
- Redundant bracket
- Redundant do
- Redundant flip
- Redundant fromInteger
- Redundant fromIntegral
- Redundant guard
- Redundant id
- Redundant if
- Redundant lambda
- Redundant list comprehension
- Redundant negate
- Redundant not
- Redundant pair
- Use !!
- Use &&
- Use ++
- Use /=
- Use 1
- Use :
- Use <
- Use <=
- Use ==
- Use >
- Use >=
- Use String
- Use all
- Use and
- Use any
- Use camelCase
- Use catMaybes
- Use concat
- Use concatMap
- Use const
- Use curry
- Use elem
- Use even
- Use floor
- Use foldr
- Use fromMaybe
- Use fst
- Use guards
- Use head
- Use id
- Use if
- Use infix
- Use init
- Use isJust
- Use isNothing
- Use last
- Use lefts
- Use list comprehension
- Use list literal pattern
- Use map
- Use map once
- Use mapMaybe
- Use maximum
- Use maybe
- Use minimum
- Use negate
- Use newtype instead of data
- Use notElem
- Use null
- Use odd
- Use or
- Use product
- Use repeat
- Use replicate
- Use rights
- Use snd
- Use sqrt
- Use sum
- Use tuple-section
- Use uncurry
- Use ||
- Using all on tuple
- Using and on tuple
- Using any on tuple
- Using concat on tuple
- Using elem on tuple
- Using foldr on tuple
- Using length on tuple
- Using maximum on tuple
- Using minimum on tuple
- Using null on tuple
- Using or on tuple
- Using product on tuple
- Using sum on tuple
```

josephcsible:
Okay, I updated what this PR does. Now, it adds rules for De Morgan's laws in all the cases when either there's an explicit not inside, or we know a hint that's equivalent to one. It's also easy to disable to avoid overlap with the teaching rules.

ndmitchell:
Wow, that's a lot of hints, and it seems pretty difficult to maintain. I'm wondering if there should be a specific "Boolean" hint wired in as a proper Haskell code that does this transformation?

josephcsible:
Good idea. Here's what I'm thinking: in addition to raw hints, we add a new syntax like `- opposites: [isJust, isNothing]`. This would then expand to these 10 hints:
```yaml
    - warn: {lhs: not (isNothing x), rhs: isJust x}
    - warn: {lhs: not (isJust x), rhs: isNothing x}
    - hint: {lhs: "not (isNothing x || y)", rhs: "isJust x && not y", name: Use De Morgan law to eliminate a not}
    - hint: {lhs: "not (x || isNothing y)", rhs: "not x && isJust y", name: Use De Morgan law to eliminate a not}
    - hint: {lhs: "not (isNothing x && y)", rhs: "isJust x || not y", name: Use De Morgan law to eliminate a not}
    - hint: {lhs: "not (x && isNothing y)", rhs: "not x || isJust y", name: Use De Morgan law to eliminate a not}

     - hint: {lhs: "not (isJust x || y)", rhs: "isNothing x && not y", name: Use De Morgan law to eliminate a not}
    - hint: {lhs: "not (x || isJust y)", rhs: "not x && isNothing y", name: Use De Morgan law to eliminate a not}
    - hint: {lhs: "not (isJust x && y)", rhs: "isNothing x || not y", name: Use De Morgan law to eliminate a not}
    - hint: {lhs: "not (x && isJust y)", rhs: "not x || isNothing y", name: Use De Morgan law to eliminate a not}
```
Thoughts?

ndmitchell:
Doing something like opposites is possible, but it's a fair bit more work... Inclined to turn this into an issue to fix, rather than try and deal with it in this PR. I think there's a design space to figure out.

josephcsible:
I'm inclined to agree at this point.

PR #677, opened 2019-06-15
Copy the not.null hints from teaching to generalise

josephcsible:
Using `null` instead of testing equality against `[]` has a real advantage: it lets you drop the `Eq` constraint on the contents, and operate on any `Foldable` rather than just on lists. As such, this is a generalisation useful in real programs, and not just for teaching.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -758,6 +758,8 @@
     - warn: {lhs: map, rhs: fmap}
     - warn: {lhs: a ++ b, rhs: a <> b}
     - warn: {lhs: "sequence [a]", rhs: "pure <$> a"}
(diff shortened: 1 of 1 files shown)


ndmitchell:
It's also longer, and that flexibility sometimes doesn't gain anything. I often prefer to stick to `/= []` so I'm not inclined to add it as a default hint.

josephcsible:
This is putting it in generalise, not in default.

ndmitchell:
Should they go in `teaching` _and_ `generalize`? I doubt many people have both of those turned on, but this does seem to be a hint where it's useful for both.

josephcsible:
Sure, why not.

ndmitchell:
Thanks!

Issue #682, opened 2019-06-21
Suggest explicit `DerivingStrategies` for newtypes


clementd-fretlink:
The behaviour of automatically derived instances for `newtype`s depend on the enabled extensions. Even though the process is deterministic, it is hard to know the actual behaviour just by looking at the code. As a concrete example, when deriving JSON instances, a newtype can either be seen as just a wrapper (when using `GeneralizedNewtypeDeriving`) or as a 1-field record (when using `DeriveAnyClass` and generic derivation). Errors because of this come up regularly in my team.

Thankfully, the [`DerivingStrategies`](https://gitlab.haskell.org/ghc/ghc/wikis/commentary/compiler/deriving-strategies)  extension allows to explicitly specify how the instances should be derived.

An optional hint suggesting to always put explicit strategies when deriving instances on a newtype could be quite helpful (we currently enforce it manually through code review, but automated help is always welcome).

ndmitchell:
Agreed, I had great problems with this in previous teams. I'd welcome a patch - it has to be an off-by-default hint, but a most welcome one.

clementd-fretlink:
Perfect, I'll give it a shot then.

divarvel:
I came up with a patch. It's not 100% satisfying yet, but the gist of it is there

clementd-fretlink:
Fixed by #683

Issue #681, opened 2019-06-20
RecursiveDo extension not recognised


georgefst:
I have the following .hlint.yaml:

    - arguments: [--color, --cpp-simple, -XQuasiQuotes, -XRecursiveDo]

    - extensions:
      - default: true
      - name: [PatternGuards, RecursiveDo, ViewPatterns]

But I get parse errors reported when using `mdo` (from the `RecursiveDo` extension) unless I put the language pragma in the source file.

As I understand it, explicitly telling hlint to use a particular extension _should_ enable that extension to be used across all files.

ndmitchell:
Thanks for the report! This will be a regression in newer HLint's which use the GHC parser.

@shayne-fletcher-da this is caused by GHC not taking the `ParseFlags.hseFlags.extensions` into account. Any extensions specified in that list should be added to the DynFlags, and similarly any explicitly disabled there should be removed. Do you want to take that?

Also, there should have been a test for this... My bad.

shayne-fletcher-da:
Absolutely @ndmitchell. I shall take care of it.

ndmitchell:
Fixed by #689

PR #683, opened 2019-06-21
Suggest using explicit deriving strategies for newtypes

divarvel:
Following #682, this is a first shot at suggesting explicit deriving strategies.

There are still a few questions before considering this patch satisfying

- ~how do I make it opt-in? (it makes a newtype-related test fail)~
- for the lack of a better solution, it always suggests `deriving newtype`, that could change behaviour (I don't think i can compute the actual, implicit behaviour)
- it covers gadt-like newtype definitions, I'm not 100% sure of the implications
- ~`NewTypeDerivingStrategies` is quite a long name and it's tedious for vertical alignment. Maybe you'd prefer a shorter name?~

### DIFF
--- a/src/Hint/NewType.hs
+++ b/src/Hint/NewType.hs
@@ -19,6 +19,9 @@ data A = A {b :: !C} -- newtype A = A {b :: C}
 data A = A Int#
 {-# LANGUAGE UnboxedTuples #-}; data WithAnn x = WithAnn (# Ann, x #)
 data A = A () -- newtype A = A ()
+newtype Foo = Foo Int deriving (Show, Eq) -- newtype Foo = Foo Int deriving newtype (Show, Eq)
+newtype Foo = Foo { getFoo :: Int } deriving (Show, Eq) -- newtype Foo = Foo { getFoo :: Int } deriving newtype (Show, Eq)
+newtype Foo = Foo Int deriving stock Show
 </TEST>
 -}
 module Hint.NewType (newtypeHint) where
@@ -26,7 +29,7 @@ module Hint.NewType (newtypeHint) where
 import Hint.Type
 
 newtypeHint :: DeclHint
-newtypeHint _ _ = newtypeHintDecl
+newtypeHint _ _ x = newtypeHintDecl x ++ newTypeDerivingStrategiesHintDecl x
 
 newtypeHintDecl :: Decl_ -> [Idea]
 newtypeHintDecl x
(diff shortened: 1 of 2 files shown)


divarvel:
@ndmitchell The only blocking issue IMO is about making the hint optional (it is the cause for the test failure)

src/Hint/NewTypeDerivingStrategies.hs
```diff
@@ -0,0 +1,38 @@
...
+newtype Foo = Foo Int deriving (Show, Eq) -- newtype Foo = Foo Int deriving newtype (Show, Eq)
+newtype Foo = Foo { getFoo :: Int } deriving (Show, Eq) -- newtype Foo = Foo { getFoo :: Int } deriving newtype (Show, Eq)
+newtype Foo = Foo Int deriving stock Show
+</TEST>
+-}
+module Hint.NewTypeDerivingStrategies (newTypeDerivingStrategiesHint) where
+
+import Hint.Type
+
+newTypeDerivingStrategiesHint :: DeclHint
```
ndmitchell: I'd just shove this in the Newtype package of hints - it's close enough already and not that long.

src/Hint/NewTypeDerivingStrategies.hs
```diff
@@ -0,0 +1,38 @@
...
+newTypeDerivingStrategiesHintDecl x
+    | Just new <- newtypeDecl x
+    = [suggestN "Use DerivingStrategies" x new
+      ]
+newTypeDerivingStrategiesHintDecl _ = []
+
+newtypeDecl :: Decl_ -> Maybe Decl_
+newtypeDecl (DataDecl x1 x2@(NewType _) x3 x4 x5 x6)
+    | any hasNoStrategy x6 = Just $ DataDecl x1 x2 x3 x4 x5 (withNewtype <$> x6)
+
```
ndmitchell: Please delete this blank line - otherwise it starts to look like a partial pattern match.

src/Hint/NewTypeDerivingStrategies.hs
```diff
@@ -0,0 +1,38 @@
...
+-}
+module Hint.NewTypeDerivingStrategies (newTypeDerivingStrategiesHint) where
+
+import Hint.Type
+
+newTypeDerivingStrategiesHint :: DeclHint
+newTypeDerivingStrategiesHint _ _ = newTypeDerivingStrategiesHintDecl
+
+newTypeDerivingStrategiesHintDecl :: Decl_ -> [Idea]
+newTypeDerivingStrategiesHintDecl x
```
ndmitchell:
```
newTypeDerivingStrategiesHintDecl x = [suggestN "Use DerivingStrategies" x new | Just new <- [newtypeDecl x]]
```

src/Hint/NewTypeDerivingStrategies.hs
```diff
@@ -0,0 +1,38 @@
...
+    | Just new <- newtypeDecl x
+    = [suggestN "Use DerivingStrategies" x new
```
ndmitchell: Instead of `suggestN` you need to define `ignoreN`.

src/Hint/NewTypeDerivingStrategies.hs
divarvel: Done

src/Hint/NewTypeDerivingStrategies.hs
divarvel: indeed, good catch

src/Hint/NewTypeDerivingStrategies.hs
divarvel: :ok_hand:

divarvel:
I've applied all of your suggestions, added a `ignoreN` in a separate commit and moved the code to `Hint.NewType`. There are now test failures because the expected rewrite is not proposed, plus a missing file error I'm not sure I understand.

src/Hint/All.hs
```diff
@@ -38,29 +38,30 @@ data HintBuiltin =
     HintList | HintListRec | HintMonad | HintLambda |
     HintBracket | HintNaming | HintPattern | HintImport | HintExport |
     HintPragma | HintExtensions | HintUnsafe | HintDuplicate | HintRestrict |
-    HintComment | HintNewType | HintSmell
+    HintComment | HintNewType | HintSmell | HintNewTypeDerivingStrategies
```
ndmitchell: I actually mean merge it even further so they are just under `HintNewtype` - no entry here at all.

src/Hint/NewType.hs
```diff
@@ -19,9 +19,12 @@ data A = A {b :: !C} -- newtype A = A {b :: C}
 data A = A Int#
 {-# LANGUAGE UnboxedTuples #-}; data WithAnn x = WithAnn (# Ann, x #)
 data A = A () -- newtype A = A ()
+newtype Foo = Foo Int deriving (Show, Eq) -- newtype Foo = Foo Int deriving newtype (Show, Eq)
```
ndmitchell: I think `-- @Ignore` as a prefix might make the test pass

divarvel:
It looks like tests are green! Last question, how can I try it on a real project?

src/Hint/All.hs
divarvel: :ok_hand:

src/Hint/NewType.hs
divarvel: It's not needed, the tests pass without the `@Ignore` annotation

ndmitchell:
Yep, if @ignore is not required then no need to put it in, and that is indeed how you test it.

ndmitchell:
Thanks! Looks great.

divarvel:
Thanks for your help!

ndmitchell:
No problem - planning to make a release once #681 is fixed.

PR #685, opened 2019-06-21
Speed up empty refactorings case

mitchellwrosen:
This patch causes hlint to bail earlier if it has no refactorings to apply.


ndmitchell:
Thanks

Issue #674, opened 2019-06-14
head/drop and last/take


jvoigtlaender:
Actually two issues/aspects here, but somewhat related, in that the decision about one would impact another suggestions as well.

First, there is this hint currently:
https://github.com/ndmitchell/hlint/blob/a84f0c04e417766ea46d87c2786adda77bddcdf5/data/hlint.yaml#L130

In my understanding, the `isNat` side condition is responsible for this:
```haskell
head (drop 3 list)
```
getting a suggestion to use `(!!)` instead:
```haskell
=> list !! 3
```
while this:
```haskell
head (drop n list)
```
gets no suggestion.

I would suggest the latter should also get the analogous suggestion:
```haskell
=> list !! n
```
(That is, the `isNat n` side condition in the hint description further above can go away.)

After all, `head (drop n list)` is exactly equivalent to `list !! n`. Whenever either of these expressions is defined, the other is also defined and has the same value. They are undefined on exactly the same inputs, and indeed I would argue that even in the undefined cases `list !! n` is better because it gives a more informative runtime error. Since, for `n` and `list` values for which `head (drop n list)` is undefined, the error message will be "head of empty list", whereas for the same values the error message for `list !! n` will be "index too large", which is the real problem here, even for the `head (drop n list)` expression, namely that one tries to access a list element that isn't there.

**So my first suggestion here is to remove the `isNat n` condition from the `head`/`drop` rule.**

Secondly, I have seen students doing something similar to `head`/`drop`, namely:
```haskell
last (take n list)
```
From the perspective of professional Haskell programming that is maybe surprising, but actually in teaching at a point where the construction of lists from `[]` and `(:)` has not yet been discussed, so the students do not have a specific left-leaning or right-leaning interpretation of lists yet, they have no real reason to prefer `head` over `last`. So they do write stuff like in the last snippet above.

My second suggestion is to add a rule, possibly in the teaching group, that propses to replace such expressions by
```haskell
=> list !! (n - 1)
```
And the resolution about the `isNat` side condition for `head`/`drop` might have an impact on whether the proposed `last`/`take` rule should have its own side condition or not.

jvoigtlaender:
Concerning the error message aspect: `list !! n` will also sometimes say "negative index" where `head (drop n list)` again only says a less informativ "head of empty list".

ndmitchell:
[...]

PR #686, opened 2019-06-24
add head/drop rule with max, for case that number not known to be non-negative

jvoigtlaender:
See https://github.com/ndmitchell/hlint/issues/674#issuecomment-505175812.

Closes https://github.com/ndmitchell/hlint/issues/674.

I also thought a bit about the `last`/`take` case, but there doesn't seem to be a nice formulation. It could be something like:
```yaml
    - warn: {lhs: last (take n x), rhs: x !! (min n (length x) - 1)}
```
But not only is this not really a nice expression on the right-hand side. There also is the issue that `x` gets duplicated (which I guess is problematic for hints in general).

One could also think of it in terms of the following two pseudo-hints:
```yaml
    - warn: {lhs: last (take n x), rhs: x !! (n - 1), side: n <= length x}
    - warn: {lhs: last (take n x), rhs: last x, side: n > length x}
```
Pseudo in that the `side` conditions here are not actually something that could be tested statically/syntactically, at least not in that generality. Maybe special cases would be possible for some `n` or for certain list literals `x`. But I'm doubting that would be worth it.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -128,6 +128,7 @@
     - warn: {lhs: "cycle [x]", rhs: repeat x}
     - warn: {lhs: head (reverse x), rhs: last x}
     - warn: {lhs: head (drop n x), rhs: x !! n, side: isNat n}
+    - warn: {lhs: head (drop n x), rhs: x !! max 0 n, side: not (isNat n)}
     - warn: {lhs: reverse (tail (reverse x)), rhs: init x, note: IncreasesLaziness}
     - warn: {lhs: reverse (reverse x), rhs: x, note: IncreasesLaziness, name: Avoid reverse}
     - warn: {lhs: isPrefixOf (reverse x) (reverse y), rhs: isSuffixOf x y}


ndmitchell:
Thanks! There were a couple of failures because I already had a rule `drop n x` where `n` is a negative literal, so I adjusted the rule to only match if n is not a literal.

jvoigtlaender:
Follow-up question: Where can one find these "checker functions", like `isNat`, `isNeg` etc.?

For a new rule I had in mind, I was wondering whether there is an existing checker for string literals. But I couldn't even find where `isNat` is defined, so couldn't look around whether something like `isString` exists or not.

ndmitchell:
https://github.com/ndmitchell/hlint/blob/master/src/Hint/Match.hs#L133-L181 is the definition - they're somewhat hard to find without already knowing about it, and even the code isn't that readable. `isLit` might be what you want, if the type already constraints it to being a string.

jvoigtlaender:
Ah, thanks! I was searching the repo for occurrences of `isNat`, but that identifier does not actually occur there, since it is implemented as `isType "Nat" ...` :smile:

As for my case, I don't know that the thing is already a string, so I have to check for that additionally. From the code link you gave, it looks like `isLitString` might work. But it also might not, if `String` is instead seen as `[Char]` and thus is not mateched appropriately in line
https://github.com/ndmitchell/hlint/blob/9f5776e6aee4c78a8ea0b0c543f0278da6240cf2/src/Hint/Match.hs#L165

Anyway, I'll try and see how far I get.

The rule I have in mind is certainly something to be used only locally. I wouldn't even propose it for the `teaching` section of `hlint` globally, since it actually changes the behavior of the program (ever so slightly).

The thing is that it bothers me students are writing stuff like
```haskell
  print "Give me the next input: "
  n <- readLn
```
when they really should be writing
```haskell
  putStrLn "Give me the next input: "
  n <- readLn
```
It's not a semantics preserving change, because extra quotations symbols get output in one case but not in the other.

jvoigtlaender:
Yeah, this does the trick for the case above:
```yaml
 - warn: {lhs: print x, rhs: putStrLn x, side: isLitString x}
```

It doesn't capture cases like:
```haskell
  print ("The sum of your inputs is " ++ show s)
```
which I would also like to tell students to write as
```haskell
  putStrLn ("The sum of your inputs is " ++ show s)
```
instead.

But I guess that hlint doesn't have a concept of the type of expressions, so probably by principle there is no way to catch that case while not also type-incorrectly proposing a rewrite from
```haskell
  print (a + b)
```
to
```haskell
  putStrLn (a + b)
```

But anyway, the `isLitString` case is already neat to be able to express in my local hint database. 
**I'm impressed by the versatility of hlint!**

ndmitchell:
Yep, HLint doesn't know the type, so it's incapable of expressing what you are after there. Glad you like the versatility, but it does suggest some more documentation is required...

PR #688, opened 2019-06-25
Adds eta-reduction for function match with a where clause

SarthakKumar1997:
Previously, if I uncommented the commented code, and wrote a function like this:

```haskell
fastFib n = fib' 0 1 n
  where
    fib' a _ 0 = a
    fib' a b n = fib' b (a + b) (n - 1)
```

Eta-reduction would be detected but not work.

This has now been rectified, using a `chooseLoc` function. It would be changed to:

```haskell
fastFib = fib' 0 1
  where
    fib' a _ 0 = a
    fib' a b n = fib' b (a + b) (n - 1)
```

### DIFF
--- a/src/Hint/Lambda.hs
+++ b/src/Hint/Lambda.hs
@@ -104,11 +104,10 @@ lambdaDecl (toFunBind -> o@(FunBind loc1 [Match _ name pats (UnGuardedRhs loc2 b
       [warn "Redundant lambda" o (gen pats bod) [Replace Decl (toSS o) s1 t1]]
     | length pats2 < length pats, pvars (drop (length pats2) pats) `disjoint` varss bind
         = [warn "Eta reduce" (reform pats bod) (reform pats2 bod2)
-            [ -- Disabled, see apply-refact #3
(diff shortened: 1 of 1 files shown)


ndmitchell:
There are a bunch of test failures - see https://travis-ci.org/ndmitchell/hlint/jobs/550291288 for example. Can you take a look through them and fix them (perhaps by adjusting the tests). Seems reasonable though.

ndmitchell:
CC @zliu41 - this seems like a refactoring fix. Is this one we've already got covered?

zliu41:
@ndmitchell Yeah, this has been fixed.

This PR works by expanding the srcspan of the hint to include the `where` block, which would be undesirable if the `where` block is big. It could have been improved by only expanding the srcspan of `ideaRefactoring`, but since apply-refact cannot apply refactorings with overlapping srcspans, it means if there's another hint within the `where` block, it can only be applied by running apply-refact twice. I made a fix in apply-refact which obviates the need to expand any srcspan from HLint.

ndmitchell:
Thanks!

PR #696, opened 2019-06-29
Convert NewType.hs to ghc-lib-parser

googleson78:
**Very** "initial"

I might need help/hand-holding for stuff outside this module, which is:
* Pretty-printing (which currently only seems to come from `HSE`, as far as I could see)
* Correctly propagating `newtypeHint`

All the things that need to be done are TODO'd

### DIFF
--- a/src/Hint/NewType.hs
+++ b/src/Hint/NewType.hs
@@ -34,10 +34,13 @@ newtype Foo = Foo Int deriving stock Show
 module Hint.NewType (newtypeHint) where
 
 import Hint.Type
+
+import Data.List (isSuffixOf)
 -- TODO: remove these qualifieds
 import qualified "ghc-lib-parser" HsDecls as Hs
 import qualified "ghc-lib-parser" HsSyn   as Hs
 import qualified "ghc-lib-parser" HsTypes as Hs
+import qualified "ghc-lib-parser" Outputable as Hs
 import qualified "ghc-lib-parser" SrcLoc  as Hs
 
 -- TODO: get deriving strategies to work
@@ -87,7 +90,7 @@ data WarnNewtype = WarnNewtype
     }
 
 -- | Given a declaration, returns the suggested \"newtype\"ized declaration following these guidelines:
--- * @MagicHash@'d stuff is __ignored__ - @data X = X Int#@
+-- * Types ending in a \"#\" are __ignored__, because they are usually unboxed primitives - @data X = X Int#@
 -- * @ExistentialQuantification@ stuff is __ignored__ - @data X = forall t. X t@
 -- * Constructors with (nonempty) constraints are __ignored__ - @data X a = (Eq a) => X a@
 -- * Single field constructors get newtyped - @data X = X Int@ -> @newtype X = X Int@
@@ -112,10 +115,20 @@ singleSimpleFieldNew _ = Nothing
 -- returning the type inside the constructor if it is. This is needed for bang/MagicHash analysis.
 simpleCons :: Hs.ConDecl Hs.GhcPs -> Maybe (Hs.HsType Hs.GhcPs)
 simpleCons (Hs.ConDeclH98 _ _ _ [] context (Hs.PrefixCon [Hs.L _ inType]) _)
-    | emptyOrNoContext context, not $ isUnboxedTuple inType = Just inType
+    | emptyOrNoContext context
+    , not $ isUnboxedTuple inType
+    , not $ isHashy inType
+    = Just inType
 simpleCons (Hs.ConDeclH98 _ _ _ [] context (Hs.RecCon (Hs.L _ [Hs.L _ (Hs.ConDeclField _ [_] (Hs.L _ inType) _)])) _)
-    | emptyOrNoContext context, not $ isUnboxedTuple inType = Just inType
+    | emptyOrNoContext context
+    , not $ isUnboxedTuple inType
+    , not $ isHashy inType
+    = Just inType
 simpleCons _ = Nothing
+
+isHashy :: Hs.HsType Hs.GhcPs -> Bool
+isHashy (Hs.HsTyVar _ _ id) = "#" `isSuffixOf` Hs.showSDocUnsafe (Hs.ppr id)
+isHashy x = False
 
 warnBang :: Hs.HsType Hs.GhcPs -> Bool
 warnBang (Hs.HsBangTy _ (Hs.HsSrcBang _ _ Hs.SrcStrict) _) = False
--- a/src/Hint/NewType.hs
+++ b/src/Hint/NewType.hs
@@ -7,7 +7,6 @@
 
 <TEST>
 data Foo = Foo Int -- newtype Foo = Foo Int
-data Foo where Foo :: Int -> Foo -- newtype Foo where Foo :: Int -> Foo
 data Foo = Foo Int deriving (Show, Eq) -- newtype Foo = Foo Int deriving (Show, Eq)
 data Foo = Foo { field :: Int } deriving Show -- newtype Foo = Foo { field :: Int } deriving Show
 data Foo a b = Foo a -- newtype Foo a b = Foo a
@@ -16,7 +15,6 @@ data S a = forall b . Show b => S b
 {-# LANGUAGE RankNTypes #-}; data Foo = Foo (forall a. a) -- newtype Foo = Foo (forall a. a)
 data Color a = Red a | Green a | Blue a
 data Pair a b = Pair a b
-data Pair a b where Pair :: a -> b -> Pair a b
 data Foo = Bar
 data Foo a = Eq a => MkFoo a
 data Foo a = () => Foo a -- newtype Foo a = Foo a
--- a/src/Hint/NewType.hs
+++ b/src/Hint/NewType.hs
@@ -31,47 +31,47 @@ newtype Foo = Foo Int deriving stock Show
 -}
 module Hint.NewType (newtypeHint) where
 
-import Hint.Type
+import Hint.Type (Idea, DeclHint', Note(DecreasesLaziness), ideaNote, ignoreNoSuggestion', suggestN')
 
 import Data.List (isSuffixOf)
 -- TODO: remove these qualifieds
-import qualified "ghc-lib-parser" HsDecls as Hs
-import qualified "ghc-lib-parser" HsSyn   as Hs
-import qualified "ghc-lib-parser" Outputable as Hs
-import qualified "ghc-lib-parser" SrcLoc  as Hs
+import "ghc-lib-parser" HsDecls
+import "ghc-lib-parser" HsSyn
+import "ghc-lib-parser" Outputable
+import "ghc-lib-parser" SrcLoc
 
 -- TODO: get deriving strategies to work
 newtypeHint :: DeclHint'
 newtypeHint _ _ x = newtypeHintDecl x ++ newTypeDerivingStrategiesHintDecl x
 
-newtypeHintDecl :: Hs.LHsDecl Hs.GhcPs -> [Idea]
+newtypeHintDecl :: LHsDecl GhcPs -> [Idea]
 newtypeHintDecl old
     | Just WarnNewtype{newDecl, insideType} <- singleSimpleFieldNew old
     = [(suggestN' "Use newtype instead of data" old newDecl)
             {ideaNote = [DecreasesLaziness | warnBang insideType]}]
 newtypeHintDecl _ = []
 
-newTypeDerivingStrategiesHintDecl :: Hs.LHsDecl Hs.GhcPs -> [Idea]
-newTypeDerivingStrategiesHintDecl decl@(Hs.L _ (Hs.TyClD _ (Hs.DataDecl _ _ _ _ dataDef))) =
+newTypeDerivingStrategiesHintDecl :: LHsDecl GhcPs -> [Idea]
+newTypeDerivingStrategiesHintDecl decl@(L _ (TyClD _ (DataDecl _ _ _ _ dataDef))) =
     [ignoreNoSuggestion' "Use DerivingStrategies" decl | not $ isData dataDef, not $ hasAllStrategies dataDef]
 newTypeDerivingStrategiesHintDecl _ = []
 
-hasAllStrategies :: Hs.HsDataDefn Hs.GhcPs -> Bool
-hasAllStrategies (Hs.HsDataDefn _ Hs.NewType _ _ _ _ (Hs.L _ xs)) = all hasStrategyClause xs
+hasAllStrategies :: HsDataDefn GhcPs -> Bool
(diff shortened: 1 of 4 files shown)


src/Hint/NewType.hs
```diff
@@ -65,3 +74,25 @@ hasNoStrategy _                      = False
...
+-- | Given a declaration, returns the suggested \"newtype\"ized declaration following these guidelines:
+-- * @MagicHash@'d stuff is __ignored__ - @data X = X Int#@
+-- * @ExistentialQuantification@ stuff is __ignored__ - @data X = forall t. X t@
+-- * Single field constructors get newtyped - @data X = X Int@ -> @newtype X = X Int@
+-- * Single record field constructors get newtyped - @data X = X {getX :: Int}@ -> @newtype X = X {getX :: Int}@
+-- * All other declarations are ignored.
+--
+-- TODO: insert ! warning (or lack thereof) somewhere
+singleSimpleFieldNew :: Hs.HsDecl pass -> Maybe (Hs.HsDecl pass)
+singleSimpleFieldNew (Hs.TyClD wtfIsThisSomePassStuff decl@(Hs.DataDecl _ name _ _ def@(Hs.HsDataDefn _ Hs.DataType ctx _ _ [Hs.L _ constructor] derives)))
```
googleson78: I can only guess what [`XTyClD p`](https://hackage.haskell.org/package/ghc-8.6.5/docs/HsExtension.html#t:XTyClD) from the `TyClD` constructor (and all other similar `X*` stuff) is.

ndmitchell: All the `X` things are for trees-that-grow extension points. Always just match with `_` and ignore them.

src/Hint/NewType.hs
```diff
@@ -65,3 +74,25 @@ hasNoStrategy _                      = False
...
+    | simpleCons constructor = Just $ Hs.TyClD wtfIsThisSomePassStuff decl {Hs.tcdDataDefn = def {Hs.dd_ND = Hs.NewType}}
+singleSimpleFieldNew _ = Nothing
+
+-- TODO: check for MAGIC#HASH
+-- TODO: test out existentials matching
+--
+-- TODO: eventually check GADTs too
+simpleCons :: Hs.ConDecl pass -> Bool
+simpleCons (Hs.ConDeclH98 _ _ _ [] _ (Hs.PrefixCon [_]) _) = True -- TODO: does this actually check for existentials??? supposedly the first empty list holds all the existentials
```
ndmitchell: General suggestion is if you think of a way you might have got it wrong, but there's not a test at the top of the module that would have caught you, just add a new test case.

src/Hint/NewType.hs
ndmitchell: Generally use `ext` for all `X` things, and `noExt` if you need to create one. Being polymorphic in the `pass` is usually not a good idea - the error messages get much worse and you don't gain any real flexibility - you probably `GhcPs` as `pass`.

src/Hint/NewType.hs
```diff
@@ -27,7 +28,14 @@ newtype Foo = Foo Int deriving stock Show
 module Hint.NewType (newtypeHint) where
 
 import Hint.Type
+-- TODO: remove these qualifieds
+import qualified "ghc-lib-parser" HsDecls as Hs
+import qualified "ghc-lib-parser" HsTypes as Hs
+import qualified "ghc-lib-parser" SrcLoc  as Hs
 
+-- TODO: Adapt to use ghc-lib-parser:
+-- * figure out pretty printing ???
```
ndmitchell: See https://github.com/ndmitchell/hlint/blob/master/src/Idea.hs#L90 - call `idea'` and it will pretty print for you.

src/Hint/NewType.hs
```diff
@@ -27,7 +28,14 @@ newtype Foo = Foo Int deriving stock Show
...
+-- * need to call stuff higher up to pass in the correct thing (probably HsDecl or LHsDecl)
 newtypeHint :: DeclHint
```
ndmitchell: Yep, need a NewDecl hint.

ndmitchell:
Be aware that https://github.com/ndmitchell/hlint/pull/698 is changing the newtype hints simultaneously.

src/Hint/NewType.hs
```diff
@@ -1,19 +1,24 @@
...
 data Foo = Foo { field :: Int } deriving Show -- newtype Foo = Foo { field :: Int } deriving Show
 data Foo a b = Foo a -- newtype Foo a b = Foo a
 data Foo = Foo { field1, field2 :: Int}
 data S a = forall b . Show b => S b
 data Color a = Red a | Green a | Blue a
 data Pair a b = Pair a b
+data Pair a b where Pair :: a -> b -> Pair a b
 data Foo = Bar
 data Foo a = Eq a => MkFoo a
+data Foo a = () => Foo a -- newtype Foo a = Foo a
```
googleson78: This case isn't currently covered.

This has very weird behaviour, unfortunately. It counts as `Just (L _ [])` for a context, but outputting it results in what would be `Nothing` if it were to be parsed again (`show . read` isn't `id` in other words). As a result the suggestion comes out without the `() =>` part.

It seems to be mostly a positive change (I don't know of use cases for `() =>` contexts), but I still have to ask, is this fine @ndmitchell?

src/Hint/NewType.hs
```diff
@@ -65,3 +77,44 @@ hasNoStrategy _                      = False
 withNewtype :: Deriving a -> Deriving a
 withNewtype (Deriving l Nothing rs)  = Deriving l (Just $ DerivNewtype l) rs
 withNewtype d                        = d
+
+data WarnNewtype = WarnNewtype
```
googleson78: I prefer to have a datatype with named fields for better documentation, instead of returning a tuple. Is this fine?

src/Idea.hs
```diff
@@ -97,9 +97,17 @@ idea' severity hint from to =
...
+           String -> a -> a -> [Refactoring R.SrcSpan] -> Idea
 ignore' = idea' Ignore
 
 ideaN severity hint from to = idea severity hint from to []
+ideaN' :: (GHC.HasSrcSpan a, Outputable.Outputable a) =>
+          Severity -> String -> a -> a -> Idea
+ideaN' severity hint from to = idea' severity hint from to []
 
 suggestN = ideaN Suggestion
+suggestN' :: (GHC.HasSrcSpan a, Outputable.Outputable a) =>
```
googleson78: Not sure where the `'` should go on these.

googleson78: I also added type signatures because it's very annoying to track the argument types every time. Is this fine?

googleson78:
> Be aware that #698 is changing the newtype hints simultaneously.

Thanks! I'll rebase at some point.

src/Hint/NewType.hs
```diff
@@ -65,3 +77,45 @@ hasNoStrategy _                      = False
...
+-- | Checks whether its argument is a \"simple constructor\" (see criteria in 'singleSimpleFieldNew')
+-- returning the type inside the constructor if it is. This is needed for bang/MagicHash analysis.
+simpleCons :: Hs.ConDecl Hs.GhcPs -> Maybe (Hs.HsType Hs.GhcPs)
+simpleCons (Hs.ConDeclH98 _ _ _ [] context (Hs.PrefixCon [Hs.L _ inType]) _) 
+    | emptyOrNoContext context = Just inType
+simpleCons (Hs.ConDeclH98 _ _ _ [] context (Hs.RecCon (Hs.L _ [Hs.L _ (Hs.ConDeclField _ [_] (Hs.L _ inType) _)])) _)
+    | emptyOrNoContext context = Just inType
+simpleCons _ = Nothing
+
+warnBang :: Hs.HsType Hs.GhcPs -> Bool
```
googleson78: TODO: @googleson78 This is redundant! It can easily be replaced by http://hackage.haskell.org/package/ghc-lib-parser-8.8.0.20190424/docs/HsTypes.html#v:getBangStrictness

src/Hint/NewType.hs
ndmitchell: Yep, absolutely 100% fine. Agreed, it's an improvement.

src/Hint/NewType.hs
ndmitchell: Personal preference - I would probably have opted for a tuple, but it's the kind of thing I consider whoever writing the code gets to do whatever they prefer. Neither is better or worse, so happy to go with your approach.

src/Idea.hs
ndmitchell: Agreed, I think the lack of type signatures before wasn't a great idea (it's just copying the style for `ideaN` etc, where the lack of type signatures was also a bad idea). Adding them is definitely positive.

ndmitchell: The existence of `'` versions is just "a bit like HSE, but for GHC" - i wouldn't take its placement or semantics too seriously. It probably wants cleaning up after we have finished the migration.

src/Hint/NewType.hs
googleson78: This is not redundant, we would still need to do pattern matching and such.

src/Hint/NewType.hs
```diff
@@ -65,3 +82,78 @@ hasNoStrategy _                      = False
...
+    , not $ isHashy inType
+    = Just inType
+simpleCons (Hs.ConDeclH98 _ _ _ [] context (Hs.RecCon (Hs.L _ [Hs.L _ (Hs.ConDeclField _ [_] (Hs.L _ inType) _)])) _)
+    | emptyOrNoContext context
+    , not $ isUnboxedTuple inType
+    , not $ isHashy inType
+    = Just inType
+simpleCons _ = Nothing
+
+isHashy :: Hs.HsType Hs.GhcPs -> Bool
```
googleson78: Probably not the most reliable way to print a type in general, but it *should* work here, because the `#` is always at the end (?)

googleson78: Also, `isKindHash` from `Utils` will be unused after this, I will delete it.

googleson78:
I've decided to not do #697 here, because it's a bit more complicated than I thought, and I see no benefit to trying to do both at once.

src/Hint/NewType.hs
```diff
@@ -26,39 +31,110 @@ newtype Foo = Foo Int deriving stock Show
 -}
 module Hint.NewType (newtypeHint) where
 
-import Hint.Type
+import Hint.Type (Idea, DeclHint', Note(DecreasesLaziness), ideaNote, ignoreNoSuggestion', suggestN')
 
-newtypeHint :: DeclHint
+import Data.List (isSuffixOf)
+import "ghc-lib-parser" HsDecls
```
googleson78: I de-qualified these. If you were to import the entire `Hint.Type` there would be name clashes, but this saves all the `Hs.` stuff. Is this fine?

src/Idea.hs
```diff
@@ -90,16 +90,30 @@ idea severity hint from to = rawIdea severity hint (srcInfoSpan $ ann from) (f f
 idea' :: (GHC.HasSrcSpan a, Outputable.Outputable a) =>
          Severity -> String -> a -> a -> [Refactoring R.SrcSpan] -> Idea
 idea' severity hint from to =
-  rawIdea severity hint (ghcSpanToHSE (GHC.getLoc from)) (f from) (Just $ f to) []
-  where
-    f = Outputable.showSDocUnsafe . Outputable.ppr
```
googleson78: I also needed this in `ignoreNoSuggestion'`, so I moved it out. Is this fine? (+ is it fine to always rely on `showSDocUnsafe`?)

src/Idea.hs
```diff
@@ -90,16 +90,30 @@ idea severity hint from to = rawIdea severity hint (srcInfoSpan $ ann from) (f f
...
+  rawIdea severity hint (ghcSpanToHSE (GHC.getLoc from)) (unsafePrettyPrint from) (Just $ unsafePrettyPrint to) []
 
 suggest = idea Suggestion
 warn = idea Warning
+
+ignoreNoSuggestion' :: (GHC.HasSrcSpan a, Outputable.Outputable a)
+                    => String -> a -> Idea
+ignoreNoSuggestion' hint x = rawIdeaN Ignore hint (ghcSpanToHSE (GHC.getLoc x)) (unsafePrettyPrint x) Nothing []
```
googleson78: I think this is better than directly calling `rawIdeaN` with arguments, but it's only currently used in `NewType.hs`. Is this fine? (naming?)

googleson78:
@ndmitchell 
I think this is ready to go (tests are passing)! I left some comments on uncertainties I had/have.

src/Hint/NewType.hs
```diff
@@ -26,39 +31,110 @@ newtype Foo = Foo Int deriving stock Show
...
+hasStrategyClause :: LHsDerivingClause GhcPs -> Bool
+hasStrategyClause (L _ (HsDerivingClause _ (Just _) _)) = True
+hasStrategyClause _ = False
+
+data WarnNewtype = WarnNewtype
+    { newDecl :: LHsDecl GhcPs
+    , insideType :: HsType GhcPs
+    }
+
+-- | Given a declaration, returns the suggested \"newtype\"ized declaration following these guidelines:
```
googleson78: This comment could probably go to the file header comment. What do you think?

src/Hint/NewType.hs
```diff
@@ -26,39 +31,110 @@ newtype Foo = Foo Int deriving stock Show
...
 singleSimpleField _ = Nothing
 
-newTypeDerivingStrategiesHintDecl :: Decl_ -> [Idea]
-newTypeDerivingStrategiesHintDecl x =
-    [rawIdeaN Ignore "Use DerivingStrategies" (srcInfoSpan $ ann x) (prettyPrint x) Nothing [] | lacksStrategy x]
+-- | Checks whether its argument is a \"simple constructor\" (see criteria in 'singleSimpleFieldNew')
+-- returning the type inside the constructor if it is. This is needed for bang/MagicHash analysis.
+simpleCons :: ConDecl GhcPs -> Maybe (HsType GhcPs)
+simpleCons (ConDeclH98 _ _ _ [] context (PrefixCon [L _ inType]) _)
+    | emptyOrNoContext context
```
googleson78: These guards are duplicated but I couldn't think of a good name to give them.

src/Hint/NewType.hs
ndmitchell: Within a single hint module, as long as the tests are good, I'm pretty flexible. It's very isolated.

src/Hint/NewType.hs
ndmitchell: Happy either way.

src/Hint/NewType.hs
ndmitchell: In the absence of a good name, I'd have matched with a case and defined a local `where f = ...`

src/Idea.hs
ndmitchell: It's fine - none of the names here are good, so we can refactor wholesale if we think of something better.

ndmitchell:
Thanks a lot - nice work!

PR #701, opened 2019-07-04
Add hints for replacing case with maybe

josephcsible:


josephcsible:
Hmm, the build looks like it failed because there are places in hlint itself where these new rules fire:
https://github.com/ndmitchell/hlint/blob/1b1b4189573a17b05407a7b41713c3b67caf21f3/src/Config/Yaml.hs#L126-L128

https://github.com/ndmitchell/hlint/blob/1b1b4189573a17b05407a7b41713c3b67caf21f3/src/HSE/All.hs#L200-L209

ndmitchell:
I agree its a possible hint, but is it always desirable? I would say in those two places in the HLint codebase it would make the code harder to read.

josephcsible:
The only difference from the `fromMaybe` version of this is the `f`. What if we just added `isAtom f` to them? Then it wouldn't fire in the second case.

ndmitchell:
I'm actually not that convinced `fromMaybe` gets it right... I'm tempted to say that `fromMaybe y x` is a valid replacement only if `y` is an atom. Similarly, I'd be tempted to say `maybe` is a valid replacement if `y` and `f` are atomic. I suggest we start there, which is a fairly obvious improvement from `case`, and then see if we there is anything to strengthen it from there.

Do you want to PR with those restrictions on `maybe` and `fromMaybe`?

josephcsible:
Sounds reasonable to me. I updated this PR to make the new rules have those restrictions.

ndmitchell:
Thanks! I've also followed up with something that restricts fromMaybe hint a bit too.

PR #703, opened 2019-07-05
Remove an unnecessary at-pattern that made all the shadowing going on even more confusing

josephcsible:


ndmitchell:
That was a seriously odd pattern!

PR #708, opened 2019-07-13
Convert Naming.hs to ghc-lib-parser

googleson78:
Things still left to do:

* [x] Implement shorten
* [x] Fix `replaceNames` (`uniplate` stuff) - currently stuck on this one, don't know how I can perform transformations on `ghc` identifiers, would love a pointer/guidance on how to do this
* [x] Clean the qualified imports up
* [ ] Fix the `uniplate` errors with more verbosity (kind of lost on this one)

### DIFF
--- a/src/GHC/Util.hs
+++ b/src/GHC/Util.hs
@@ -234,5 +234,9 @@ declName (ForD _ ForeignImport{fd_name}) = occNameString $ occName $ unLoc fd_na
 declName (ForD _ ForeignExport{fd_name}) = occNameString $ occName $ unLoc fd_name
 declName _ = ""
 
+-- \"Unsafely\" in this case means that it uses the following 'DynFlags' for printing -
+-- <http://hackage.haskell.org/package/ghc-lib-parser-8.8.0.20190424/docs/src/DynFlags.html#v_unsafeGlobalDynFlags unsafeGlobalDynFlags>
+-- This could lead to the issues documented there, but it also might not be a problem for our use case.
+-- TODO: Decide whether this really is unsafe, and if it is, what needs to be done to make it safe.
 unsafePrettyPrint :: (Outputable.Outputable a) => a -> String
 unsafePrettyPrint = Outputable.showSDocUnsafe . Outputable.ppr
--- a/src/GHC/Util.hs
+++ b/src/GHC/Util.hs
@@ -13,6 +13,7 @@ module GHC.Util (
   , Located
   , readExtension
   , declName
+  , unsafePrettyPrint
   -- Temporary : Export these so GHC doesn't consider them unused and
   -- tell weeder to ignore them.
   , isAtom, addParen, paren, isApp, isOpApp, isAnyApp, isDot, isSection, isDotApp
@@ -232,3 +233,6 @@ declName (SigD _ (ClassOpSig _ _ (x:_) _)) = occNameString $ occName $ unLoc x
 declName (ForD _ ForeignImport{fd_name}) = occNameString $ occName $ unLoc fd_name
 declName (ForD _ ForeignExport{fd_name}) = occNameString $ occName $ unLoc fd_name
 declName _ = ""
+
+unsafePrettyPrint :: (Outputable.Outputable a) => a -> String
+unsafePrettyPrint = Outputable.showSDocUnsafe . Outputable.ppr
--- a/src/Hint/Naming.hs
+++ b/src/Hint/Naming.hs
@@ -143,7 +143,7 @@ suggestName x
 
 replaceNamesNew :: Data a => [(String, String)] -> a -> a
 replaceNamesNew rep = transformBi replace
-    where 
+    where
         replace :: Hs.RdrName -> Hs.RdrName
         replace (Hs.Unqual (Hs.unsafePrettyPrint -> name)) = Hs.Unqual $ Hs.mkOccName Hs.srcDataName $ fromMaybe name $ lookup name rep
         replace x = x
--- a/src/Hint/Naming.hs
+++ b/src/Hint/Naming.hs
@@ -84,10 +84,11 @@ naming seen x = [suggest "Use camelCase" x' x2' [Replace Bind (toSS x) [] (prett
           x' = shorten x
           x2' = shorten x2
 
--- TODO: shorten pattern names
 shortenNew :: Hs.LHsDecl Hs.GhcPs -> Hs.LHsDecl Hs.GhcPs
 shortenNew (Hs.L locDecl (Hs.ValD ttg0 bind@(Hs.FunBind _ _ matchGroup@(Hs.MG _ (Hs.L locMatches matches) _) _ _))) =
     Hs.L locDecl (Hs.ValD ttg0 bind {Hs.fun_matches = matchGroup {Hs.mg_alts = Hs.L locMatches $ map shortenMatch matches}})
+shortenNew (Hs.L locDecl (Hs.ValD ttg0 bind@(Hs.PatBind _ _ grhss@(Hs.GRHSs _ rhss _) _))) =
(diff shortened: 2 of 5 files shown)


src/Refact.hs
```diff
@@ -11,3 +18,15 @@ toRefactSrcSpan ss = R.SrcSpan (srcSpanStartLine ss)
...
+toSrcSpan' x =
+    R.SrcSpan (Hs.srcSpanStartLine span)
+              (Hs.srcSpanStartCol span)
+              (Hs.srcSpanEndLine span)
+              (Hs.srcSpanEndCol span)
+    where
+        span :: Hs.RealSrcSpan
+        span = case Hs.getLoc x of
+                 Hs.RealSrcSpan s -> s
+                 _ -> error "Got a bad ghc SrcSpan! Report a bug to hlint please!"
```
googleson78: Is this error fine? Not sure how we can handle a non-"real" src span here.

src/GHC/Util.hs
```diff
@@ -232,3 +233,6 @@ declName (SigD _ (ClassOpSig _ _ (x:_) _)) = occNameString $ occName $ unLoc x
 declName (ForD _ ForeignImport{fd_name}) = occNameString $ occName $ unLoc fd_name
 declName (ForD _ ForeignExport{fd_name}) = occNameString $ occName $ unLoc fd_name
 declName _ = ""
+
+unsafePrettyPrint :: (Outputable.Outputable a) => a -> String
```
googleson78: Moved this here because it often seems useful. Not sure if it's a good idea, as it *is* "unsafe".

src/Hint/Naming.hs
```diff
@@ -110,3 +146,9 @@ replaceNames :: Data a => [(String,String)] -> a -> a
 replaceNames rep = transformBi f
     where f (Ident _ x) = Ident an $ fromMaybe x $ lookup x rep
           f x = x
+replaceNamesNew :: Data a => [(String,String)] -> a -> a
+replaceNamesNew rep = transformBi replace
+    where 
+        replace :: Hs.RdrName -> Hs.RdrName
+        replace (Hs.Unqual (Hs.unsafePrettyPrint -> name)) = Hs.Unqual $ Hs.mkOccName Hs.srcDataName $ fromMaybe name $ lookup name rep
```
googleson78: The part I'm unsure about!

0. Is it fine to only match on `Unqual` here?
1. Is it fine to construct `OccName`s with `srcDataName` here?

googleson78:
```
Found:
  yes_foo = yes_foo + yes_foo
Perhaps:
  yesFoo = yesFoo `+` yesFoo
```

Uh-oh! This won't actually get shown after I implement `shorten`, but it sure does seem like a bug waiting to happen!

src/Hint/Naming.hs
```diff
@@ -56,6 +85,23 @@ naming seen x = [suggest "Use camelCase" x' x2' [Replace Bind (toSS x) [] (prett
           x' = shorten x
           x2' = shorten x2
 
+shortenNew :: Hs.LHsDecl Hs.GhcPs -> Hs.LHsDecl Hs.GhcPs
+shortenNew (Hs.L locDecl (Hs.ValD ttg0 bind@(Hs.FunBind _ _ matchGroup@(Hs.MG _ (Hs.L locMatches matches) _) _ _))) =
+    Hs.L locDecl (Hs.ValD ttg0 bind {Hs.fun_matches = matchGroup {Hs.mg_alts = Hs.L locMatches $ map shortenMatch matches}})
+shortenNew (Hs.L locDecl (Hs.ValD ttg0 bind@(Hs.PatBind _ _ grhss@(Hs.GRHSs _ rhss _) _))) =
+    Hs.L locDecl (Hs.ValD ttg0 bind {Hs.pat_rhs = grhss {Hs.grhssGRHSs = map shortenLGRHS rhss}})
```
googleson78: Either I don't understand what `PatBind` is for, or it doesn't do anything at all. Even if I comment this part out, the test that I added (`yes_fooPattern Nothing = 0 -- yesFooPattern Nothing = ...`) still passes.

googleson78:
This should be ready to go.

I am however worried about [the PatBinds thing](https://github.com/ndmitchell/hlint/pull/708#discussion_r303256323), [the \`\` around operators thing](https://github.com/ndmitchell/hlint/pull/708#issuecomment-511153272) and [the Unqual thing](https://github.com/ndmitchell/hlint/pull/708#discussion_r303218622).

Could you take a look at them before merging?

googleson78:
Uh.. I see that `uniplate` prints some `error` output when run with more verbosity, but the tests pass otherwise. What have I done wrong?

src/GHC/Util.hs
ndmitchell: Would be good to know what the invariants are on it. Does it work after some step in the process? Is that step guaranteed? I'm not sure if `unsafePrettyPrint` is the right name until we've clarified if the unsafety is something we care about.

src/Refact.hs
ndmitchell: I think a fake 0 0 0 0 span is the only viable option - I'd much rather not crash here, since if we get here, odds are it's a GHC bug we inherited, or someone has preprocessed their syntax tree, and a report to HLint isn't going to achieve much.

ndmitchell:
Sorry, I've been away from computer, and will continue to be only sporadically available until the weekend.

I see the Uniplate error message. That means that one of the fields in GHC is failing to be "faked up" by Uniplate. When that happens you get a 3x slowdown in Uniplate operations, so that's very important to fix. It's also pretty low level in Uniplate, so probably one that has to be left for me.

Excluding that, do all the tests pass locally? If so, the code looks good, and I can merge once Uniplate is sorted.

src/GHC/Util.hs
googleson78: I'm not sure about this one.

Looking at the docs it uses [`unsafeGlobalDynFlags`](http://hackage.haskell.org/package/ghc-lib-parser-8.8.0.20190424/docs/src/DynFlags.html#v_unsafeGlobalDynFlags):
```
-- Do not use unsafeGlobalDynFlags!
--
-- unsafeGlobalDynFlags is a hack, necessary because we need to be able
-- to show SDocs when tracing, but we don't always have DynFlags
-- available.
--
-- Do not use it if you can help it. You may get the wrong value, or this
-- panic!
```

To me it seems that the things in there shouldn't affect us in this case, however I really don't know enough about `ghc` to make this call.
I have the following two ideas in mind if this needs changing:
1. Can make this use `GHC.Util.baseDynFlags`.
2. Use `GHC.Util.parsePragmasIntoDynFlags` and pass the flags down in some form to the pretty printer.

If you're fine with this, let's fix this in another issue/pr - it works for now, and I see no benefit to doing both things at once. (I will add a TODO-like comment to this though)

googleson78:
> Sorry, I've been away from computer, and will continue to be only sporadically available until the weekend.

No problem, thanks for your time!

> Excluding that, do all the tests pass locally? If so, the code looks good, and I can merge once Uniplate is sorted.

Yes, `hlint test` passes locally.

ndmitchell:
I decided in the end that its easier to merge this, then restore UNIPLATE_VERBOSE after.

src/GHC/Util.hs
shayne-fletcher-da: One argument for us attempting to arrange for "safe" pretty prints with the dyn flags used for parsing is that some pretty-prints need to take extensions into account e.g. `ImportQualifiedPost`.

ndmitchell: Do we expect to actually honour that? My guess is no, even if the extension is set, since it adds a lot of complexity for relatively minor benefit.

shayne-fletcher-da: Nah, I'd be casual about that. If the situation changes significantly we can revisit.

PR #710, opened 2019-07-17
declName returns Maybe; Clarify declName logic with a comment

### LINKED ISSUE #706: declName to return Maybe String + comment on what the function does
Looking at the function `GHC.Util.declName` it seems to only return names for declarations that you would want to report (aka declarations that create new binders, not ones that refer to already existing binders).

What do you think about changing its return type to `Maybe String`, to make this more explicit, and adding a comment explaining the logic behind the function?

(of course I volunteer to do this if you agree)

googleson78:
Fixes #706

### DIFF
--- a/src/Apply.hs
+++ b/src/Apply.hs
@@ -52,7 +52,7 @@ applyHintsReal settings hints_ ms = concat $
     [ map (classify classifiers . removeRequiresExtensionNotes (hseModule m)) $
         order [] (hintModule hints settings nm m) `merge`
         concat [order [fromNamed d] $ decHints d | d <- moduleDecls (hseModule m)] `merge`
-        concat [order [declName $ GHC.unLoc d] $ decHints' d | d <- hsmodDecls $ GHC.unLoc $ ghcModule m]
+        concat [order (maybeToList $ declName $ GHC.unLoc d) $ decHints' d | d <- hsmodDecls $ GHC.unLoc $ ghcModule m]
     | (nm, m) <- mns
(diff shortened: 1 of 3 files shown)


src/Apply.hs
```diff
@@ -52,7 +52,7 @@ applyHintsReal settings hints_ ms = concat $
     [ map (classify classifiers . removeRequiresExtensionNotes (hseModule m)) $
         order [] (hintModule hints settings nm m) `merge`
         concat [order [fromNamed d] $ decHints d | d <- moduleDecls (hseModule m)] `merge`
-        concat [order [declName $ GHC.unLoc d] $ decHints' d | d <- hsmodDecls $ GHC.unLoc $ ghcModule m]
+        concat [order (maybeToList $ declName $ GHC.unLoc d) $ decHints' d | d <- hsmodDecls $ GHC.unLoc $ ghcModule m]
```
googleson78: Well.. this is different compared to before (`[""]` vs `[]`), but the tests seem to pass, so it should be fine? Or do I have a false sense of security here?

googleson78:
I just pushed a slight change which reduces code duplication.
Don't know what happened with [the appveyor build](https://ci.appveyor.com/project/ndmitchell/hlint/builds/26426889) but I don't think I've broken it with this.
@ndmitchell any opinions on this?

ndmitchell:
Thanks! Sorry I missed this earlier (was on holiday and still catching up with PR's)

ndmitchell:
Btw, Appveyor is out of commission because we are waiting for a Stackage update, and for bad reasons HLint got blocked behind hsdev...

Issue #713, opened 2019-07-22
'pattern' identifier causes a parse error despite -XNoPatternSynonyms in .hlint.yaml


23Skidoo:
So there's an identifier called `pattern` in `Data.Swagger`. With HLint 2.2, it causes a parse error:
```
Error: Parse error
Found:
    [...]
  > import Data.Swagger (NamedSchema(..), ToSchema(..), SwaggerType(..), pattern, type_)
    [..]
```

I have the following in `.hlint.yaml`:
```
- arguments: [ -XTypeApplications
  , -XNoPatternSynonyms # There's an identifier named "pattern" in Data.Swagger.
  ]
[...]
```

This fixes the issue with HLint 2.1.22, but not with 2.2.

ndmitchell:
@shayne-fletcher-da - I suspect this is because on the GHC path we don't interpret `No` as a prefix of an extension to disable it. Can you take a look?

shayne-fletcher-da:
Hmmm. `ghcExtensionsFromParseFlags` in `All.hs` looks like it's trying to get this right. Yes, I'll look into it.

shayne-fletcher-da:
It's not on the GHC path. It's actually a HSE parse failure.

shayne-fletcher-da:
Ok got it. PR coming.

ndmitchell:
Fixed in #714

ndmitchell:
Released as v2.2.1. Thanks @shayne-fletcher-da !

23Skidoo:
Thanks for fixing it so fast!

PR #714, opened 2019-07-22
Do not default enable extensions for GHC

shayne-fletcher-da:
This PR fixes setting enabled/disabled language extensions for GHC parsing. The previous code was relying on assumptions about HSE flags parsing that turned out to be wrong.


shayne-fletcher-da:
Fixes https://github.com/ndmitchell/hlint/issues/713.

ndmitchell:
Thanks for the great comment, it answered all the questions I had about this code!

PR #718, opened 2019-07-24
Add missing IncreasesLaziness

josephcsible:
`fromJust (fmap (const "Foo") Nothing)` is ⊥, but `const "Foo" (fromJust Nothing)` is `"Foo"`.


ndmitchell:
Thanks! How did you manage to spot this? Took me at least 10 minutes
Of head scratching to even decide I agreed with you. Nice catch!

josephcsible:
I have no idea; I was just staring at hlint.yaml for something else, and that happened to catch my attention.

ndmitchell:
Cool, nice spot!

PR #720, opened 2019-07-29
Expose 'readFilesConfig' and 'splitSettings' from HLint4

shayne-fletcher-da:
Exposing `readFilesConfig` and `splitSettings` helps us out in DAML land.

### DIFF
--- a/src/Language/Haskell/HLint4.hs
+++ b/src/Language/Haskell/HLint4.hs
@@ -19,7 +19,7 @@ module Language.Haskell.HLint4(
     -- * Settings
     Classify(..),
     getHLintDataDir, autoSettings, argsSettings,
-    findSettings, readSettingsFile,
+    findSettings, readSettingsFile, readFilesConfig, splitSettings,
     -- * Hints
     Hint, resolveHints,
     -- * Parse files


ndmitchell:
Not against it if it helps, but I would have thought the existing API is better for the kind of things I expect you to do. I'd also rather avoid exposing `Setting` as a type from HLint API at all, if possible. I would imagine you just call `findSettings` repeatedly and then `uncurry3` on the results and concat them together. I imagine:

```
findSettings (const ("file.yaml", contents)) (Just "file.yaml")
```

Would do it for you. Am I wrong.

shayne-fletcher-da:
I don't know what's going on with stackage : `ghc-lib-parser-8.8.0.20190723` has been on hackage for a week now 😕 Anyway, that's why the appveyor build is failing.

shayne-fletcher-da:
Ah well, I don't know if I can rephrase. You'd be the quicker judge. I'm doing what the hlint exe does (to enable per-project config for the IDE):
```yaml
hlintSettings :: FilePath -> IO ([Classify], Hint)
hlintSettings hlintDataDir = do
  curdir <- getCurrentDirectory
  home <- catchIOError ((:[]) <$> getHomeDirectory) (const $ return [])
  implicit <-
    findM System.Directory.Extra.doesFileExist $
      map (</> ".dlint.yaml") (ancestors curdir ++ home)
  let settingsFiles = map (, Nothing) $
        (hlintDataDir </> "hlint.yaml") : maybeToList implicit
  -- `readFilesConfig` calls `readFileConfigYaml` which in turn calls
  -- `decodeFileEither` from the `yaml` library.  Annoyingly that
  -- function catches async exceptions and in particular, it ends up
  -- catching `ThreadKilled`. So, we have to mask to stop it from
  -- doing that.
  (_, classify, hints) <- splitSettings <$>
    (mask $ \unmask -> (unmask . readFilesConfig) settingsFiles)
  return (classify, hints)
  where
    ancestors = init . map joinPath . reverse . inits . splitPath
```

ndmitchell:
ghc-lib-parser-8.8.0.20190424 is still the one in Stackage because hsdev is holding up the HLint upgrade. See https://github.com/commercialhaskell/stackage/issues/4673

shayne-fletcher-da:
> commercialhaskell/stackage#4673

Ooh. Thanks!

ndmitchell:
Yep, I think you can replace:

```
  (_, classify, hints) <- splitSettings <$>
    (mask $ \unmask -> (unmask . readFilesConfig) settingsFiles)
```

With calls to `findSettings` for each file you have. `readFileConfig file contents` would be roughly `findSettings (const (file, contents)) (Just file)` and putting them together should be pretty easy.

I wouldn't be averse to a more direct function that was the composition of splitSettings and readFilesConfig, as I appreciate the `findSettings` is a bit obtuse. By combining them you can avoid leaking the `Setting` type.

shayne-fletcher-da:
Right you are. I'll have a look at this in the morning and when I've got it working, will withdraw this PR!

shayne-fletcher-da:
Yep, this strategy works. Thanks!

PR #721, opened 2019-07-31
Add ignoreN'

shayne-fletcher-da:
Tiny PR. Just adds `ignoreN'` to `Idea.hs` which was missing before now ~and I have a need of it~.


ndmitchell:
Thanks!

PR #724, opened 2019-07-31
Replace some disabled arrow hints

josephcsible:
Most of our arrow hints are set to `ignore` (reasonably so), and the arrow methods they use have non-arrow equivalents in these cases. Replace these hints with their functor/bifunctor/applicative equivalents.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -15,6 +15,7 @@
     - import Data.Foldable(asum, sequenceA_, traverse_, for_)
     - import Data.Traversable(traverse, for)
     - import Control.Applicative
+    - import Data.Bifunctor
     - import Data.Function
     - import Data.Int
     - import Data.Char
@@ -335,12 +336,15 @@
     - warn: {lhs: id *** g, rhs: second g}
     - warn: {lhs: f *** id, rhs: first f}
     - ignore: {lhs: zip (map f x) (map g x), rhs: map (f Control.Arrow.&&& g) x}
-    - ignore: {lhs: "\\(x,y) -> (f x, g y)", rhs: f Control.Arrow.*** g}
     - ignore: {lhs: "\\x -> (f x, g x)", rhs: f Control.Arrow.&&& g}
-    - ignore: {lhs: "\\(x,y) -> (f x,y)", rhs: Control.Arrow.first f}
-    - ignore: {lhs: "\\(x,y) -> (x,f y)", rhs: Control.Arrow.second f}
-    - ignore: {lhs: "(f (fst x), g (snd x))", rhs: (f Control.Arrow.*** g) x}
     - hint: {lhs: "(fst x, snd x)", rhs:  x, note: DecreasesLaziness, name: Redundant pair}
+
+    # BIFUNCTOR
+
+    - hint: {lhs: "\\(x,y) -> (f x, g y)", rhs: Data.Bifunctor.bimap f g, note: IncreasesLaziness}
+    - hint: {lhs: "\\(x,y) -> (f x,y)", rhs: Data.Bifunctor.first f, note: IncreasesLaziness}
+    - hint: {lhs: "\\(x,y) -> (x,f y)", rhs: Data.Bifunctor.second f, note: IncreasesLaziness}
+    - hint: {lhs: "(f (fst x), g (snd x))", rhs: Data.Bifunctor.bimap f g x}
 
     # FUNCTOR
 


data/hlint.yaml
```diff
@@ -334,14 +335,14 @@
...
-    - ignore: {lhs: "\\(x,y) -> (f x, g y)", rhs: f Control.Arrow.*** g}
-    - ignore: {lhs: "\\x -> (f x, g x)", rhs: f Control.Arrow.&&& g}
-    - ignore: {lhs: "\\(x,y) -> (f x,y)", rhs: Control.Arrow.first f}
-    - ignore: {lhs: "\\(x,y) -> (x,f y)", rhs: Control.Arrow.second f}
-    - ignore: {lhs: "(f (fst x), g (snd x))", rhs: (f Control.Arrow.*** g) x}
     - hint: {lhs: "(fst x, snd x)", rhs:  x, note: DecreasesLaziness, name: Redundant pair}
 
+    # BIFUNCTOR
+
+    - hint: {lhs: "\\(x,y) -> (f x, g y)", rhs: bimap f g, note: IncreasesLaziness}
```
ndmitchell: Can you make sure it says `Data.Bifunctor.bimap` - that way HLint will be able to suggest the module name too. Ditto for the others below.

data/hlint.yaml
```diff
@@ -367,6 +368,8 @@
     - hint: {lhs: pure x <*> y, rhs: x <$> y}
     - warn: {lhs: x <* pure y, rhs: x}
     - warn: {lhs: pure x *> y, rhs: "y"}
+    - hint: {lhs: zip (map f x) (map g x), rhs: map (liftA2 (,) f g) x}
```
ndmitchell: Not convinced on this one.

data/hlint.yaml
```diff
@@ -367,6 +368,8 @@
...
+    - hint: {lhs: "\\x -> (f x, g x)", rhs: liftA2 (,) f g}
```
ndmitchell: Not convinced either.

data/hlint.yaml
```diff
@@ -334,14 +335,14 @@
 
     - warn: {lhs: id *** g, rhs: second g}
     - warn: {lhs: f *** id, rhs: first f}
-    - ignore: {lhs: zip (map f x) (map g x), rhs: map (f Control.Arrow.&&& g) x}
-    - ignore: {lhs: "\\(x,y) -> (f x, g y)", rhs: f Control.Arrow.*** g}
-    - ignore: {lhs: "\\x -> (f x, g x)", rhs: f Control.Arrow.&&& g}
-    - ignore: {lhs: "\\(x,y) -> (f x,y)", rhs: Control.Arrow.first f}
-    - ignore: {lhs: "\\(x,y) -> (x,f y)", rhs: Control.Arrow.second f}
-    - ignore: {lhs: "(f (fst x), g (snd x))", rhs: (f Control.Arrow.*** g) x}
```
ndmitchell: Good idea!

data/hlint.yaml
```diff
@@ -474,6 +477,7 @@
 
     - warn: {lhs: fst (unzip x), rhs: map fst x}
     - warn: {lhs: snd (unzip x), rhs: map snd x}
+    - hint: {lhs: "\\(x,y) -> (x,f y)", rhs: fmap f}
```
ndmitchell: Not totally convinced  by this. I'd prefer `Bifunctor.second` over fmap, since that's clearer.

josephcsible:
Data.Bifunctor existed since base version 4.8.0.0, which corresponds to GHC 7.10, released over 4 years ago.

data/hlint.yaml
josephcsible: Does that just need to be changed in the hints themselves, or do I need to change the import to qualified too?

data/hlint.yaml
josephcsible: Yeah, `&&&` is the one thing arrows offer without a nice clean replacement. Instead of `liftA2 (,) f g`, we could always do `(,) <$> f <*> g`, but I feel like you'd dislike that even more. I may just remove these ones.

data/hlint.yaml
josephcsible: Is it really clearer? It's dragging in Bifunctor, which is much less common/well-known than Functor.

data/hlint.yaml
ndmitchell: Just the hints themselves.

data/hlint.yaml
ndmitchell: Yeah, I think saying nothing is probably best.

data/hlint.yaml
ndmitchell: I'm inclined to say because neither is obviously clearer to leave it out entirely. I think the functor instance of `(,)` comes up when people are thinking tricky, whereas for Bifunctor the most common type is `(,)` by far.

In truth, I'd use https://hackage.haskell.org/package/extra-1.6.17/docs/Data-Tuple-Extra.html#v:second which is restricted to tuples - which I think is a good idea - but want to stick to the base library for hints.

ndmitchell:
I hadn't realised Data.Bifunctor was so old! That's fine to hint towards.

ndmitchell:
Thanks. There were a couple of failing tests, but I've merged and fixed those up - thanks!

PR #726, opened 2019-08-02
Convert duplicates to ghc-lib

shayne-fletcher-da:
Convert "Reduce duplication" hints to GHC. Also adds some more hitherto missing top-level sigs ~and performs one renaming `declName` => `nameOfDecl`~.

### DIFF
--- a/src/GHC/Util.hs
+++ b/src/GHC/Util.hs
@@ -13,11 +13,13 @@ module GHC.Util (
   , Located
   , readExtension
   , commentText, isCommentMultiline
-  , declName
+  , nameOfDecl, nameOfModule
   , unsafePrettyPrint
   , eqMaybe
   , noloc, unloc, getloc, noext
   , isForD
+  , W(..)
+  , SrcSpanD(..)
   -- Temporary : Export these so GHC doesn't consider them unused and
   -- tell weeder to ignore them.
   , isAtom, addParen, paren, isApp, isOpApp, isAnyApp, isDot, isSection, isDotApp
@@ -43,11 +45,14 @@ import "ghc-lib-parser" Panic
 import "ghc-lib-parser" HscTypes
 import "ghc-lib-parser" HeaderInfo
 import "ghc-lib-parser" ApiAnnotation
+import "ghc-lib-parser" Module
 
 import Data.List.Extra
+import Data.Function
 import System.FilePath
 import Language.Preprocessor.Unlit
 import qualified Data.Map.Strict as Map
+import Data.Default
 
 fakeSettings :: Settings
 fakeSettings = Settings
@@ -241,20 +246,24 @@ isCommentMultiline :: Located AnnotationComment -> Bool
 isCommentMultiline (L _ (AnnBlockComment _)) = True
 isCommentMultiline _ = False
 
-declName :: HsDecl GhcPs -> String
-declName (TyClD _ FamDecl{tcdFam=FamilyDecl{fdLName}}) = occNameString $ occName $ unLoc fdLName
-declName (TyClD _ SynDecl{tcdLName}) = occNameString $ occName $ unLoc tcdLName
-declName (TyClD _ DataDecl{tcdLName}) = occNameString $ occName $ unLoc tcdLName
-declName (TyClD _ ClassDecl{tcdLName}) = occNameString $ occName $ unLoc tcdLName
-declName (ValD _ FunBind{fun_id})  = occNameString $ occName $ unLoc fun_id
-declName (ValD _ VarBind{var_id})  = occNameString $ occName var_id
-declName (ValD _ (PatSynBind _ PSB{psb_id})) = occNameString $ occName $ unLoc psb_id
-declName (SigD _ (TypeSig _ (x:_) _)) = occNameString $ occName $ unLoc x
-declName (SigD _ (PatSynSig _ (x:_) _)) = occNameString $ occName $ unLoc x
-declName (SigD _ (ClassOpSig _ _ (x:_) _)) = occNameString $ occName $ unLoc x
-declName (ForD _ ForeignImport{fd_name}) = occNameString $ occName $ unLoc fd_name
-declName (ForD _ ForeignExport{fd_name}) = occNameString $ occName $ unLoc fd_name
-declName _ = ""
+nameOfDecl :: HsDecl GhcPs -> String
+nameOfDecl (TyClD _ FamDecl{tcdFam=FamilyDecl{fdLName}}) = occNameString $ occName $ unLoc fdLName
+nameOfDecl (TyClD _ SynDecl{tcdLName}) = occNameString $ occName $ unLoc tcdLName
+nameOfDecl (TyClD _ DataDecl{tcdLName}) = occNameString $ occName $ unLoc tcdLName
+nameOfDecl (TyClD _ ClassDecl{tcdLName}) = occNameString $ occName $ unLoc tcdLName
+nameOfDecl (ValD _ FunBind{fun_id})  = occNameString $ occName $ unLoc fun_id
+nameOfDecl (ValD _ VarBind{var_id})  = occNameString $ occName var_id
+nameOfDecl (ValD _ (PatSynBind _ PSB{psb_id})) = occNameString $ occName $ unLoc psb_id
+nameOfDecl (SigD _ (TypeSig _ (x:_) _)) = occNameString $ occName $ unLoc x
+nameOfDecl (SigD _ (PatSynSig _ (x:_) _)) = occNameString $ occName $ unLoc x
+nameOfDecl (SigD _ (ClassOpSig _ _ (x:_) _)) = occNameString $ occName $ unLoc x
+nameOfDecl (ForD _ ForeignImport{fd_name}) = occNameString $ occName $ unLoc fd_name
+nameOfDecl (ForD _ ForeignExport{fd_name}) = occNameString $ occName $ unLoc fd_name
+nameOfDecl _ = ""
+
+nameOfModule :: HsModule GhcPs -> String
+nameOfModule HsModule {hsmodName=Nothing} = "Main"
+nameOfModule HsModule {hsmodName=Just (L _ n)} = moduleNameString n
 
 -- \"Unsafely\" in this case means that it uses the following
 -- 'DynFlags' for printing -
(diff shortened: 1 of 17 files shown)


shayne-fletcher-da:
The diff of `Duplicates.hs` really belies the effort it took for this one 😉

src/GHC/Util.hs
```diff
@@ -13,11 +13,13 @@ module GHC.Util (
   , Located
   , readExtension
   , commentText, isCommentMultiline
-  , declName
+  , nameOfDecl, nameOfModule
```
ndmitchell: These seem very OCaml-style names for the things. I think declName and moduleName would be the more Haskelly names.

src/GHC/Util.hs
```diff
@@ -288,3 +297,25 @@ getloc = getLoc
 
 noext :: NoExt
 noext = noExt
+
+--
+
+newtype SrcSpanD = SrcSpanD SrcSpan -- SrcSpan, deriving `Default`.
+instance Default SrcSpanD where def = SrcSpanD noSrcSpan
+instance Outputable SrcSpanD where ppr (SrcSpanD s) = ppr s
+instance Eq SrcSpanD where (==) (SrcSpanD a) (SrcSpanD b) = a == b
```
ndmitchell: Can this use `GeneralizedNewtypeDeriving`? Should work for all but Default.

src/GHC/Util.hs
```diff
@@ -288,3 +297,25 @@ getloc = getLoc
...
+-- The issue is that at times, terms we work with in this program are
+-- not in `Eq` and `Ord` and we need them to be. This work-around
+-- resorts to implementing `Eq` and `Ord` for the these types via
+-- lexicographical comparisons of string representations. As long as
+-- two different terms never map to the same string representation,
+-- basing `Eq` and `Ord` on their string representations rather than
+-- the term types themselves, leads to identical results.
+wToStr :: Outputable a => W a -> String
+wToStr (W e) = showPpr baseDynFlags e
+instance (Outputable a) => Outputable (W a) where ppr (W e) = ppr e
```
ndmitchell: Use newtype deriving here.

src/GHC/Util.hs
```diff
@@ -288,3 +297,25 @@ getloc = getLoc
...
+
+--
+
+newtype SrcSpanD = SrcSpanD SrcSpan -- SrcSpan, deriving `Default`.
+instance Default SrcSpanD where def = SrcSpanD noSrcSpan
+instance Outputable SrcSpanD where ppr (SrcSpanD s) = ppr s
+instance Eq SrcSpanD where (==) (SrcSpanD a) (SrcSpanD b) = a == b
+instance Ord SrcSpanD where compare (SrcSpanD a) (SrcSpanD b) = compare a b
+
+newtype W a = W a -- Wrapper of terms.
```
ndmitchell: Not sure I like that name `W`. Any better ideas? I don't have anything immediate...

src/Hint/ListRec.hs
```diff
@@ -51,6 +51,7 @@ listRecHint _ _ = concatMap f . universe
             return $ idea severity ("Use " ++ use) o y [Replace Decl (toSS o) [] (prettyPrint y)]
 
 
+recursiveStr :: String
```
ndmitchell: Not sure this one is worth it. Don't care overly though.

src/Hint/Monad.hs
```diff
@@ -64,7 +64,9 @@ import qualified Refact.Types as R
 import Prelude
 
 
+badFuncs :: [String]
```
ndmitchell: Ditto, not sure these are worth it.

src/Test/All.hs
```diff
@@ -52,14 +52,16 @@ test CmdTest{..} main dataDir files = do
...
         when cmdQuickCheck $ wrap "Hint QuickChecking" $
-            progress >> testQuickCheck cmdDataDir cmdTempDir [h | (file, h) <- testFiles, takeFileName file /= "Test.hs"]
+            progress >> testQuickCheck cmdDataDir cmdTempDir (hs testFiles)
 
         when (null files && not hasSrc) $ liftIO $ putStrLn "Warning, couldn't find source code, so non-hint tests skipped"
         getIdeas
     whenLoud $ mapM_ print ideas
     return failures
+    where
+      hs testFiles = [h | (file, h) <- testFiles, takeFileName file /= "Test.hs"]
```
ndmitchell: Did this get caught by the duplicate check? If so, did you make the duplicate check more powerful? Rather than generalise this why, why not `let hs = ...` before `cmdTypeCheck` and then use it twice? That avoids duplicate compilation as well as duplicate code.

src/Hint/Duplicate.hs
```diff
@@ -70,9 +89,13 @@ add pos (v:vs) (Dupe p mp) = Dupe p $ Map.insertWith f v (add pos vs $ Dupe pos
 duplicateOrdered :: (Ord pos, Default pos, Ord val) => Int -> [[(pos,val)]] -> [(pos,pos,[val])]
```
ndmitchell: If you do `forall pos val .` on this line, then you can avoid restating the `Ord`/`Default` constraints on the lines below. You might need to turn on `ScopedTypeVariables`.

src/GHC/Util.hs
shayne-fletcher-da: `moduleName` has a clash with HSE, so I went with `nameOfModule` and if we are doing `nameOfModule` then `nameOfDecl` is changed for consistency.

src/GHC/Util.hs
shayne-fletcher-da: I'd have to research that. Perhaps you can give me a pointer?

src/GHC/Util.hs
shayne-fletcher-da: Sorry - can you elaborate please?

src/GHC/Util.hs
ndmitchell: Can we do `modName` or `ghcModuleName`? (And then remove the `ghc` prefix in a future state)

src/GHC/Util.hs
shayne-fletcher-da: I went through various incantations involving `Wrap`. In the end I forsee this as being somewhat ubiquitous and reason short is good.

src/Hint/ListRec.hs
shayne-fletcher-da: I'm trying to get it clean of warnings with `-Wmissing-sigs` enabled.

src/Hint/Monad.hs
shayne-fletcher-da: Again, I'm linting.

src/GHC/Util.hs
ndmitchell: `newtype SrcSpanD = SrcSpanD SrcSpan deriving (Outputable, Eq, Ord)` and add `GeneralizedNewtypeDeriving` at the top of the file when the compiler yells at you.

ndmitchell:
You just conflicted with another PR I just merged - but this one is super trivial.

src/Test/All.hs
shayne-fletcher-da: It got caught by the duplicate check. Evidently, yes, it got more powerful (not by design). I can change particular phrase along these lines this if you think it important.

src/Hint/Duplicate.hs
shayne-fletcher-da: Ok. Thanks for the hint.

src/GHC/Util.hs
ndmitchell: See 10 lines up.

src/GHC/Util.hs
ndmitchell: OK, fair enough.

shayne-fletcher-da:
You are killing me with these conflicts :)

src/Hint/ListRec.hs
ndmitchell: Will you ever do that with the huge number of `isFoo` methods in the HSE.Utils?

ndmitchell:
I promise you're the next one to merge! I need to keep my merge queue shorter...

src/Test/All.hs
ndmitchell: I think this instance is worth changing. Interesting to know it got a bit more powerful, but as long as the tests still pass, this instance is not an unreasonable thing to say has too much duplication.

shayne-fletcher-da:
Conflicts are resolved. If you don't mind (and judge it good enough for now), land it and I'll apply whatever you deem necessary based on the above comments in a follow up. I'm going to the pub!

src/GHC/Util.hs
shayne-fletcher-da: Yes, we can do that.

src/GHC/Util.hs
shayne-fletcher-da: Ok!

src/Hint/ListRec.hs
shayne-fletcher-da: Heh :) Sounds like I've created a switch to beat myself with. I guess I'll desist then.

src/GHC/Util.hs
shayne-fletcher-da: This is done.

src/Test/All.hs
shayne-fletcher-da: Fixed

src/Hint/Duplicate.hs
shayne-fletcher-da: Done.

src/GHC/Util.hs
shayne-fletcher-da: ✔️

src/GHC/Util.hs
shayne-fletcher-da: ✔️

shayne-fletcher-da:
I'm back from the pub and have made the requested changes. Good to go.

src/GHC/Util.hs
```diff
@@ -241,6 +247,26 @@ isCommentMultiline :: Located AnnotationComment -> Bool
 isCommentMultiline (L _ (AnnBlockComment _)) = True
 isCommentMultiline _ = False
 
+declName :: HsDecl GhcPs -> Maybe String
```
googleson78: RIP the comment? :(

googleson78:
Did you decide to not rename `declName` to `nameOfDecl`?

shayne-fletcher-da:
> Did you decide to not rename `declName` to `nameOfDecl`?

Yes! `declName` has been restored 😄

src/GHC/Util.hs
shayne-fletcher-da: Sorry @googleson78 something wrong? What do you mean here?

googleson78: In #710 I added a bit of a comment to the function. It seems to be red in the diff for `Utils.hs` here? I'm guessing something while merging/rebasing?

googleson78: https://github.com/ndmitchell/hlint/pull/726/files#diff-fbbb414b66284e60d388bd81dc2b606cL277
This part here

shayne-fletcher-da: Oh! I see what you mean. I have no idea how that happened. Sorry, restored.

googleson78: Thanks! <3

ndmitchell:
Thanks!

PR #731, opened 2019-08-08
Convert 'Smell.hs' to ghc-lib

shayne-fletcher-da:
First pass at converting 'Smell.hs'.


ndmitchell:
The fact this was so easy means there can't be enough tests, and you might have broken something. But the lack of tests is my bug, and thus any bugs it introduces are my bugs, so I'll merge and accept any fallout!

shayne-fletcher-da:
I assure you, working this out was not easy 😉

PR #732, opened 2019-08-10
Convert 'Restrict.hs' to ghc-lib

shayne-fletcher-da:
This PR converts `Restrict.hs` to operate on `ghc-lib-parser` parse trees.

### DIFF
--- a/src/GHC/Util.hs
+++ b/src/GHC/Util.hs
@@ -14,7 +14,7 @@ module GHC.Util (
   , Located
   , readExtension
   , commentText, isCommentMultiline
-  , declName, modName
+  , declName, modName, rdrNameName
   , unsafePrettyPrint
   , eqMaybe
   , noloc, unloc, getloc, noext
@@ -22,6 +22,7 @@ module GHC.Util (
   , W(..)
   , SrcSpanD(..)
   , isDot, isDol
+  , flags, langExts
    -- Temporary : Export these so GHC doesn't consider them unused and
    -- tell weeder to ignore them.
   , isAtom, addParen, paren, isApp, isOpApp, isAnyApp, isSection, isDotApp
@@ -49,6 +50,7 @@ import "ghc-lib-parser" HeaderInfo
 import "ghc-lib-parser" ApiAnnotation
 import "ghc-lib-parser" Module
 
+import Data.Maybe
 import Data.List.Extra
 import Data.Function
 import System.FilePath
@@ -273,6 +275,9 @@ declName x = occNameString . occName <$> case x of
     ForD _ ForeignExport{fd_name} -> Just $ unLoc fd_name
     _ -> Nothing
 
+rdrNameName :: RdrName -> String
+rdrNameName = occNameString . rdrNameOcc
+
 modName :: HsModule GhcPs -> String
 modName HsModule {hsmodName=Nothing} = "Main"
 modName HsModule {hsmodName=Just (L _ n)} = moduleNameString n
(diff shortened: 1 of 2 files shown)


src/GHC/Util.hs
```diff
@@ -325,3 +330,33 @@ wToStr :: Outputable a => W a -> String
 wToStr (W e) = showPpr baseDynFlags e
 instance Outputable a => Eq (W a) where (==) a b = wToStr a == wToStr b
 instance Outputable a => Ord (W a) where compare = compare `on` wToStr
+
+-- GHC parse trees don't contain pragmas. We work around this with
```
googleson78: What is in charge of reading, e.g. `LANGUAGE` pragmas, and setting parser configuration (for things like `LambdaCase`)?

shayne-fletcher-da: It's done in a pre-pass by a function called [`parseDynamicFilePragma`](https://hackage.haskell.org/package/ghc-lib-parser-8.8.0.20190723/docs/DynFlags.html#v:parseDynamicFilePragma) (or rather `parseDynamicFlagsCmdLine` and its implementation details). See http://blog.shaynefletcher.org/2019/06/have-ghc-parsing-respect-dynamic-pragmas.html. In this case, the parse is already deemed successful so I think this simple code is rigorous enough for the job.

src/GHC/Util.hs
```diff
@@ -325,3 +330,33 @@ wToStr :: Outputable a => W a -> String
...
+    , "#" `isPrefixOf` body && "#" `isSuffixOf` body
+    , let s = trim $ take (length body - 2) (drop 1 body)
+  ]
+
+-- Flags. The first element of the pair is the (located) annotation
+-- comment that sets the flags enumerated in the second element of the
+-- pair.
+flags :: ApiAnns -> [(Located AnnotationComment, [String])]
+flags anns =
+  [(c, opts) | (c, s) <- pragmas anns, "OPTIONS_GHC" `isPrefixOf` s
```
ndmitchell: I think `OPTIONS` alone without `OPTIONS_GHC` probably also wants to be detected as a pragma. Pragmas are also case insensitive, so `options` and `optIONs` work just fine.

src/GHC/Util.hs
```diff
@@ -325,3 +330,33 @@ wToStr :: Outputable a => W a -> String
...
+    , let opts = words $ fromJust (stripPrefix "OPTIONS_GHC" s)]
+
+-- Language extensions. The first element of the pair is the (located)
+-- annotation comment that enables the extensions enumerated by he
+-- second element of the pair.
+langExts :: ApiAnns -> [(Located AnnotationComment, [String])]
+langExts anns =
+  [(c, exts) | (c, s) <- pragmas anns, "LANGUAGE" `isPrefixOf` s
```
ndmitchell: Pragmas are case insensitive.

ndmitchell: Much better to use `Just rest <- [stripPrefix "LANGUAGE" s]` rather than adding a `fromJust` in the path. Ditto for `flags`.

src/GHC/Util.hs
```diff
@@ -325,3 +330,33 @@ wToStr :: Outputable a => W a -> String
...
+-- (nasty) parsing of comments.
+
+-- Pragmas. Comments not associated with a span in the annotations
+-- that have the form @{-# ...#-}@.
+pragmas :: ApiAnns -> [(Located AnnotationComment, String)]
+pragmas anns =
+  [ (c, s) |
+      c@(L _ (AnnBlockComment comm)) <- fromMaybe [] $ Map.lookup noSrcSpan (snd anns)
+    , let body = trimCommentDelims comm
+    , "#" `isPrefixOf` body && "#" `isSuffixOf` body
```
ndmitchell: I'd rather a stripPrefix and stripSuffix,

src/GHC/Util.hs
```diff
@@ -325,3 +330,33 @@ wToStr :: Outputable a => W a -> String
...
+flags :: ApiAnns -> [(Located AnnotationComment, [String])]
+flags anns =
+  [(c, opts) | (c, s) <- pragmas anns, "OPTIONS_GHC" `isPrefixOf` s
+    , let opts = words $ fromJust (stripPrefix "OPTIONS_GHC" s)]
+
+-- Language extensions. The first element of the pair is the (located)
+-- annotation comment that enables the extensions enumerated by he
+-- second element of the pair.
+langExts :: ApiAnns -> [(Located AnnotationComment, [String])]
+langExts anns =
```
ndmitchell: Why not take in the output from `pragmas`, so we only have to do this once? Or do you think its not worth bothering to share the work? (Might very well be...)

src/GHC/Util.hs
shayne-fletcher-da: I see. The next PR in the queue does something about `OPTIONS`.

src/GHC/Util.hs
shayne-fletcher-da: Argh. My list comprehension fu is getting there but I'm no master yet 😉

src/GHC/Util.hs
shayne-fletcher-da: Right.

src/GHC/Util.hs
shayne-fletcher-da: No deep reason I believe. It's just the way the code emerged as I worked from the bottom to the end goal. I don't mind revisiting.

shayne-fletcher-da:
Thanks for the merge. Good to get the stack clear. I'll come back with another PR that addresses the suggestions raised here!

src/GHC/Util.hs
shayne-fletcher-da: I remember now. There was a twitter thread about `oPTionS` just recently!

src/GHC/Util.hs
shayne-fletcher-da: The `Just rest <- [stripPrefix "LANGUAGE" s]` is such a nifty pattern by the way. Deep art.

src/GHC/Util.hs
shayne-fletcher-da: ✔️

src/GHC/Util.hs
shayne-fletcher-da: ✔️

src/GHC/Util.hs
shayne-fletcher-da: ✔️

src/GHC/Util.hs
shayne-fletcher-da: ✔️

PR #734, opened 2019-08-12
Convert pragma to ghc-lib-parser

shayne-fletcher-da:
This PR converts `Pragma.hs` to GHC parse trees. It is based off the `convert-restrict` branch which is under review in https://github.com/ndmitchell/hlint/pull/732. It needs to be rebased on `master` if/once `convert-restrict` is merged.

PR #735, opened 2019-08-13
Act on review comments of the last couple of PRs

shayne-fletcher-da:
This PR follows through on the suggestions raised in https://github.com/ndmitchell/hlint/pull/732 and https://github.com/ndmitchell/hlint/pull/734.

### DIFF
--- a/src/GHC/Util.hs
+++ b/src/GHC/Util.hs
@@ -13,7 +13,7 @@ module GHC.Util (
   , SDoc
   , Located
   , readExtension
-  , commentText, isCommentMultiline
+  , comment, commentText, isCommentMultiline
   , declName, modName, rdrNameName
   , unsafePrettyPrint
   , eqMaybe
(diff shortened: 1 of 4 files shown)


src/GHC/Util.hs
```diff
@@ -345,30 +351,39 @@ pragmas anns =
...
+      , Just rest <- [stripSuffix "#" =<< stripPrefix "#" body]
+      , let s = trim rest
     ]
 
+-- Utility for a case insensitive prefix strip.
+stripPrefixCI :: String -> String -> Maybe String
+stripPrefixCI pref str =
+  let pref' = Text.toCaseFold (Text.pack pref)
+      (str_pref, rest) = Text.splitAt (Text.length pref') (Text.pack str)
+  in if Text.toCaseFold str_pref == pref' then Just (Text.unpack rest) else Nothing
```
ndmitchell: Why go via Text here? Why not just use `lower` to make things lower case? Or `upper` to make them upper case.

ndmitchell:
Thanks!

src/GHC/Util.hs
shayne-fletcher-da: https://github.com/ndmitchell/hlint/pull/736

Issue #733, opened 2019-08-11
Update the version used by Code Climate


neongreen:
The version they use currently is a year old.

https://hub.docker.com/r/codeclimate/codeclimate-hlint

ndmitchell:
I didn't write the code climate integration, I believe it was done by @pbrisbin.

pbrisbin:
:wave: hey there. I've lost touch with the Code Climate ecosystem since first submitting this engine... I've reached out to see what the current process is for getting it updated. Will report back soon, hopefully.

neongreen:
I DM'ed them on Twitter and they are planning to fork the engine to the codeclimate org and proceed from there. I haven't gotten any further replies from them since Monday though 🤔 

Pinging @davehenton.

maxjacobson:
Hi folks -

Just chatted with Dave. We had a misunderstanding and didn't realize that pbrisbin/codeclimate-hlint was [merged upstream](https://github.com/ndmitchell/hlint/pull/383) into this repo. So there's no need for us to fork that repo, or this one. We just need to update our internal system to point to a newer SHA in this repository. I've tried doing that, but I'm running into an error while building the image:

<details>
 <summary>Output</summary>

```text
$ git checkout v2.2.2
$ cc/build
Sending build context to Docker daemon  546.8kB
Step 1/33 : FROM fpco/stack-build:lts as builder
 ---> ed931fed86a2
Step 2/33 : MAINTAINER Pat Brisbin <pbrisbin@gmail.com>
 ---> Using cache
 ---> 87d8e219945e
Step 3/33 : ENV LANG en_US.UTF-8
 ---> Using cache
 ---> 6919f044cbd5
Step 4/33 : ENV PATH /root/.local/bin:$PATH
 ---> Using cache
 ---> ec6959d3e4c5
Step 5/33 : RUN mkdir -p /src
 ---> Using cache
 ---> 6bf87eea97c6
Step 6/33 : WORKDIR /src
 ---> Using cache
 ---> 0d03bbb440c3
Step 7/33 : COPY hlint.cabal /src/
 ---> Using cache
 ---> 5bc20125511a
Step 8/33 : RUN stack init
 ---> Running in d58137fea410
Looking for .cabal or package.yaml files to use to init the project.
Using cabal packages:
- ./

Selecting the best among 16 snapshots...

* Partially matches lts-14.1
    ghc-lib-parser version 8.8.0.20190424 found
        - hlint requires ==8.8.0.20190723
[...]

PR #737, opened 2019-08-14
Update the Code Climate image

neongreen:
Related to #733.

### DIFF
--- a/cc/Dockerfile
+++ b/cc/Dockerfile
@@ -1,46 +1,49 @@
 #
 # Build stage
 #
-FROM fpco/stack-build:lts as builder
+FROM haskell as builder
 MAINTAINER Pat Brisbin <pbrisbin@gmail.com>
 
-ENV LANG en_US.UTF-8
-ENV PATH /root/.local/bin:$PATH
+ENV LANG C.UTF-8
 
 RUN mkdir -p /src
 WORKDIR /src
 
-# Initialize package/stack
+# Initialize package
 COPY hlint.cabal /src/
-RUN stack init
-RUN stack setup
+RUN cabal v2-update
 
 # Install HLint
-RUN stack install --dependencies-only
+RUN cabal v2-build --dependencies-only
 COPY ./src /src/src
 COPY ./data /src/data
 COPY ./LICENSE /src/LICENSE
-RUN stack install
+COPY ./README.md /src/README.md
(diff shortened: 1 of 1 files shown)


cc/Dockerfile
```diff
@@ -1,46 +1,49 @@
...
 #
-FROM fpco/stack-run:lts
+FROM debian:stretch
 MAINTAINER Pat Brisbin <pbrisbin@gmail.com>
 
-ENV LANG en_US.UTF-8
+ENV LANG C.UTF-8
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends libgmp-dev
```
neongreen: I don't really like this part but I haven't found a haskell-env image.

cc/Dockerfile
```diff
@@ -1,46 +1,49 @@
 #
 # Build stage
 #
-FROM fpco/stack-build:lts as builder
+FROM haskell as builder
 MAINTAINER Pat Brisbin <pbrisbin@gmail.com>
```
neongreen: I can put my email here or leave Pat's, it looks like we both are pretty responsive so it likely doesn't matter much.

cc/Dockerfile
```diff
@@ -1,46 +1,49 @@
...
+RUN cabal v2-update
 
 # Install HLint
-RUN stack install --dependencies-only
+RUN cabal v2-build --dependencies-only
 COPY ./src /src/src
 COPY ./data /src/data
 COPY ./LICENSE /src/LICENSE
-RUN stack install
+COPY ./README.md /src/README.md
```
neongreen: This part is fragile but I wanted to avoid rewriting the Dockerfile completely. Edits welcome.

maxjacobson:
Thanks for opening this. I'll defer to the haskell developers on the particulars. Feel free to ping me if you have any questions about how to QA this change.

neongreen:
Not sure why CI fails – it doesn't seem to be related to this PR. Any advice?

ndmitchell:
I have no real opinion here. If no one objects, I'll merge.

pbrisbin:
:wave: sorry I was on vacation last week so didn't get to addressing this. I just wanted to say thanks and mention for posterity that I had no affinity for Stack here, it was just path of least resistance at the time. So if moving to cabal indeed fixings things (and avoids the bitrot/non-reproduciblility of a `stack init` in a layer), than big :+1: from me.

PR #740, opened 2019-08-14
Fix a failing test

josephcsible:
Replace the first argument to fromMaybe with an atom, to reflect the change made to the hint in 54aa9a5.


ndmitchell:
Thanks!

PR #742, opened 2019-08-18
Update source location handling

shayne-fletcher-da:
For your consideration. This PR adds nothing new instead, it updates source location handling to use the new `dL` view-pattern and `cL` function. Thus, this set of modifications I view as being about "future proofing" and "observing best practices". The background is that the `HsSyn` types are gradually being transformed in accordance with "trees that grow" principles. In particular for us here, the way that source code locations are handled. See [this Wiki page](https://gitlab.haskell.org/ghc/ghc/wikis/implementing-trees-that-grow/handling-source-locations) and [this issue](https://gitlab.haskell.org/ghc/ghc/issues/15495) for more details.


ndmitchell:
Thanks!

PR #743, opened 2019-08-20
Fix Codeclimate ("hlint lint" is broken)

neongreen:
I don't know how this slipped, but apparently `hlint lint` doesn't work anymore and thus Codeclimate HLint integration is broken.

This PR removes "hlint lint" from the Codeclimate Dockerfile, but we might also want to remove mentions of "`[lint]`" from the CLI docs:

```
hlint lint --help
HLint v2.2.2, (C) Neil Mitchell 2006-2019

hlint [lint] [OPTIONS] [FILE/DIR]
...
```

```
hlint lint .
Couldn't find file: lint
```

### DIFF
--- a/cc/Engine.hs
+++ b/cc/Engine.hs
@@ -63,8 +63,7 @@ main = do
     c <- readConfig "/config.json"
 
     callProcess "hlint" $
-        [ "lint"
-        , "--cc"
+        [ "--cc"
         , "--datadir", "/opt/hlint"
         , "--no-exit-code"
         ]


ndmitchell:
`hlint lint .` works absolutely fine for me... I wonder if it's your build, some shell script wrapper, an old version, mismatched compiler/libraries or what... Can anyone else reproduce it not working?

neongreen:
Oh, I found how to reproduce it.

```yaml
# .hlint.yaml
- arguments:
    - -XTypeApplications
    - -XQuasiQuotes
```

If you have `- arguments:` set in your .hlint.yaml, it breaks.

ndmitchell:
The arguments are prepended to the command line, which works really badly if you have a mode at the start. To be honest, I'm still not sure how `arguments` in the YAML file should be handled. Maybe at the end is better?

neongreen:
I think this should be discussed in a separate issue, and for now we should decide what to do with this PR. Is it safe to remove the mode from the HLint invocation? If yes, I'd rather unbreak Codeclimate first. If it's not {safe/future-proof/etc} I'll close this PR and we can get to the drawing board.

ndmitchell:
The only reason to type lint explicitly is that there's a risk the first argument might itself be lint (or one of the other HLint modes). Since you explicitly pass --cc first that can't be the case, so this change seems nice and safe.

I agree there's a bigger issue around arguments to be discussed elsewhere.

PR #749, opened 2019-08-24
List recursion rules, conversion to ghc-lib

shayne-fletcher-da:
Convert `ListRec.hs` to `ghc-lib-parser`.


src/GHC/Util/FreeVars.hs
```diff
@@ -0,0 +1,259 @@
+{-# LANGUAGE CPP #-}
```
ndmitchell: Why is CPP on here?

ndmitchell:
Nice work. The random CPP should definitely go, but otherwise its close enough that anything else is just quibbling.

src/GHC/Util/FreeVars.hs
ndmitchell: Get rid of that - it supports GHC's older than HLint does. Just delete the lines.

PR #751, opened 2019-08-26
Update ghc-lib to version 8.8.1

ghost:
Update to `ghc-lib-parser-8.8.1`.


ndmitchell:
I see the stackage date is tomorrow. They don't always cut one, and I suspect 8.8.1 won't be in it anyway, so why pick that date?

ndmitchell:
I don't think there's any reason to change the stackage date at all. I just pushed 13009029d92cf62249c2df2154a41fe9b05b7b1d with your changes minus that one.

PR #752, opened 2019-08-26
Remove srcloc qualifications

shayne-fletcher-da:
This PR is conditional on https://github.com/ndmitchell/hlint/pull/749 and will need rebasing on master after that one has landed before merge.

This gist of this PR is to (1) remove `qualified` from import declarations like `import qualified "ghc-lib-parser" X`  (2) replace `GHC.x` with simply `x` and (3) usages of view pattern `dL ->` with the synonym `LL`. Overall, the resulting code should read much more naturally.


ndmitchell:
Looks reasonable - thanks.

PR #753, opened 2019-08-26
Remove usage of CPP

shayne-fletcher-da:
Following up on the review of https://github.com/ndmitchell/hlint/pull/749, removes the use of `CPP`.


ndmitchell:
Thanks

PR #754, opened 2019-09-02
Use Numerical Value For Nonroot User To Avoid Additional Layer

brandshaide:
Avoid user  (`app`) creation as it's creating a unnecessary additional layer and the user itself is not being used anyway. Instead prefer to use numerical values, i. e. as this PR is introducing with a `uid` of `1000` for a `nonroot` user.

### DIFF
--- a/cc/Dockerfile
+++ b/cc/Dockerfile
@@ -1,4 +1,12 @@
 #
+#
+#############
+#
+# Please be aware of the specification as of: 
+# https://github.com/codeclimate/spec/blob/master/SPEC.md
+#
+############
+#
 # Build stage
 #
 FROM haskell as builder
--- a/cc/Dockerfile
+++ b/cc/Dockerfile
@@ -50,7 +50,6 @@ RUN mkdir -p /code
 VOLUME /code
 WORKDIR /code
 
-RUN useradd app --uid 9000
-USER app
+USER 1000
 
 CMD ["/usr/bin/engine"]
--- a/cc/Dockerfile
+++ b/cc/Dockerfile
@@ -50,6 +50,7 @@ RUN mkdir -p /code
 VOLUME /code
 WORKDIR /code
 
-USER 1000
+RUN useradd app --uid 9000
+USER app
 
 CMD ["/usr/bin/engine"]


ndmitchell:
CC @pbrisbin for a view on this. I have no real idea, but happy to push the merge button if those who contributed think its some good idea.

pbrisbin:
I get where you're coming from, and it's a good idea layer-wise, but the SPEC actually says specifcally an `app` user with id `9000`:

> Engines are always run as a user `app` with UID and GID of 9000, and never `root`.

https://github.com/codeclimate/spec/blob/master/SPEC.md#security-restrictions

I don't know how much they enforce this, and having any non-root user should satisfy the actual security concerns (I think?), but technically this would make the image violate the spec as it's currently written. So I'm a soft-no, myself.

brandshaide:
thank, @pbrisbin for diggin into this. Of course we want to comply to the specs and thus leave it like it is. 

>  [...]don't know how much they enforce this, and having any non-root user should satisfy the actual security concerns (I think?) [...]
me neither - and AFAIK it doesn't make as user "more secure" by giving him a `UID` way larger than 1000. 

> Engines are always run as a user app with UID and GID of 9000, and never root.
Hm, but the `Dockerfile` does not contain creation of a correspondig `GID`...

ndmitchell:
If the decision is to leave it as it is, but for slightly subtle reasons, maybe a comment in the Docker file would be useful?

pbrisbin:
> Hm, but the Dockerfile does not contain creation of a correspondig GID...

:man_shrugging: 

> maybe a comment in the Docker file would be useful?

A comment at the top linking to the SPEC sounds pretty reasonable. There are a few things a `Dockerfile` is supposed to do, so that might be better then a comment specifically near or about the `app` user?

ndmitchell:
Happy with whatever comments you think are best.

brandshaide:
I'll leave a comment on top of the  `Dockerfile` as suggested by @pbrisbin .

brandshaide:
> maybe a comment in the Docker file would be useful? 

done :-)

pbrisbin:
Works for me!

ndmitchell:
Thanks!

Issue #756, opened 2019-09-04
Warnings with -fwarn-incomplete-patterns


ndmitchell:
It currently gives warnings on things such as:

```
src\Hint\Naming.hs:106:1: warning: [-Wincomplete-patterns]
    Pattern match(es) are non-exhaustive
    In an equation for `getNames': Patterns not matched: _
    |
106 | getNames (LL _ decl) = maybeToList (declName decl) ++ getConstructorNames decl
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I'm thinking LL pattern synonym needs a COMPLETE pragma? https://gitlab.haskell.org/ghc/ghc/wikis/pattern-synonyms/complete-sigs

CC @shayne-fletcher-da

shayne-fletcher-da:
Thank-you. I'll pick it up. Leave it with me.

shayne-fletcher-da:
Despite contradictory documentation in the Wiki page referenced above, it appears that the `COMPLETE` pragma must be associated with the defining module of the pattern synonym.
```
    • Orphan COMPLETE pragmas not supported
      A COMPLETE pragma must mention at least one data constructor
      or pattern synonym defined in the same module.
 ```

shayne-fletcher-da:
Worked around for the existing code base in https://github.com/ndmitchell/hlint/pull/757. I'll look into an upstream fix.

ndmitchell:
Shame - never mind though, let's close this issue then.

PR #757, opened 2019-09-04
Convert 'Hint.Brackets'

shayne-fletcher-da:
Converts `src/Hint/Brackets.hs` to `ghc-lib-parser`.


shayne-fletcher-da:
Looks like the inclusion of `ExactPrint/Types` does not play well with 8.4.

shayne-fletcher-da:
The last commit deals with the incomplete pattern warnings described by issue https://github.com/ndmitchell/hlint/issues/756 and enables `-Wincomplete-patterns` in `stack.yaml`.

#.hlint.yaml#
```diff
@@ -0,0 +1,78 @@
+# HLint configuration file
```
ndmitchell: Mistake? Feel free to add a gitignore pattern that catches this file.

.#.hlint.yaml
```diff
@@ -0,0 +1 @@
+shaynefletcher@shaynefletcherd2jgh7.97454
```
ndmitchell: Mistake?

.hlint.yaml
```diff
@@ -14,8 +14,10 @@
   - name: [ExistentialQuantification, MultiParamTypeClasses, NamedFieldPuns]
   - name: [FlexibleContexts, FlexibleInstances]
   - name: [PackageImports]
+  - {name: ConstraintKinds, within: [GHC.Util.Language.Haskell.GHC.ExactPrint.Types]}
```
ndmitchell: This is inoffensive, so just allow ConstraintKinds everywhere.

.hlint.yaml
```diff
@@ -14,8 +14,10 @@
...
   - {name: CPP, within: [HsColour]} # so it can be disabled to avoid GPL code
-  - {name: TypeFamilies, within: [GHC.Util.HsExpr, GHC.Util.FreeVars, Hint.Export, Hint.Restrict]} # type family constraints here
+  - {name: TypeFamilies, within: [GHC.Util.HsExpr, GHC.Util.FreeVars, Hint.Export, Hint.Restrict, Hint.Bracket]} # type family constraints here
+  - {name: RankNTypes, within: [GHC.Util.Language.Haskell.GHC.ExactPrint.Types]}
```
ndmitchell: Again, RankNTypes and TypeFamilies are inoffensive.

hlint.cabal
```diff
@@ -60,7 +60,9 @@ library
         extra >= 1.6.6,
         refact >= 0.3,
         aeson >= 1.1.2.0,
-        ghc-lib-parser == 8.8.1
+        ghc-lib-parser == 8.8.1,
+        syb >= 0.7.1,
+        mtl >= 2.2.2
```
ndmitchell: Usually transformers is a better choice.

src/CC.hs
```diff
@@ -12,7 +12,6 @@ module CC
 
 import Data.Aeson (ToJSON(..), (.=), encode, object)
 import Data.Char (toUpper)
-import Data.Monoid ((<>))
```
ndmitchell: Why touch this module at all?

src/GHC/Util.hs
```diff
@@ -17,10 +17,12 @@ module GHC.Util (
   , parsePragmasIntoDynFlags, parseFileGhcLib
   ) where
 
+import GHC.Util.Brackets
```
ndmitchell: Why touch this module but not change its API?

src/GHC/Util/Brackets.hs
```diff
@@ -0,0 +1,155 @@
...
+
+  addParen' e = noLoc $ HsPar noExt e
+
+  isAtom' (LL _ x) = case x of
+      HsVar{} -> True
+      HsUnboundVar{} -> True
+      HsRecFld{} -> True
+      HsOverLabel{} -> True
+      HsIPVar{} -> True
+      -- Note that sections aren't atoms (but parenthesized sections are).
```
ndmitchell: Are there such things as non-paren sections? Are those not user writable?

src/GHC/Util/HsDecl.hs
```diff
@@ -1,12 +1,15 @@
...
 
-module GHC.Util.HsDecl (declName) where
+module GHC.Util.HsDecl (declName,isForD') where
 
 import "ghc-lib-parser" HsSyn
 import "ghc-lib-parser" OccName
 import "ghc-lib-parser" SrcLoc
 
+isForD' :: LHsDecl GhcPs -> Bool
+isForD' (LL _ ForD{}) = True; isForD' _ = False
```
ndmitchell: Why the ' after isForD?

src/GHC/Util/Language/Haskell/GHC/ExactPrint/Types.hs
```diff
@@ -0,0 +1,378 @@
+-- Adapted from https://github.com/alanz/ghc-exactprint.git.
```
ndmitchell: Could we depend on ghc-exactprint? Copying this code wholescale is scary.

hlint.cabal
ndmitchell: But not at the cost of rewriting copied over code!

ndmitchell:
With the two garbage files removed, I'm happy to merge this in.

#.hlint.yaml#
shayne-fletcher-da: Oh dear. Will fix.

.#.hlint.yaml
shayne-fletcher-da: Oh most definitely.

.hlint.yaml
shayne-fletcher-da: ✔️

.hlint.yaml
shayne-fletcher-da: ✔️

hlint.cabal
shayne-fletcher-da: ✔️ (won't fix).

src/CC.hs
shayne-fletcher-da: Ah! Suppress a warning that exhibits with 8.8.1!

src/GHC/Util.hs
shayne-fletcher-da: An error. Vestige of an intermediate approach. Will remove.

src/GHC/Util/Brackets.hs
shayne-fletcher-da: As far as I can tell, it doesn't seem lexically possible to write a naked section! Ready to be corrected. What I believe to be true is : (1) HSE sees `(+ 1)` as a section, the parens are implicit and the pretty printer is made to write them (2) GHC sees `(+ 1)` as a `HsParen (HsSection ...))`, the parens are explicit and as mentioned above, I don't think it's lexically possible for a user to write a section that isn't parenthesized in this way.

src/GHC/Util/HsDecl.hs
shayne-fletcher-da: For the time being, I'm writing these ticks into all utilities that can end up cohabiting with HSE code in intermediate states as I convert modules containing before and after versions of a program in parallel - it's a device.  It's on the `isForD` for consistencies' sake at this time. When the conversion is done, the ticks will go away completely (so I hope you're not too fussed at this state of affairs today)!

src/GHC/Util/Language/Haskell/GHC/ExactPrint/Types.hs
shayne-fletcher-da: If we put our mind to it I suppose we could. My thoughts though are running in the other direction. This module is required by the apply-refact algorithm but without having looked at it deeply, really only for the type declarations - nothing substantial. I have been thinking that until proven otherwise, maybe a better thing to do is rewrite the algorithm to use our own types and drop this module. Something to come back to? Maybe an issue?

shayne-fletcher-da: > Could we depend on ghc-exactprint?
Almost certainly at the cost of introducing longer build times - we'd need `ghc-lib` I think. You'll note comparing the code with the original, I carefully avoided that bullet thus far :) To clarify : it might be a good idea to depend on exact-print and apply-refact and get this for free. Don't know right now.

src/GHC/Util/HsDecl.hs
ndmitchell: 👍

src/GHC/Util/Brackets.hs
ndmitchell: TIL - thanks

src/GHC/Util/Language/Haskell/GHC/ExactPrint/Types.hs
ndmitchell: Fine, let's go with what we have now, as its not a big problem right now, and revisit it later.

shayne-fletcher-da:
As always, quality review. Thanks! Looking forward to continuing the conversation around the vendored-in code. Perhaps you'd like to kick off an issue and we can work this out?

ndmitchell:
For the vendored stuff, I'd just forget it for now, and revisit it once we are totally on GHC.

src/GHC/Util/Brackets.hs
shayne-fletcher-da: Me too. Brought to you by lots of head scratching here 😉

src/GHC/Util/Language/Haskell/GHC/ExactPrint/Types.hs
shayne-fletcher-da: ✔️

PR #758, opened 2019-09-07
Relax to syb >= 0.7

shayne-fletcher-da:
Relax the lower bound on syb (for whatever reason, DA CI highlights the need for this).


ndmitchell:
Sure. Happy to relax the bounds further if anyone cares - I suspect we don't have any hard dependencies on that particular version.

PR #760, opened 2019-09-09
Apply fixity adjustments in 'createModuleEx'

shayne-fletcher-da:
Run the `applyFixities` algorithm over the parse tree given to `createModuleEx`.


shayne-fletcher-da:
I too am dissatisfied and believe we'll revisit when the dust has settled. It does appear to be the most direct approach for the moment. Thank-you!

PR #765, opened 2019-09-16
Somehow '*' ended up in here. Sorry!

shayne-fletcher-da:
I cleverly managed write an exclusion rule for all files in the `.gitignore`. Obviously I was stress testing the code-review process. This fixes it.


ndmitchell:
Oops! But it does happen to everyone.

PR #782, opened 2019-10-15
Convert patterns to ghc-lib

shayne-fletcher-da:
Reimplement `Pattern.hs` to use GHC parse trees.

### DIFF
--- a/src/GHC/Util/HsPat.hs
+++ b/src/GHC/Util/HsPat.hs
@@ -1,9 +1,9 @@
 {-# LANGUAGE MultiParamTypeClasses , FlexibleInstances, FlexibleContexts #-}
 
-module GHC.Util.Pat (
+module GHC.Util.HsPat (
     strToPat', patToStr'
   , Brackets'(..)
-  , fromPChar', isPFieldWildcard', hasPFieldsDotDot'
+  , fromPChar', isPFieldWildcard', hasPFieldsDotDot', isPWildCard'
   ) where
 
 import HsSyn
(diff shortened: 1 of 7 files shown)


src/Hint/Pattern.hs
```diff
@@ -118,81 +102,133 @@ hints gen (Pattern pats (GuardedRhss _ [GuardedRhs _ [Generator _ pat (App _ op
...
-patHint False strict o@(PIrrPat _ x) | f x = [warn "Redundant irrefutable pattern" o x [r]]
-    where f (PParen _ x) = f x
-          f (PAsPat _ _ x) = f x
-          f PWildCard{} = True
-          f PVar{} = True
-          f _ = False
-          r = Replace R.Pattern (toSS o) [("x", toSS x)] "x"
-patHint lang strict o@(PAsPat u v PWildCard{}) = [warn "Redundant as-pattern" o v []]
+asPattern :: LHsDecl GhcPs  -> [(Pattern, String -> Pattern -> [Refactoring R.SrcSpan] -> Idea)]
+asPattern (LL loc x) = concatMap decl (universeBi x) -- ++ concatMap alt (universeBi x)
```
googleson78: Commented out code left here?

shayne-fletcher-da: Nice catch.

src/GHC/Util/HsPat.hs
```diff
@@ -1,9 +1,9 @@
 {-# LANGUAGE MultiParamTypeClasses , FlexibleInstances, FlexibleContexts #-}
```
ndmitchell: Why the rename?

ndmitchell:
Thanks!

src/GHC/Util/HsPat.hs
shayne-fletcher-da: Not a technical reason. The naming scheme for these modules is that they are named after the (module/)datatype for which they relate e.g. `GHC.Util.HsExpr` for `HsExpr`, `GHC.Util.HsDecl` for `HsDecl` and so on.  Actually, reviewing this I see I that in fact, though the module name is `HsPat`, the type is now `Pat` (doh!) so it was right the way it was. I'll revert the rename.

All the above said, don't know if you are aware - on HEAD there is lots of churn here to introduce a new "GHC module hierarchy" and everything is getting moved.

PR #784, opened 2019-10-16
Restore HsPat.hs -> Pat.hs

shayne-fletcher-da:
See https://github.com/ndmitchell/hlint/pull/782 for an explanation of this rename.


ndmitchell:
I'm not overly fussed either way, but if this is the end state, let's jump to it.

PR #785, opened 2019-10-17
Don't warn on redundant cases for generated code

shayne-fletcher-da:
Slight refinement to `expHint` for patterns : don't warn about redundant cases on code that is generated.


ndmitchell:
Where do case expressions arise not marked FromSource?

shayne-fletcher-da:
> Where do case expressions arise not marked FromSource?

DAML's generics preprocessor (for example).

PR #789, opened 2019-10-28
Fix a typo ("usally" should be "usually")

ocharles:


ndmitchell:
Thanks!

shayne-fletcher-da:
Bet it was my fault. I'm always getting my mucking furds wuddled 😄 Thank-you!

PR #790, opened 2019-10-30
Convert Monad.hs to ghc-lib

shayne-fletcher-da:
Convert `Monad.hs` to ghc-lib.

### DIFF
--- a/src/GHC/Util/HsExpr.hs
+++ b/src/GHC/Util/HsExpr.hs
@@ -8,13 +8,14 @@
 
 module GHC.Util.HsExpr (
     noSyntaxExpr'
-  , isDol', isDot', isSection', isRecConstr', isRecUpdate', isVar', isPar', isApp', isAnyApp', isLexeme', isReturn'
+  , isTag', isDol', isDot', isSection', isRecConstr', isRecUpdate', isVar', isPar', isApp', isAnyApp', isLexeme', isReturn'
   , dotApp'
   , simplifyExp', niceLambda', niceDotApp'
   , Brackets'(..)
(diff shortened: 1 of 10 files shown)


src/GHC/Util/HsExpr.hs
```diff
@@ -63,17 +64,27 @@ paren' x
...
+isDol' x = isTag' x "$"
+isReturn' x = isTag' x "return" || isTag' x "pure" -- Allow both 'pure' and 'return' as they have the same semantics.
 isRecConstr' (LL _ RecordCon{}) = True; isRecConstr' _ = False
 isRecUpdate' (LL _ RecordUpd{}) = True; isRecUpdate' _ = False
 isDotApp' (LL _ (OpApp _ _ op _)) = isDot' op; isDotApp' _ = False
 isLexeme' (LL _ HsVar{}) = True;isLexeme' (LL _ HsOverLit{}) = True;isLexeme' (LL _ HsLit{}) = True;isLexeme' _ = False
--- Allow both 'pure' and 'return' as they have the same semantics.
-isReturn' (LL _ (HsVar _ (LL _ (Unqual x)))) = occNameString x == "return" || occNameString x == "pure"; isReturn' _ = False
+
+--
```
googleson78: Comment with nothing in it here?

googleson78:
Very minor thing but if we're following Haddock markup shouldn't all the code samples be in `@@`s, not in single quotes?

src/GHC/Util/HsExpr.hs
ndmitchell: I tend to use double newlines rather than `--` breaks

ndmitchell:
Nice!

PR #795, opened 2019-11-25
Convert match

shayne-fletcher-da:
WIP : Unification implemented and all of `Match.hs` up to but not including `matchIdea` which is where the porting effort needs to be resumed from.

### DIFF
--- a/src/GHC/Util/Unify.hs
+++ b/src/GHC/Util/Unify.hs
@@ -1,0 +1,178 @@
+{-# LANGUAGE PatternGuards, ViewPatterns, FlexibleContexts, ScopedTypeVariables #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module GHC.Util.Unify(
+    Subst', fromSubst',
+    validSubst', substitute',
+    unifyExp'
+    ) where
+
+import Control.Monad
+import Data.Generics.Uniplate.Operations
+import Data.Char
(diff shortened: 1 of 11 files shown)


ndmitchell:
Nice. Will merge once CI passes.

src/GHC/Util/Unify.hs
```diff
@@ -0,0 +1,178 @@
...
+type NameMatch' = Located RdrName -> Located RdrName -> Bool
+
+-- | Unification, obeys the property that if @unify a b = s@, then
+-- @substitute s a = b@.
+unify' :: Data a => NameMatch' -> Bool -> a -> a -> Maybe (Subst' (LHsExpr GhcPs))
+unify' nm root x y
+    | Just (x, y) <- cast (x, y) = unifyExp' nm root x y
+    | Just (x, y) <- cast (x, y) = unifyPat' nm x y
+    | Just (x, y) <- cast (x, y) = unifyType' nm x y
+    -- | Just (x :: S) <- cast x = Just mempty -- Confused by this :-/
```
shayne-fletcher-da: @ndmitchell Can you comment on this line please?

ndmitchell: This is to avoid descending into the source span bit, so the equivalent would be `SrcSpan` in the new world. It ensures that any two src spans match regardless.

ndmitchell:
Thanks!

PR #797, opened 2019-11-27
Convert match

shayne-fletcher-da:
The conversion of `Match.hs` is done. The only thing not done yet is to try executing it.


ndmitchell:
Looks plausible...

ndmitchell:
CI says yes! Thanks - I await an actual test of it.

shayne-fletcher-da:
Yes, there is at least one known bug and more PRs to follow but for now, here is the first bit of proof positive.

Screen Shot 2019-11-27 at 7 21 03 PM

PR #798, opened 2019-11-29
Debug match

shayne-fletcher-da:
This PR fixes up yaml parsing so that GHC scopes are populated. `resolveHints` continues to use the HSE `readMatch` function, all tests pass. To switch to the ghc implementation, `readMatch'` toggle the manifest constant `HLINT_GHC_MATCH_ENABLE` in `Hint/All.hs`. Many tests pass but there remain some failures that need debugging before we can make the switch.

### DIFF
--- a/src/Hint/All.hs
+++ b/src/Hint/All.hs
@@ -77,8 +77,14 @@ builtinHints = [(drop 4 $ show h, builtin h) | h <- [minBound .. maxBound]]
 
 -- | Transform a list of 'HintBuiltin' or 'HintRule' into a 'Hint'.
 resolveHints :: [Either HintBuiltin HintRule] -> Hint
-resolveHints xs = mconcat $ mempty{hintDecl=const $ readMatch rights} : map builtin (nubOrd lefts)
-    where (lefts,rights) = partitionEithers xs
+resolveHints xs =
+  if False then
+      -- GHC
+    mconcat $ mempty{hintDecl'=const $ readMatch' rights} : map builtin (nubOrd lefts)
+  else
+      -- HSE
+    mconcat $ mempty{ hintDecl=const $  readMatch rights} : map builtin (nubOrd lefts)
+  where (lefts,rights) = partitionEithers xs
 
 -- | Transform a list of 'HintRule' into a 'Hint'.
 hintRules :: [HintRule] -> Hint
--- a/src/Hint/Match.hs
+++ b/src/Hint/Match.hs
@@ -40,7 +40,7 @@ not . not ==> id
 not . not . x ==> x
 -}
 
-module Hint.Match(readMatch) where
+module Hint.Match(readMatch,readMatch') where
 
 import Control.Applicative
 import Data.List.Extra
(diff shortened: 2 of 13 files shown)


shayne-fletcher-da:
We also now apply fixities to yaml expressions. This brings the test count failure down a bit.

src/Hint/All.hs
```diff
@@ -75,10 +76,20 @@ builtin x = case x of
 builtinHints :: [(String, Hint)]
 builtinHints = [(drop 4 $ show h, builtin h) | h <- [minBound .. maxBound]]
 
+-- Toggle implementation.
+#define HLINT_GHC_MATCH_ENABLE 0
 -- | Transform a list of 'HintBuiltin' or 'HintRule' into a 'Hint'.
```
ndmitchell: Do we need CPP here? Why not just if True then ... else ...

src/Hint/Match.hs
```diff
@@ -327,13 +329,13 @@ checkSide' x bind = maybe True bool x
       isType "InfixAp" (GHC.LL _ GHC.OpApp{}) = True
       isType "Paren" (GHC.LL _ GHC.HsPar{}) = True
       isType "Tuple" (GHC.LL _ GHC.ExplicitTuple{}) = True
-      isType typ _ = error $ "Hint.Match.checkSide', unknown side condition: '" ++ "is" ++ typ ++ "'"
+      isType _ _ = False
```
ndmitchell: This seems dodgy

ndmitchell:
Awesome. Let me know on the two points I flagged and I'll merge.

src/Hint/Match.hs
```diff
@@ -327,13 +329,13 @@ checkSide' x bind = maybe True bool x
...
 
       asInt :: GHC.LHsExpr GHC.GhcPs -> Maybe Integer
       asInt (GHC.LL _ (GHC.HsPar _ x)) = asInt x
       asInt (GHC.LL _ (GHC.NegApp _ x _)) = negate <$> asInt x
       asInt (GHC.LL _ (GHC.HsLit _ (GHC.HsInt _ (GHC.IL _ neg x)) )) = Just $ if neg then -x else x
-      asInt _ = Nothing
+      asInt x = Nothing
```
googleson78: Why the unused name?

shayne-fletcher-da: Just a left-over from debugging.

src/Hint/All.hs
shayne-fletcher-da: Because stupid.

src/Hint/Match.hs
shayne-fletcher-da: Improved this now in two ways : (1) Take into account overloaded literals and (2) the last case now respects the spirit of the original (but likely not the end of the story here - we'll see). Failure count has come down.

shayne-fletcher-da:
@ndmitchell Can you revisit this now please?

ndmitchell:
Merged- thanks!

PR #799, opened 2019-11-29
Debug match

shayne-fletcher-da:
Found and patched an issue with free variable calculations for match groups. Now have 13/129 match test failures (down from 22) or, 90% success.


ndmitchell:
Thanks!

PR #801, opened 2019-11-30
Debug match

shayne-fletcher-da:
Fix match test failures.

(1) Rephrase free vars patch.
    - Equivalent, more elegant code : no change expected.
(2) Parenthesize sections in `dotApps'`.
(3) Fix spelling mistake for string literal (`"isInfixAp"` / `"isInfixApp`") in `checkSides`.

TL;DR 3 failures all due to slight output format differences.

- There are 687 tests
- 170 are match tests
- 41  are from `Test.hs`; these are excluded
- there are 129 other match tests
- there are 4 test failures:
  - one is a grep test
  - the remaining 3 failures are due to output format

(0)
```
TEST FAILURE IN tests/_grep_2
DIFFER ON LINE: 1
GOT : <EOF>
WANT: tests/grep2.hs:1:8:
FULL OUTPUT FOR GOT:
```
(1)
```
GOT :   if isNothing x then (- 1.0) else fromJust x
WANT:   if isNothing x then (-1.0) else fromJust x
```
(2)
```Found:
  concat . map f . baz . bar
Perhaps:
  (concatMap f . (baz . bar))

WANTED: concatMap f . baz . bar

(3)
```Found:
  typeOf (undefined :: Foo Int)
Perhaps:
  typeRep (Proxy :: Proxy Foo Int)

WANTED: typeRep (Proxy :: Proxy (Foo Int))
```

### DIFF
--- a/src/GHC/Util/HsExpr.hs
+++ b/src/GHC/Util/HsExpr.hs
@@ -289,7 +289,7 @@ replaceBranches' x = ([], \[] -> x)
 -- removed from haskell-src-exts-util-0.2.2.
 needBracketOld' :: Int -> LHsExpr GhcPs -> LHsExpr GhcPs -> Bool
 needBracketOld' i parent child
-  | isDotApp' parent, isDotApp' child, i == 1 = False
+  | isDotApp' parent, isDotApp' child, i == 2 = False
   | otherwise = needBracket' i parent child
 
 transformBracketOld' :: (LHsExpr GhcPs -> Maybe (LHsExpr GhcPs)) -> LHsExpr GhcPs -> LHsExpr GhcPs
--- a/src/GHC/Util/FreeVars.hs
+++ b/src/GHC/Util/FreeVars.hs
@@ -25,7 +25,6 @@ import Data.Set (Set)
 import qualified Data.Set as Set
 import Prelude
 
-
 ( ^+ ) :: Set OccName -> Set OccName -> Set OccName
 ( ^+ ) = Set.union
 ( ^- ) :: Set OccName -> Set OccName -> Set OccName
@@ -33,6 +32,12 @@ import Prelude
 
 -- See [Note : Spack leaks lurking here?] below.
 data Vars' = Vars'{bound' :: Set OccName, free' :: Set OccName}
+
+-- Useful for debugging.
+instance Show Vars' where
+  show (Vars' bs fs) = "bound : " ++
+    show (map occNameString (Set.toList bs)) ++
+    ", free : " ++ show (map occNameString (Set.toList fs))
 
 instance Semigroup Vars' where
     Vars' x1 x2 <> Vars' y1 y2 = Vars' (x1 ^+ y1) (x2 ^+ y2)
@@ -217,7 +222,7 @@ instance AllVars' (LHsBind GhcPs) where
 instance AllVars' (LMatch GhcPs (LHsExpr GhcPs)) where
   allVars' (dL -> L _ (Match _ FunRhs {mc_fun=name} pats grhss)) = allVars' (VarPat noExt name :: Pat GhcPs) <> allVars' pats <> allVars' grhss -- A pattern matching on an argument of a function binding.
   allVars' (dL -> L _ (Match _ (StmtCtxt ctxt) pats grhss)) = allVars' ctxt <> allVars' pats <> allVars' grhss -- Pattern of a do-stmt, list comprehension, pattern guard etc.
-  allVars' (dL -> L _ (Match _ _ pats grhss)) = allVars' pats <> allVars' grhss -- Everything else.
+  allVars' (dL -> L _ (Match _ _ pats grhss)) = inVars' (allVars' pats) (allVars' grhss) -- Everything else.
 
   allVars' _ = mempty -- New ctor.
 
@@ -234,7 +239,7 @@ instance AllVars' (GRHSs GhcPs (LHsExpr GhcPs)) where
   allVars' _ = mempty -- New ctor.
 
 instance AllVars' (LGRHS GhcPs (LHsExpr GhcPs)) where
-  allVars' (dL -> L _ (GRHS _ guards expr)) =  let gs = allVars' guards in Vars' (bound' gs) (free' gs ^+ (freeVars' expr ^- bound' gs))
+  allVars' (dL -> L _ (GRHS _ guards expr)) = Vars' (bound' gs) (free' gs ^+ (freeVars' expr ^- bound' gs)) where gs = allVars' guards
 
   allVars' _ = mempty -- New ctor.
 
--- a/src/GHC/Util/Unify.hs
+++ b/src/GHC/Util/Unify.hs
@@ -167,7 +167,10 @@ unifyPat' nm (LL _ (VarPat _ x)) (LL _ (VarPat _ y)) =
   Just $ Subst' [(rdrNameStr' x, strToVar'(rdrNameStr' y))]
 unifyPat' nm (LL _ (VarPat _ x)) (LL _ (WildPat _)) =
   let s = rdrNameStr' x in Just $ Subst' [(s, strToVar'("_" ++ s))]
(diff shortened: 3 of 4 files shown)


shayne-fletcher-da:
(4) Also, please take note of this newly added block:
```haskell
unifyPat' nm (LL _ (ConPatIn x _)) (LL _ (ConPatIn y _)) | rdrNameStr' x /= rdrNameStr' y =
  Nothing
```

This gets
```haskell
foo x t f = case x of {True -> t; False -> f}
```
to work correctly.

I think this comes down to `toConstr` on all `OccName` values returning `"{abstract:OccName}"` breaking `gzip` in that where `OccName` values are being unified, the condition of the first equation is never satisfied so `gzip` can never return `Nothing`:
```haskell
gzip :: Data a => (forall b . Data b => b -> b -> c) -> a -> a -> Maybe [c]
gzip f x y | toConstr x /= toConstr y =
             Nothing
           | otherwise =
             Just $ zipWith op (gmapQ Box x) (gmapQ Box y)
         -- unsafeCoerce is safe because gmapQ on the same constr gives the same fields
         -- in the same order
    where op (Box x) (Box y) = f x (unsafeCoerce y)
```

src/GHC/Util/HsExpr.hs
```diff
@@ -59,10 +59,15 @@ noSyntaxExpr' =
...
+parSec' x@(LL _ SectionL{}) = noLoc $ HsPar noExt x
+parSec' x@(LL _ SectionR{}) = noLoc $ HsPar noExt x
+parSec' x = x
+
 dotApps' :: [LHsExpr GhcPs] -> LHsExpr GhcPs
 dotApps' [] = error "GHC.Util.HsExpr.dotApps', does not work on an empty list"
-dotApps' [x] = x
-dotApps' (x : xs) = dotApp' x (dotApps' xs)
+dotApps' [x] = parSec' x
+dotApps' (x : xs) = dotApp' (parSec' x) (dotApps' xs)
```
ndmitchell: The bracketing here is wrong.

ndmitchell:
(0) I can live with the Grep tests being wrong. Grep is a side function that no one knows about.

The lack of Test.hs is a bit sad. That's where all the module name tests live. Maybe it needs porting or the test rewriting to capture the same behaviour elsewhere? Keen not to block this change on that, so we can accept it and fix later.

(1) The difference in spaces between old and new is fine - the new looks more correct.

(3) Looks like a bug that your bracket check on types is not correct. It should insert brackets, but doesn't.

(2) Those brackets shouldn't be there. I suspect it's the fact you are paranthesising sections in dotApps, which you shouldn't do.

shayne-fletcher-da:
> (0) I can live with the Grep tests being wrong. Grep is a side function that no one knows about.

It fails because `Grep.hs` hasn't been updated to support GHC. Needs porting.

> The lack of Test.hs is a bit sad. That's where all the module name tests live. Maybe it needs porting or the test rewriting to capture the same behaviour elsewhere?

Something like that. Was looking for your steer on what you want here.

> Keen not to block this change on that, so we can accept it and fix later.

Ok.

> (1) The difference in spaces between old and new is fine - the new looks more correct.

Ok.

> (3) Looks like a bug that your bracket check on types is not correct. It should insert brackets, but doesn't.

Yes. I haven't looked into it.

> (2) Those brackets shouldn't be there. I suspect it's the fact you are paranthesising sections in dotApps, which you shouldn't do.

No, it's unrelated. 

I agree that parenthesizing the sections is an unprincipled hack. It went in on the observation that somehow naked sections are getting thrown up in the hints and without it rules like `not . (a /=)` fail to unify. It was treating the symptom not the issue and so I've removed it.

Thus we have 3 problems left to solve : (a) sections (`not . (a /=)`), (b) over bracketing (`(concatMap f . (baz . bar))`) and (c) under bracketing  (`typeRep (Proxy :: Proxy Foo Int)`).

With this change are you comfortable landing this small step forward now?

```
TEST FAILURE IN tests/_hint_3
DIFFER ON LINE: 3
GOT :   if isNothing x then (- 1.0) else fromJust x
WANT:   if isNothing x then (-1.0) else fromJust x
FULL OUTPUT FOR GOT:
tests/brackets.hs:1:8: Warning: Use fromMaybe
Found:
  if isNothing x then (- 1.0) else fromJust x
Perhaps:
  fromMaybe (- 1.0) x

1 hint

.......
Hint names ...
Hint annotations .
TEST FAILURE (1 hints generated)
SRC: /Users/shaynefletcher/tmp/hlint/.stack-work/install/x86_64-osx/nightly-2019-08-07/8.6.5/share/x86_64-osx-ghc-8.6.5/hlint-2.2.4/hlint.yaml:816:0
INPUT: yes = foo . bar . concat . map f . baz . bar
OUTPUT: /Users/shaynefletcher/tmp/hlint/.stack-work/install/x86_64-osx/nightly-2019-08-07/8.6.5/share/x86_64-osx-ghc-8.6.5/hlint-2.2.4/hlint.yaml:1:19: Warning: Use concatMap
Found:
  concat . map f . baz . bar
Perhaps:
  (concatMap f . (baz . bar))

WANTED: concatMap f . baz . bar


TEST FAILURE (0 hints generated)
SRC: /Users/shaynefletcher/tmp/hlint/.stack-work/install/x86_64-osx/nightly-2019-08-07/8.6.5/share/x86_64-osx-ghc-8.6.5/hlint-2.2.4/hlint.yaml:826:0
INPUT: yes = not . (a ==)
WANTED: (a /=)


TEST FAILURE (0 hints generated)
SRC: /Users/shaynefletcher/tmp/hlint/.stack-work/install/x86_64-osx/nightly-2019-08-07/8.6.5/share/x86_64-osx-ghc-8.6.5/hlint-2.2.4/hlint.yaml:827:0
INPUT: yes = not . (== a)
WANTED: (/= a)


TEST FAILURE (0 hints generated)
SRC: /Users/shaynefletcher/tmp/hlint/.stack-work/install/x86_64-osx/nightly-2019-08-07/8.6.5/share/x86_64-osx-ghc-8.6.5/hlint-2.2.4/hlint.yaml:828:0
INPUT: yes = not . (a /=)
WANTED: (a ==)


TEST FAILURE (0 hints generated)
SRC: /Users/shaynefletcher/tmp/hlint/.stack-work/install/x86_64-osx/nightly-2019-08-07/8.6.5/share/x86_64-osx-ghc-8.6.5/hlint-2.2.4/hlint.yaml:829:0
INPUT: yes = not . (/= a)
WANTED: (== a)


TEST FAILURE (1 hints generated)
SRC: /Users/shaynefletcher/tmp/hlint/.stack-work/install/x86_64-osx/nightly-2019-08-07/8.6.5/share/x86_64-osx-ghc-8.6.5/hlint-2.2.4/hlint.yaml:1000:0
INPUT: foo = typeOf (undefined :: Foo Int)
OUTPUT: /Users/shaynefletcher/tmp/hlint/.stack-work/install/x86_64-osx/nightly-2019-08-07/8.6.5/share/x86_64-osx-ghc-8.6.5/hlint-2.2.4/hlint.yaml:1:7: Suggestion: Use typeRep
Found:
  typeOf (undefined :: Foo Int)
Perhaps:
  typeRep (Proxy :: Proxy Foo Int)

WANTED: typeRep (Proxy :: Proxy (Foo Int))
```

src/GHC/Util/HsExpr.hs
shayne-fletcher-da: Don't disagree (it happens that it helps with a particular set of tests but is not principled). See my notes below.

shayne-fletcher-da:
The
```
Found:
  concat . map f . baz . bar
Perhaps:
  (concatMap f . (baz . bar))
```
is fixed. This slight difference comes down to HSE not counting `.` as a term (it's a `QVarOp`) and so differs from `children e :: [LHSExpr GhcPs]`.

ndmitchell:
Sounds good. I've merged this thus far, and we can layer the next two fixes on top.

PR #803, opened 2019-12-04
Bump ghc-lib

cocreature:
The new release has set exposed to False in the cabal file which
should fix some issues in IHaskell which also depends on HLint.

Thanks for the pull request!

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -3,6 +3,6 @@
 # * https://github.com/commercialhaskell/stackage/issues/4731
 resolver: nightly-2019-08-07
 packages: [.]
-extra-deps: [ghc-lib-parser-8.8.1]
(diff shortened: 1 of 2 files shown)


stack.yaml
```diff
@@ -3,6 +3,6 @@
 # * https://github.com/commercialhaskell/stackage/issues/4731
 resolver: nightly-2019-08-07
 packages: [.]
-extra-deps: [ghc-lib-parser-8.8.1]
+extra-deps: [ghc-lib-parser-8.8.1.20191204.tar.gz]
```
ndmitchell: Why the .tar.gz at the end?

cocreature: oops, my fault. That was leftover from testing. I removed it.

ndmitchell:
Thanks!

PR #806, opened 2019-12-08
Convert match

shayne-fletcher-da:
In GHC, sections require parens to be operators. With this fix we are left with only one outstanding problem : (3) under-bracketing in `typeRep (Proxy :: Proxy Foo Int)`.


ndmitchell:
Thabks!

PR #807, opened 2019-12-13
Convert match

shayne-fletcher-da:
Implements a fix for the remaining test failure.  The issue is that there's never been a need to bracket the interior of a type applied to a type application after substitution using HSE because its pretty printer implicitly printed brackets. Now with this, GHC gives `typeRep (Proxy :: Proxy (Foo Int))` as we'd hope.


ndmitchell:
Sweet! Are we in a position to switch to the new match now?

shayne-fletcher-da:
Yes. I'll put a PR together for the switch.

Issue #809, opened 2019-12-15
Reinstate disabled tests (that rely on match rules)


shayne-fletcher-da:
Placeholder.

PR #810, opened 2019-12-15
Convert match

shayne-fletcher-da:
This PR switches the match rule implementation over to the ghc-lib version. It disables some tests (reinstating them is tracked in issue https://github.com/ndmitchell/hlint/issues/809) .


ndmitchell:
Sweet - this isn't nearly as brutal as I imagined.

PR #812, opened 2019-12-22
Add rules for any/all/or/and corresponding to those for foldMap/fold

josephcsible:


ndmitchell:
Hmm, not so sure about this one. While correct, treating the Maybe/Either as Foldable always surprises me.

josephcsible:
Even though I put them in `generalise-for-conciseness`? Isn't that where you usually put "surprising" rules like these?

ndmitchell:
Ah, the diff didn't show me enough context to see that bit. They make perfect sense there!

PR #814, opened 2019-12-31
Add a rule to use `find`

josephcsible:


ndmitchell:
Great idea! Just needs a quick fix then i'll merge

josephcsible:
Do you mean something you'll fix yourself, or something you want me to fix in it? If the latter, what is it?

ndmitchell:
I wrote the comments, but as part of a review that I forgot to submit, so you couldn't see it. I've pushed the fixes myself - you need to write `x` if you want it to be a wildcard, since `xs` is more than one letter, so matches only the literal `xs`.

ndmitchell:
And thanks for the patch!

PR #816, opened 2020-01-06
Recognise HsType tyCon as Type RType in bracket refactorings

bubba:
Thanks for the pull request!


bubba:
Fixes https://github.com/haskell/haskell-ide-engine/issues/1554

A sample case where this doesn't work

```haskell
foo :: (String)
foo = "asdf"
```

```bash
hlint Foo.hs --refactor
```

ndmitchell:
Thanks! I'll be honest that I don't completely understand this, but it seems plausible. CC @shayne-fletcher-da who wrote the code.

PR #817, opened 2020-01-06
Add the ability to blacklist identifiers from a module

taktoa:
This allows you to prevent someone from importing a specific identifier from a module. This is useful if you want to, for example, maintain a policy that people never import identifiers that are reexported (e.g.: in MTL).

I am not sure how the hlint testing framework works, so I have not been able to test this code.

### DIFF
--- a/src/Hint/Restrict.hs
+++ b/src/Hint/Restrict.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TypeFamilies #-}
 
@@ -18,6 +20,7 @@ import Hint.Type(ModuHint,ModuleEx(..),Idea(..),Severity(..),warn',rawIdea')
 import Config.Type
 
 import Data.Generics.Uniplate.Operations
+import qualified Data.Set as Set
 import qualified Data.Map as Map
 import Data.List
 import Data.Maybe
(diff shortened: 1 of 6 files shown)


googleson78:
You can run
`stack run hlint test`
to do tests (or the equivalent `cabal` command)

ndmitchell:
Can you rebase/merge `master` into this so the tests will run properly.

ndmitchell:
Can you give an example of the kind of thing this hint will spot, and not spot, so I can understand it a bit better? Does it only work with explicit imports?

taktoa:
The point is that if you do `badidents: ["foo", "bar"]` then you won't be able to import `foo` and `bar` explicitly from that module. You can still import them implicitly, but in practice many codebases have a rule that disallows implicit imports.

ndmitchell:
Did you try banning the functions with a hint with `Foo.bar` as the LHS, and "bad user" as the RHS? That way it should work even if they import qualified or don't use explicit import lists.

taktoa:
I haven't tried that, but the goal here is to prevent users from importing reexported functions, so I am not sure if haskell-src-exts will resolve the reexport and ban the original identifier. In general, it just seems weird to hack the hint mechanism into providing a facility that users will naturally expect to be in the import management mechanism.

taktoa:
Also, I think this technically allows you to ban explicit imports of type synonyms and pattern synonyms, and I don't think the hint system can do that.

ndmitchell:
OK, that makes sense. Unfortunately there is nearly no testing for the restriction mechanism, so there's no good place to put it, so we can live dangerously and not have tests (which I've been fine with up until now - so can't complain about!) or as a follow up patch someone can write tests (which would be great, but you are under no obligation to do so).

It would be good to get just a sentence of two about this in the README though https://github.com/ndmitchell/hlint#restricting-items

src/Hint/Restrict.hs
```diff
@@ -105,22 +109,52 @@ checkPragmas modu flags exts mps =
...
 
+importListToIdents :: IE GhcPs -> [String]
+importListToIdents =
+  catMaybes .
+  \case (IEVar _ n)              -> [fromName n]
+        (IEThingAbs _ n)         -> [fromName n]
+        (IEThingAll _ n)         -> [fromName n]
+        (IEThingWith _ n _ ns _) -> fromName n : map fromName ns
+        IEModuleContents {}      -> error "importListToIdents: this is only supposed to happen in an export list"
+        IEGroup {}               -> []
```
ndmitchell: Could we just use `_ -> []` for all these? I would rather be less tied to the GHC API if possible.

src/Hint/Restrict.hs
```diff
@@ -105,22 +109,52 @@ checkPragmas modu flags exts mps =
...
     ]
 
+importListToIdents :: IE GhcPs -> [String]
+importListToIdents =
+  catMaybes .
+  \case (IEVar _ n)              -> [fromName n]
+        (IEThingAbs _ n)         -> [fromName n]
+        (IEThingAll _ n)         -> [fromName n]
+        (IEThingWith _ n _ ns _) -> fromName n : map fromName ns
+        IEModuleContents {}      -> error "importListToIdents: this is only supposed to happen in an export list"
```
ndmitchell: No need to error here.

ndmitchell:
Thanks!

PR #819, opened 2020-01-11
Fix bifunctor hints

josephcsible:
Use single character variables so they work with other names, and fix swapped variables in the bimap-bimap hint.


ndmitchell:
Thanks! Good catches.

PR #820, opened 2020-01-11
Eta-expand hints

josephcsible:
Eta-expand instead of using composition, so that we can match them both ways.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -278,15 +278,15 @@
     - hint: {lhs: "\\x y -> f (x,y)", rhs: curry f}
     - hint: {lhs: "\\(x,y) -> f x y", rhs: uncurry f, note: IncreasesLaziness}
     - warn: {lhs: f (fst p) (snd p), rhs: uncurry f p}
-    - warn: {lhs: ($) . f, rhs: f, name: Redundant $}
+    - warn: {lhs: ($) (f x), rhs: f x, name: Redundant $}
     - warn: {lhs: (f $), rhs: f, name: Redundant $}
     - warn: {lhs: (Data.Function.& f), rhs: f, name: Redundant Data.Function.&}
     - hint: {lhs: \x -> y, rhs: const y, side: isAtom y && not (isWildcard y)}
         # If any isWildcard recursively then x may be used but not mentioned explicitly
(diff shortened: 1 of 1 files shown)


data/hlint.yaml
```diff
@@ -278,15 +278,15 @@
...
     - warn: {lhs: (f $), rhs: f, name: Redundant $}
     - warn: {lhs: (Data.Function.& f), rhs: f, name: Redundant Data.Function.&}
     - hint: {lhs: \x -> y, rhs: const y, side: isAtom y && not (isWildcard y)}
         # If any isWildcard recursively then x may be used but not mentioned explicitly
     - warn: {lhs: flip f x y, rhs: f y x, side: isApp original}
     - warn: {lhs: id x, rhs: x}
-    - warn: {lhs: id . x, rhs: x, name: Redundant id}
-    - warn: {lhs: x . id, rhs: x, name: Redundant id}
+    - warn: {lhs: id (f x), rhs: f x, name: Redundant id}
+    - warn: {lhs: f (id x), rhs: f x, name: Redundant id}
```
ndmitchell: There is already a rule `id x ==> x`. I think this ends up duplicating that?

data/hlint.yaml
```diff
@@ -278,15 +278,15 @@
...
+    - warn: {lhs: ($) (f x), rhs: f x, name: Redundant $}
     - warn: {lhs: (f $), rhs: f, name: Redundant $}
     - warn: {lhs: (Data.Function.& f), rhs: f, name: Redundant Data.Function.&}
     - hint: {lhs: \x -> y, rhs: const y, side: isAtom y && not (isWildcard y)}
         # If any isWildcard recursively then x may be used but not mentioned explicitly
     - warn: {lhs: flip f x y, rhs: f y x, side: isApp original}
     - warn: {lhs: id x, rhs: x}
-    - warn: {lhs: id . x, rhs: x, name: Redundant id}
-    - warn: {lhs: x . id, rhs: x, name: Redundant id}
+    - warn: {lhs: id (f x), rhs: f x, name: Redundant id}
```
ndmitchell: Dupe of `id x ==> x` ?

josephcsible:
Fixed, and merge conflicts resolved.

ndmitchell:
Thanks!

Issue #699, opened 2019-07-03
Binaries using hlint have hlint_datadir hardcoded to build directory, not destroot


essandess:
I’m trying to port [IHaskell](https://github.com/gibiansky/IHaskell/) to MacPorts. IHaskell uses `hlint`.

This issue is that both `stack` and `cabal` builds create binaries that point to the build directory in `.stack` or `.cabal-sandbox`, not the install directory with prefix, say, `/opt/local/bin`. After the build, these temporary build directories are discarded, and neither `ihaskell` nor `hlint` binaries work.

I see that there is an `hlint --data-dir` option:

https://github.com/ndmitchell/hlint/blob/c62b61696621cd5ebba408ec5bcc2fcdadb603cc/cc/Engine.hs#L65-L70

It seems like it should be straightforward, but I’m not sure how to build projects that use hlint to set its data directory to have prefix `/opt/local` so that destrooting the build will work.

Any pointers would be appreciated.

Related:
* https://github.com/macports/macports-ports/pull/4706
* https://github.com/gibiansky/IHaskell/issues/1061
* https://github.com/ndmitchell/hlint/issues/503

ndmitchell:
I think this is a cabal/stack thing. When configuring cabal, you pass `--prefix` or some other flag and it will wire that into HLint. The `--data-dir` flag is mostly there as a backup, in case you get into a state where you can't do that.

vaibhavsagar:
It looks like Cabal has a [`--datadir`](https://www.haskell.org/cabal/users-guide/installing-packages.html#cmdoption-setup-configure-datadir) flag. I wasn't able to find the equivalent for Stack.

essandess:
I see that there’s a (four year old!) upstream request for a `stack --prefix` option at https://github.com/commercialhaskell/stack/issues/848.

I don’t see a `cabal --prefix` option, but there is a `runhaskell --prefix`.

Any suggestions on the correct build calls for `hlint` or `ihaskell` that use `--prefix`?

vaibhavsagar:
I think you want `--data-dir`: `cabal configure --data-dir=<path>` or `cabal install --data-dir=<path>`.

essandess:
Thanks again! It's all working now. See https://github.com/macports/macports-ports/pull/4706

I'll defer to you if you want to close this issue. I was able to work around it with a lot of hackery, but both upstream `cabal` and `stack` need destroy capability.

ndmitchell:
[...]

Issue #620, opened 2019-04-14
Depend on Hlint as a library


fendor:
This problem is discussed in [HIE #1143](https://github.com/haskell/haskell-ide-engine/issues/1143).

If you depend on hlint as a library, it demands that the location of its data-files are known at compile-time.
This is fine, if you are using binary packages of hlint or build it with cabal/stack from source.
However, if you depend on hlint as a library, and want to distribute a static executable of your application, hlint will usually fail, because the location of its data-files is known at compile-time and baked into the application. Thus, it might happen that the data-files are not present on the installation system at the expected location.

To solve this problem, the following approaches can be chosen:

* The application is responsible for distributing the data-files and configuring hlint correctly to find it.
    * Problem: How to configure the paths correctly?
* Patch hlint to provide the data-files via source code, e.g. translate the hlint.yaml into a haskell source record and drop the data-files required for hlint. For customisation, a hlint.yaml can be written to [$XDG_CONFIG_HOME](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) on startup.
    * Problem: Might bloat the executable?
* Applications should not depend on hlint as a library and create a run-time dependency for it.
    * Problem: Opens applications up for version conflicts.

What is your view on this issue? Which way would you suggest, and is there another, possibly better, way to address this issue?

ndmitchell:
My current mental model is the first one. I believe HLint already looks at several places for the data files, and there is a flag --datadir which let's you specify where to find data files. There are also places in the API that let you specify where files are. 

I'm not such a fan of anything TH related as my experience is TH makes things deeply unpleasant - failures on users systems, slow compile, version changes etc. 

Does that work for you?

fendor:
We can make it work, it is a bit complicated though, since we have to distribute the data-files somehow.
I was thinking about using `file-embed`, to embed the data-files at compile-time and write them into some location and configure hlint accordingly. But this requires TH and, afaik, increases compile-time even more.
Another way, we can transform the data-files into default values ourselves in the project and write them to disk when we need them and configure hlint accordingly. This might, however, introduce a new hazard to keep hlint configurations up to date with upstream.
What may be the more sensible way? 

To patch hlint, I wouldn't have thought about using TH, but rather replacing the data-files needed at run-time with default values in the executable.

ndmitchell:
[...]

PR #824, opened 2020-01-14
Embed data files

Avi-D-coder:
closes:#699

Embeds data files so the hlint executable is self contained.
This allows for easy packaging with stack and cabal.
And enables hlint to be used as a library without shipping data files.

From https://github.com/ndmitchell/hlint/issues/620#issuecomment-570777653
> ndmitchell would you reconsider the TH approach?
> 
> It's not just nicer for using `hlint` as a library it's nicer when using the executable too. I'm on a 256G laptop and I delete `~/.stack` when ever it gets past 10G. I install hlint using `stack install`, so when I delete `~/.stack` `hlint` has to be rebuilt, just to get 3 data files.
> 
> Here's a [commit](https://github.com/Avi-D-coder/hlint/commit/b50d9cf08e409d4ca20abf37fb541baebff60391) using `file-embed`. The TH is relegated to one file. The executable actually gets slightly smaller `48117184` to `48020968` (GHC 8.8).

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -54,6 +54,8 @@ library
         base == 4.*, process, filepath, directory, containers,
         unordered-containers, vector, text, bytestring,
         transformers,
+        file-embed,
+        utf8-string,
         data-default >= 0.3,
         cpphs >= 1.20.1,
         cmdargs >= 0.10,
@@ -100,6 +102,7 @@ library
         Refact
         Timing
         CC
+        EmbedData
         Config.Compute
         Config.Haskell
         Config.Read
--- a/src/CmdLine.hs
+++ b/src/CmdLine.hs
@@ -30,6 +30,7 @@ import System.Info.Extra
 import System.Process
 import System.FilePattern
 
+import EmbedData
 import Util
 import Paths_hlint
 import Data.Version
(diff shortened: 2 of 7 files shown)


Avi-D-coder:
`TemplateHaskell` is confined to `src/EmbedData.hs`. If it causes a problem on a particular platform, an alternative module could be conditionally compiled with the old `readFile $ cmdDataDir </> "default.yaml"` method.

src/CmdLine.hs
```diff
@@ -217,12 +217,11 @@ mode = cmdArgsMode $ modes
 --   We want more important hints to go last, since they override
 cmdHintFiles :: Cmd -> IO [FilePath]
 cmdHintFiles cmd = do
-    let explicit1 = [cmdDataDir cmd </> "hlint.yaml" | null $ cmdWithHints cmd]
```
ndmitchell: If `cmdWithHints` is specified then we don't use `cmdDataDir` for the hints. It seems in your patch that changes?

src/Config/Read.hs
```diff
@@ -6,12 +6,14 @@ import Config.Haskell
...
 
 readFilesConfig :: [(FilePath, Maybe String)] -> IO [Setting]
 readFilesConfig files = do
-        yaml <- mapM (uncurry readFileConfigYaml) yaml
+        yaml <- mapM (uncurry readFileConfigYaml) yaml'
         haskell <- mapM (uncurry readFileConfigHaskell) haskell
         return $ concat haskell ++ settingsFromConfigYaml yaml
     where
         (yaml, haskell) = partition (\(x,_) -> lower (takeExtension x) `elem` [".yml",".yaml"]) files
+        yaml' = if any ((takeFileName (fst hlintYaml) ==) . takeFileName . fst) yaml then yaml else hlintYaml : yaml
```
ndmitchell: I don't get what this line is doing

hlint.cabal
```diff
@@ -54,6 +54,8 @@ library
         base == 4.*, process, filepath, directory, containers,
         unordered-containers, vector, text, bytestring,
         transformers,
+        file-embed,
+        utf8-string,
```
ndmitchell: Where is the utf8-string dependency coming from?

Avi-D-coder: embed stores a `ByteString` of the UTF-8 data files, so we have to convert it back.

src/CmdLine.hs
Avi-D-coder: My bad, should be fixed now

ndmitchell:
That's way nicer. Thanks a lot!

Avi-D-coder:
@ndmitchell Could you push a release to hackage so hie can use this?

Avi-D-coder:
@ndmitchell Could you push a release to hackage so hie can use this?

ndmitchell:
Sorry @Avi-D-coder - I've been on holiday. Should have this done by this evening.

Avi-D-coder:
Great thanks

PR #825, opened 2020-01-15
Fix #822

### LINKED ISSUE #822: Why `isVar f` here?
https://github.com/ndmitchell/hlint/blob/1e22e318fa40411d7782d391a7d9396d9fc72675/data/hlint.yaml#L173

I assume `isVar f` is here for a reason, but I can't figure out what that reason is.

josephcsible:


ndmitchell:
I see the test failure:

```
INPUT: yes = map (uncurry (+)) $ zip [1 .. 5] [6 .. 10]
OUTPUT: /home/travis/.cabal/share/x86_64-linux-ghc-8.8.1/hlint-2.2.7/hlint.yaml:1:7: Warning: Use zipWith
Found:
  map (uncurry (+)) $ zip [1 .. 5] [6 .. 10]
Perhaps:
  zipWith (+) [1 .. 5] [6 .. 10]
OUTPUT: /home/travis/.cabal/share/x86_64-linux-ghc-8.8.1/hlint-2.2.7/hlint.yaml:1:7: Suggestion: Use zipWith
Found:
  map (uncurry (+)) $ zip [1 .. 5] [6 .. 10]
Perhaps:
  zipWith (curry (uncurry (+))) [1 .. 5] [6 .. 10]
WANTED: zipWith (+) [1 .. 5] [6 .. 10]
```

We should only be hinting one of those. I don't see why we hint the second, as its an app? Or have we got confused because its actually a bracketed app? Also, do we need a curry/uncurry rule?

josephcsible:
I have no idea why we got the second either. As for a curry/uncurry rule, isn't that what you just merged in #826?

ndmitchell:
The problem was there was another rule suggesting zipWith if it was an uncurry as the first argument. I removed that and updated the tests in c67afbcd5ca56cecc1f69188fa8186dc55e6d721.

PR #827, opened 2020-01-16
Simplify fixity calculations

shayne-fletcher-da:
This PR simplifies the `GHC.Util.Refact.Fixity` module and in doing so enables us to remove `GHC.Util.Language.Haskell.GHC.ExactPrint.Types` and `GHC.Util.Refact.Utils`.  That in turn allows us to remove the dependencies on syb and mtl.


ndmitchell:
Thanks, I like the simplification. Unfortunately the patch is hard to review. Can you tell me you're reasonably sure it right? What lead you to this simplification? Easy reduction? Some clever insight? Reusing ideas from elsewhere? Then I'll be happy to merge.

shayne-fletcher:
I'm 100% sure it's no more wrong that it was before 😁 You'll recall, we copied the algorithm from Refact which in turn has been copied from the GHC renamer. The renamer version is in the presence of the TcRn monad (but actually neither looks into or modifies it). The Refact algorithm replaces that monad with a state monad that maintains a map of annotations as it does the fixity adjustments. That makes sense for Refact but for us, we never used the annotations map! We passed an empty one as input and received an empty one as output incidentally along with the fixity adjusted parse tree. Eliminating the map means you can then eliminate the monad and all the supporting types, functions modules and dependencies (from exactprint etc.) that at the time we introduced this you characterized as "the only thing that makes you sad about this code is all that".

The background for what motivated me to go in and make these changes I'd like to discuss with you offline!

ndmitchell:
Thanks, makes sense!

PR #830, opened 2020-01-20
Introduce dependency on ghc-lib-parser-ex

shayne-fletcher-da:
With this PR, we are off to the gentlest of starts introducing a dependency on ghc-lib-parser-ex. Cabal and stack files are updated and the parse function logic begins introduction of its usage.


hlint.cabal
```diff
@@ -67,14 +67,15 @@ library
         extra >= 1.6.6,
         refact >= 0.3,
         aeson >= 1.1.2.0,
-        filepattern >= 0.1.1
+        filepattern >= 0.1.1,
+        ghc-lib-parser-ex == 8.8.2
```
ndmitchell: Can I not be 8.8.* anymore? If not, I'm going to have to rerelease HLint with 8.8.3

hlint.cabal
```diff
@@ -67,14 +67,15 @@ library
...
     if !flag(ghc-lib) && impl(ghc >= 8.8.0) && impl(ghc < 8.9.0)
         build-depends:
           ghc == 8.8.*,
           ghc-boot-th
     else
         build-depends:
-          ghc-lib-parser == 8.8.*
+          ghc-lib-parser == 8.8.2
```
ndmitchell: Ditto

src/GHC/Util.hs
```diff
@@ -79,21 +69,18 @@ parseFileGhcLib filename str flags =
               if takeExtension filename /= ".lhs" then str else unlit filename str
     parseState = mkPState flags buffer location
 
+-- TODO(SF): GhclibParserEx.parsePragmasIntoDynFlags doesn't take
+-- enabled/disabled extensions into consideration yet.
```
ndmitchell: Isn't that a super important thing to do?

src/GHC/Util.hs
```diff
@@ -37,40 +37,30 @@ import GHC.Util.RdrName
...
 parseImportGhcLib :: String -> DynFlags -> ParseResult (LImportDecl GhcPs)
-parseImportGhcLib = parseGhcLib Parser.parseImport
+parseImportGhcLib = GhclibParserEx.parseImport
 
-parseFileGhcLib :: FilePath
-                -> String
-                -> DynFlags
-                -> ParseResult (Located (HsModule GhcPs))
+-- TODO(SF): GhclibParserEx.parseFile doesn't take 'unlit' into
+-- consideration yet.
```
ndmitchell: This is OK because we run unlit ourselves before?

stack.yaml
```diff
@@ -3,6 +3,14 @@
 # * https://github.com/commercialhaskell/stackage/issues/4731
 resolver: nightly-2019-08-07
 packages: [.]
-extra-deps: [ghc-lib-parser-8.8.1.20191204]
+extra-deps: [ghc-lib-parser-8.8.2, ghc-lib-parser-ex-8.8.2]
```
ndmitchell: Are these both in the latest stackage nightly? If so, can they be deleted from here - and potentially the whole stack.yaml deleted?

hlint.cabal
shayne-fletcher: We can leave `ghc-lib-parser` at 8.8.*.

I need to be more specific for `ghc-lib-parser-ex` for now because otherwise I could release new versions which `hlint` would then pick up automatically perhaps breaking it until its adapted.

My thinking then was to pin them both for the moment and take responsibility for caring for them during this time in the face of upgrades.

hlint.cabal
shayne-fletcher: Yes. See above.

src/GHC/Util.hs
shayne-fletcher: Right - I will enhance the `ghc-lib-parser-ex` implementation so that we can replace this version. Just wanted to get `ghc-lib-parser-ex` seated first before new changes.

src/GHC/Util.hs
shayne-fletcher: The functions in this file are working around the inadequacies in the `ghc-lib-parser-ex` versions. When `ghc-lib-parser-ex` catches up, the todos and indeed these functions themselves can go away.

hlint.cabal
ndmitchell: Hmm, that's going to be problematic in the long run - I wonder if ghc-lib-parser-ex should be `8.8` as the GHC version (with the promise that it works with the whole 8.8 series), and then the next digit should be the ghc-lib-parser-ex breaking version (starting at 0), and the final one the minor revision (starting at 0 and increasing), then i'd pin to `8.8.0.*`. Thoughts?

hlint.cabal
ndmitchell: If we follow my suggestion above this can stay at `8.8.*`

src/GHC/Util.hs
ndmitchell: Makes sense - sounds good.

stack.yaml
shayne-fletcher: That's a good question. `https://www.stackage.org/nightly-2020-01-21` :
Screen Shot 2020-01-21 at 6 08 37 AM
I'll  delete the extra deps. It would be helpful for me to retain the stack.yaml though for my hlint development if that's ok?

src/GHC/Util.hs
ndmitchell: The ghc versions of unlit and cpphs both call external tools and were a pig to integrate into ghc-lib. The ones in this package use cpphs which doesn't require external tools, which makes the more robust, doesn't require a GHC installation etc. I'd rather stick with the cpphs ones, and I suspect that should stay in hlint rather than move to GHC.

shayne-fletcher: Actually, we've never run `unlit` in the context of this function before neither in hlint or ghc-lib-parser-ex. The context in hlint for calling it is such that we've never needed to. Do you think this family of functions should in general `unlit`?

stack.yaml
ndmitchell: My standard approach is to generate a stack with `stack init`, and put that on gitignore, so then people who want to dev with stack can, and my CI can always test it works with stackage head. Does that work for you?

src/GHC/Util.hs
shayne-fletcher: So is the net result saying, let's stay with the status quo and `ghc-lib-parser-ex` doesn't do `unlit`? That would work for me.

stack.yaml
shayne-fletcher: Yes, that would be fine. I'll delete stack.yaml accordingly.

hlint.cabal
shayne-fletcher: Like `ghc-lib`, I also intend to keep `ghc-lib-parser-ex` working against GHC head builds (e.g. `ghc-lib-parser-0.20200121`). What to do?! 😁

shayne-fletcher: > Can I not be 8.8.* anymore? If not, I'm going to have to rerelease HLint with 8.8.3

May I suggest we go with this approach for now? Careful management of the compatibility of these dependencies is guaranteed and there's no way we're going to miss an upgrade of `ghc-lib-parser` 😉 That said, still open to adopting a better approach.

hlint.cabal
shayne-fletcher: Yes, see below.

shayne-fletcher:
That's unfortunate. Removing stack.yaml causes this:
Screen Shot 2020-01-21 at 7 07 13 AM

If it is what I think it is, it's fixed in 8.10. What shall we do in the meantime?

googleson78:
@shayne-fletcher 
What's the issue you encountered on win?

shayne-fletcher:
> @shayne-fletcher
> What's the issue you encountered on win?

On this branch, we have deleted `stack.yaml`. If you then,
```
stack --resolver=nightly-2020-01-21 init
stack build
```
the build terminates abnormally with that error. I assume (given that nature of the message), that it relates to https://gitlab.haskell.org/ghc/ghc/issues/17599 and https://gitlab.haskell.org/ghc/ghc/issues/17575) and now exhibits because the given resolver gives a ghc-8.8.1 vs a ghc-8.6.5 that the now deleted stack.yaml resolver was providing.

shayne-fletcher:
In the last commit:
  - hlint.cabal sets `ghc-lib-parser==8.8.*`
  - unwound the deletion of stack.yaml (due to the Windows failure).

shayne-fletcher:
> That's unfortunate. Removing stack.yaml causes this:
> [...]
> If it is what I think it is, it's fixed in ~8.10~ 8.8.2. What shall we do in the meantime?

We resolved off ticket to wait for 8.8.2 before removing it.

PR #834, opened 2020-01-24
Move not out of any and all

josephcsible:


josephcsible:
I'm open to suggestions for less confusing names for these.

googleson78:
Are these going to be turned on by default? I think this is mostly a preference thing.

wrt. better name: "de morgan law" ?

ndmitchell:
While equivalent, I think these might be confusing - often when I'm writing logic I'm either thinking of a forall or an existential - and while they are related by a negation, often one feels more natural than the other. If you see the teaching section at https://github.com/ndmitchell/hlint/blob/ed7d5880414a10f7e9d714ff1f34f0988bd67b25/data/hlint.yaml#L826-L835, I wonder if they'd be more applicable there?

I'm not convinced by the name de morgan though. As someone with a CS background, it makes perfect sense. But I know many Haskell programmers who learnt Haskell without that background, so wouldn't understand it as intuitively.

josephcsible:
Hmm, this isn't De Morgan's Law itself, but rather a generalization of it. Also, re putting it in teaching, isn't this useful even in production programs because for very long lists, saving the `not .` for each element will add up?

ndmitchell:
Hmm, I guess the performance aspect makes it more compelling... I also note there are two instances in the HLint codebase itself (which is why CI has failed) - they need fixing. Maybe a name of "Hoist not" would be better, since it's hoisting things out of a loop?

ndmitchell:
Looks fine to me. Thoughts @googleson78 ?

josephcsible:
Hmm, this new failure seems to be spurious.

ndmitchell:
@josephcsible probably is spurious - I gave it a rerun. I don't imagine your hint would break only Windows, so happy to merge with this one failure (after giving @googleson78 a chance to weigh in)

ndmitchell:
OK, merged - thanks for the change.

googleson78:
Sorry for not responding! I didn't actually expect to be asked directly :sweat_smile: 

(for the record: I would prefer it if this was left to preference by default, but the performance reasoning is indeed compelling. Maybe we can have a magical compiler one day which does all sorts of stuff like this!)

ndmitchell:
Having tried this hint on my code it was significantly better than I expected. Some of the time I was doing not outside as well and they cancelled. Sometimes it was a nested any/any and I pulled the not all the way out. So I'm glad it's included!

PR #838, opened 2020-01-25
Avoid using head with a sloppy default

josephcsible:


josephcsible:
Looks like the build is failing because HLint itself triggers this rule. The code where it does so is the same part of the code I'm touching in #834, so I'll wait until it's merged to fix this one to avoid an unnecessary merge conflict.

ndmitchell:
`fromMaybe x (listToMaybe y)` is quite a lot longer and (to me) less intuitive than `head (x ++ [y])`. If you use safe you can use `headDef` which is cleaner, but without it, I don't see a nicer pattern.

josephcsible:
I thought up an alternative. Do you like this replacement better?

ndmitchell:
Not really - that seems a lot more confusing than the Maybe variant.

josephcsible:
I'll table this for now unless/until there's something better then.

PR #843, opened 2020-01-27
Avoid the null...head anti-pattern

josephcsible:
Checking `null` of a list and then using `head` if it's not is an anti-pattern. Case-match instead.

### DIFF
--- a/src/GHC/Util/Scope.hs
+++ b/src/GHC/Util/Scope.hs
@@ -19,6 +19,7 @@ import GHC.Util.RdrName
 import Outputable
 
 import Data.List
+import Data.List.Extra
 import Data.Maybe
 
 -- A scope is a list of import declarations.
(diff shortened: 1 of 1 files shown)


src/GHC/Util/Scope.hs
```diff
@@ -66,10 +67,10 @@ scopeMatch' (a, x) (b, y)
 -- scope that will refer to the same thing. If the resulting name is
 -- ambiguous, pick a plausible candidate.
 scopeMove' :: (Scope', Located RdrName) -> Scope' -> Located RdrName
-scopeMove' (a, x@(fromQual' -> Just name)) (Scope' b)
-  | null imps = head $ real ++ [x]
-  | all ideclQualified imps = noLoc $ mkRdrQual (unLoc $ head (mapMaybe ideclAs imps ++ map ideclName imps)) name
-  | otherwise = unqual' x
+scopeMove' (a, x@(fromQual' -> Just name)) (Scope' b) = case imps of
+  [] -> head $ real ++ [x]
+  (imp:_) | all ideclQualified imps -> noLoc $ mkRdrQual (unLoc . fromMaybe (ideclName imp) $ firstJust ideclAs imps) name
```
ndmitchell: No need for brackets here.
```suggestion
  imp:_ | all ideclQualified imps -> noLoc $ mkRdrQual (unLoc . fromMaybe (ideclName imp) $ firstJust ideclAs imps) name
```

src/GHC/Util/Scope.hs
```diff
@@ -66,10 +67,10 @@ scopeMatch' (a, x) (b, y)
...
+          | otherwise -> unqual' x
```
ndmitchell:
```suggestion
    _ -> unqual' x
```
Makes the completeness of the patterns more obvious

josephcsible: Are you sure? I feel like this makes it less clear when the final case will be entered, and GHC not giving a warning is already evidence that it's complete.

ndmitchell: I'd have done it that way, but that's just my style - I'm not sufficiently fussed either way though. Let's go with how you wrote it.

PR #845, opened 2020-01-29
Handle operator applications in checkDefine'

### LINKED ISSUE #774: Don't warn about 'Redundant compare' in `(==)` and `(/=)` definitions
This is a minor improvement. It's often useful to have one implementation of compare and implement `(==)` in terms of it:

```
instance Eq MyType where
    l == r = compare l r == EQ

instance Ord MyType where
    compare l r = ...
```

However currently the hLint warns on the implementation of `(==)` and it seems that you cannot annotate instance method so you need to waive the hint for whole module.

zliu41:
Fixes #774


ndmitchell:
Many thanks!

PR #846, opened 2020-01-29
Add splitAt warnings

zliu41:


data/hlint.yaml
```diff
@@ -942,6 +945,9 @@
 # main = take (-y) x
 # main = take 4 x
 # main = let (first, rest) = (takeWhile p l, dropWhile p l) in rest -- span p l
+# main = let (first, rest) = (take n l, drop n l) in rest -- splitAt n l
+# main = fst (splitAt n l) -- take n l
+# main = snd $ splitAt n l -- drop n l
```
ndmitchell: When adding new hints such as these, I sometimes add tests, but often don't bother if they are perfectly vanilla. Up to you in future, I'm happy either way.

PR #847, opened 2020-01-29
Upgrade haskell-src-exts

shayne-fletcher:
- Newly added tests break unless on `haskell-src-exts-1.23.0`; fix cabal;
- This version isn't in the stackage nightlies yet; fix stack.yaml

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -3,7 +3,10 @@
 # * https://github.com/commercialhaskell/stackage/issues/4731
 resolver: nightly-2020-01-25
 packages: [.]
-extra-deps: [ghc-lib-parser-8.8.2, ghc-lib-parser-ex-8.8.3.0]
+extra-deps:
+  - ghc-lib-parser-8.8.2
+  - ghc-lib-parser-ex-8.8.3.0
+  - haskell-src-exts-1.23.0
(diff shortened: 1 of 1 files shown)


ndmitchell:
This was deliberate - often upgrading haskell-src-exts is problematic for other consumers, since it's got lots of quirks (including the latest versions no longer build on some Windows machines!). Therefore I tend to benefit from the latest, but not demand it in the .cabal unless otherwise would require CPP.

Adding it to stack isn't unreasonable though!

shayne-fletcher:
Ok understood. I'll roll back the change in the cabal file and push, I'll keep the stack change  - let's see.

ndmitchell:
I thought I had rolled stackage yesterday in order to fix this bug in CI. I may have inadvertently just tripped into the Windows linker bug instead and thought I'd fixed it. Thanks!

shayne-fletcher:
Right indeed, I see what's happened. Yes, we can't go to an 8.8.1/8.8.2 resolver because of the Windows linker bug - we have to wait for an 8.8.3. I'll raise a PR that rolls back to nightly-2020-01-25. Hopefully that should see Appveyor green again.

PR #848, opened 2020-01-29
Roll back to nightly-2019-08-07.

shayne-fletcher:
8.8.1 (and 8.8.2) based resolvers trip the Windows "relocation 0" bug. This change should get Appveyor passing again.


shayne-fletcher:
Yep, Appveyor fixed 🎉 .

ndmitchell:
Ah, thanks! I was just going to accept broken CI in the interim, which was indeed a crap solution.

PR #851, opened 2020-02-02
Add a rule for "maybe Nothing Just"

josephcsible:
This seems contrived at first glance, but I've seen it in real code, such as https://github.com/jgm/pandoc/blob/b47a3cdcb6b9ecc302567fa28b05459810e03095/src/Text/Pandoc/PDF.hs#L156-L159


ndmitchell:
Thanks, good idea!

PR #855, opened 2020-02-04
Remove an unnecessary use of head from reduceImports

josephcsible:


ndmitchell:
Thanks! Are you aiming to put `head` on the restrict list for HLint after these patches all land?

josephcsible:
Eventually yes, but there's still a lot more uses of `head` left that I didn't even write patches for yet. (And `tail` is on my radar too.)

josephcsible:
Or maybe not; see #861.

PR #856, opened 2020-02-04
Use NonEmpty to remove partial functions

josephcsible:


josephcsible:
It looks like this will be the first use of NonEmpty in hlint, so I'm not sure what your preferred style for it is. (By the way, it was introduced in GHC 8.0, and it looks like we don't support anything older than that anymore anyway.)

src/Config/Haskell.hs
```diff
@@ -55,7 +56,7 @@ readSetting s (FunBind _ [Match _ (Ident _ (getSeverity -> Just severity)) pats
         let (a,b) = readSide $ childrenBi bind in
         let unit = GHC.noLoc $ GHC.ExplicitTuple GHC.noExt [] GHC.Boxed in
         [SettingMatchExp $
-         HintRule severity (head $ snoc names defaultHintName) s (fromParen lhs) (fromParen rhs) a b
+         HintRule severity (NEE.head $ names NEE.|: defaultHintName) s (fromParen lhs) (fromParen rhs) a b
```
ndmitchell: Would adding `headDef` not be a clearer semantics here? We can define a local `headDef` in a Util module.

src/GHC/Util/Scope.hs
```diff
@@ -68,7 +69,7 @@ scopeMatch' (a, x) (b, y)
 -- ambiguous, pick a plausible candidate.
 scopeMove' :: (Scope', Located RdrName) -> Scope' -> Located RdrName
 scopeMove' (a, x@(fromQual' -> Just name)) (Scope' b) = case imps of
-  [] -> head $ real ++ [x]
+  [] -> NEE.head $ real NEE.|: x
```
ndmitchell: Again, headDef seems the right semantics.

src/Hint/Pragma.hs
```diff
@@ -52,7 +53,7 @@ optToPragma :: [(Located AnnotationComment, [String])]
              -> [(Located AnnotationComment, [String])]
              -> [Idea]
 optToPragma flags langExts =
-  [pragmaIdea (OptionsToComment (map fst old) ys rs) | old /= []]
+  [pragmaIdea (OptionsToComment (fst <$> o NE.:| os) ys rs) | o:os <- [old]]
```
ndmitchell:
```suggestion
  [pragmaIdea (OptionsToComment (fst <$> old2) ys rs) | old2 <- [NE.nonEmpty old]]
```

Is that a neater formulation?

src/Config/Haskell.hs
josephcsible: How about putting that in `Data.List.Extra`?

src/Hint/Pragma.hs
josephcsible: Wouldn't you'd need `| Just old2 <-` instead of `| old2 <-` for that to be valid? Given that, do you still consider it neater? If so, I'm happy to change to it.

ndmitchell: Yep, you would need to do that. I would consider it neater (no chance of accidentally taking apart the list and putting it back in the wrong order)

src/Config/Haskell.hs
ndmitchell: Hmm... Maybe. But where do you stop? Maybe head/tail is exactly where you stop, because that's the one step constructor selectors. I'm not sure.

josephcsible: I opened ndmitchell/extra#50. I'll revisit these once it closes.

ndmitchell:
Thanks!

PR #860, opened 2020-02-05
Fix sortBy hints

josephcsible:
1. The `flip` was in the wrong place
2. `sortOn` isn't actually better than `sortBy` when the function is as cheap as `fst` or `snd`. Looking at [the definition](https://hackage.haskell.org/package/base-4.12.0.0/docs/src/Data.OldList.html#sortOn) will reveal why.


ndmitchell:
I was aware that fst/snd weren't faster, but they did seem cleaner. That said, special casing those two doesn't seem unreasonable at all.

ndmitchell:
Oh, and thanks for catching the bug!

PR #865, opened 2020-02-06
Add a hint to use biList

josephcsible:


googleson78:
If you have `[fst x, snd x]` somewhere in your code, wouldn't you prefer to rewrite it as a match on `x` and `[x1, x2]`? To me it seems like the mental strain of having to figure out which `Bifoldable` is used here is not worth it, especially considering how simple the original snippet is.

My main concern is this:
In my experience most people writing Haskell would prefer to match on `x` instead, and as a result not encounter this hint. The ones who tend to write things like this a lot instead of matching are beginners/students, who have no idea what a `Foldable` is, let alone a `Bifoldable`

ndmitchell:
I have to say I find bilist somewhat odd as a function - it's got inconsistent capitalisation to the other functions in this module, and does a slightly weird thing. Perhaps we could put this in generalise for conciseness?

josephcsible:
Fair enough; done.

ndmitchell:
Thanks!

Issue #587, opened 2019-01-26
I cannot get GHC and HLint to agree..


LeventErkok:
Versions:

```
$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.6.3
$ hlint --version
HLint v2.1.13, (C) Neil Mitchell 2006-2019
```

For this module:

```haskell
{-# LANGUAGE ImplicitParams       #-}
{-# LANGUAGE TypeSynonymInstances #-}
{-# LANGUAGE FlexibleInstances    #-}

module Test where

data X a = X
type XI = X Int

class Foo a where
  foo :: (?p :: Int) => a -> Int

instance Foo XI where foo = undefined
```

I don't seem to be able to make both GHC and HLint happy at the same time:

```
$ hlint Test.hs
Test.hs:2:1: Warning: Unused LANGUAGE pragma
Found:
  {-# LANGUAGE TypeSynonymInstances #-}
Perhaps you should remove it.
Note: Extension TypeSynonymInstances is implied by FlexibleInstances

Test.hs:3:1: Warning: Unused LANGUAGE pragma
Found:
  {-# LANGUAGE FlexibleInstances #-}
Perhaps you should remove it.
Note: Extension FlexibleInstances is implied by ImplicitParams

2 hints
```

Of the 3 `LANGUAGE` extensions in the module, I tried all 8 possible combinations. (Having none to having all of them) and there isn't a single instance where both GHC accept the program and HLint not complain about it.

In fact, it seems that GHC requires all three pragmas to be present to compile it; and HLint isn't happy about that.

Is there some other trick I'm missing to make both happy? Or is this an hlint oversight?

ndmitchell:
Thanks very much for the wonderful test case, really appreciated! The fact that GHC and HLint disagree is always an HLint bug.

Looking at the code, removing `TypeSynonymInstances` still compiles, so that seems a legitimate warning. However, `FlexibleInstances` is required, and I have no idea why GHC would make `ImplicitParams` imply that. @yairchu - how did you get the original list of extensions that implied others? It seems like ImplicitParams does not imply the extensions that are claimed. I can remove that, but might there be any other bugs in it?

Given the bug, I'll make a release tomorrow.

yairchu:
I got it from
https://downloads.haskell.org/~ghc/master/users-guide/glasgow_exts.html#language-options
But translation into code was a manual process so an error may have slipped in

ndmitchell:
[...]

PR #867, opened 2020-02-08
Upgrade to ghc-lib-parser-ex-8.8.5.0

shayne-fletcher:
## 8.8.5.0 released 2020-02-07
- Expose `impliedGFlags` and friends from `DynFlags` (https://github.com/shayne-fletcher/ghc-lib-parser-ex/issues/19, https://github.com/ndmitchell/hlint/issues/587).

### DIFF
--- a/.weeder.yaml
+++ b/.weeder.yaml
@@ -8,4 +8,4 @@
 
 - message:
   - name: Redundant build-depends entry
-  - depends: [ghc-lib, ghc, ghc-boot-th] # Only some subset is supported
+  - depends: [ghc-lib, ghc, ghc-boot-th, ghc-boot] # Only some subset is supported
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -68,11 +68,12 @@ library
         refact >= 0.3,
         aeson >= 1.1.2.0,
         filepattern >= 0.1.1,
-        ghc-lib-parser-ex == 8.8.4.*
+        ghc-lib-parser-ex == 8.8.5.*
     if !flag(ghc-lib) && impl(ghc >= 8.8.0) && impl(ghc < 8.9.0)
         build-depends:
           ghc == 8.8.*,
-          ghc-boot-th
+          ghc-boot-th,
+          ghc-boot
     else
         build-depends:
           ghc-lib-parser == 8.8.*
--- a/stack.yaml
+++ b/stack.yaml
@@ -5,7 +5,7 @@ resolver: nightly-2019-08-07 # Don't roll to an 8.8.1 or 8.8.2 resolver because 
 packages: [.]
 extra-deps:
   - ghc-lib-parser-8.8.2.20200205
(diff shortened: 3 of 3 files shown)


shayne-fletcher:
## 8.8.5.1 released 2020-02-09
- Expression predicate tests.

ndmitchell:
Why did the addition of a few exports necessitate a 0.0.1 bump in ghc-lib-parser-ex?

ndmitchell:
Thanks for the patch, I'll make a release now.

shayne-fletcher:
> Why did the addition of a few exports necessitate a 0.0.1 bump in ghc-lib-parser-ex?

The bump from 8.8.4.0 to 8.8.5.0 came about from the addition of the `impliedXFlags` additions because it technically changed the library interface. I too thought about if it were overkill but realize if I put it out as 8.8.4.1,  hlint were to then depend on it and later, for, some reason later needed to roll back to 8.8.4.0, hlint would be broken so, indeed there's the value.The bump to 8.8.5.1 reflects the addition of lots of tests for the recently added expression predicates. These don't affect the interface (and of course, neither the hlint cabal file)

shayne-fletcher:
> Thanks for the patch, I'll make a release now.

Welcome.

ndmitchell:
I suggest that for adding things to the interface you only bump the 0.0.0.1 value in future, as that saves us as a Stackage break. If we ever needed to roll back HLint could demand 8.8.4.0, but I see no reason to roll back as the interface hasn't lost anything, and should be compatible.

In fact, I'd go with the slightly more pragmatic measure. If ghc-lib-parser-ex doesn't work with the existing HLint add 0.0.1, but if it does, add 0.0.0.1. That's somewhat over-indexing on HLint, but given that's the only user for now, that's not unreasonable.

In future I also suggest we upgrade the libraries in lockstep. e.g. raise a ticket with the upgrade patch, then wait for me to accept it before uploading to Hackage (I'll trust you), and then I'll release immediately after. That stops the Stackage maintainers having to take manual action.

shayne-fletcher:
> In fact, I'd go with the slightly more pragmatic measure. If ghc-lib-parser-ex doesn't work with the existing HLint add 0.0.1, but if it does, add 0.0.0.1. That's somewhat over-indexing on HLint, but given that's the only user for now, that's not unreasonable.

Ok so roughly, if the change wouldn't break hlint as is (just extends, does not change or narrow the interface), just bump the 0.0.0.1. That's pragmatism I can live with.

> In future I also suggest we upgrade the libraries in lockstep. e.g. raise a ticket with the upgrade patch, then wait for me to accept it before uploading to Hackage (I'll trust you), and then I'll release immediately after. That stops the Stackage maintainers having to take manual action.

Anything to avoid that so yes, we can coordinate on 0.0.1 bumps. We can/should expect though that given the above, these become quite rare  now though and we should mostly only see 0.0.0.1 bumps.

PR #869, opened 2020-02-10
Improve hints for Maybe and Bool

josephcsible:
Add hints for `fromMaybe` on a `Bool`, and remove worse ones from generalise-for-conciseness. Also, put the `Just`s on the LHS of the `==`, so it expands to be prettier when the other side is a `do` block.


ndmitchell:
Thanks!

PR #873, opened 2020-02-15
Add warnings for the applicative homomorphism law

josephcsible:


josephcsible:
Okay, updated. (I originally did it that way to match the "Monad law, left identity" hint.)

ndmitchell:
To be honest I'm not convinced Monad law left identity is a particularly helpful name anymore...

ndmitchell:
And Thanks!

PR #875, opened 2020-02-15
Switch 'extensionImplications' to use 'impliedXFlags'

shayne-fletcher:
This is a follow up to https://github.com/ndmitchell/hlint/issues/587. `extensionImplications` is switched to use `impliedXFlags` (introduced by https://github.com/ndmitchell/hlint/pull/867). It's already making things better pointing out the hitherto undetected `DeriveTraversable` implies `DeriveFunctor` and `DeriveFoldable` (hence the update to tests here).


src/HSE/Util.hs
```diff
@@ -393,28 +394,18 @@ extensionImpliedBy :: Extension -> [Extension]
...
-    , (PolyKinds                , [KindSignatures])
-    , (RankNTypes               , [ExplicitForAll])
-    , (ScopedTypeVariables      , [ExplicitForAll])
-    , (TypeOperators            , [ExplicitNamespaces])
-    , (TypeFamilies             , [ExplicitNamespaces, KindSignatures, MonoLocalBinds])
-    , (TypeFamilyDependencies   , [ExplicitNamespaces, KindSignatures, MonoLocalBinds, TypeFamilies])
-    ]
+extensionImplications = map (toHse . g) (foldl' f [] impliedXFlags)
+  where
+    f m (e, _, _) = if e `notElem` m then e : m else m
```
ndmitchell: Does this end up being O(n^2)? Would it be better to just do `e : m` and then a `nubOrd` at the end?

src/HSE/Util.hs
```diff
@@ -15,7 +15,8 @@ import System.FilePath
 import HSE.Type
 import Data.Functor
 import Prelude
-
+import qualified GHC.LanguageExtensions.Type()
```
ndmitchell: Why import this qualified and with nothing in scope. Drop the qualified, and either drop the import entirely, or comment which orphan instance requires it.

tests/hint.test
```diff
@@ -1,9 +1,7 @@
 ---------------------------------------------------------------------
 RUN tests/newtype-derive.hs --with= --hint=data/hlint.yaml
 FILE tests/newtype-derive.hs
-{-# LANGUAGE DeriveFoldable #-}
-{-# LANGUAGE DeriveFunctor #-}
-{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE DeriveTraversable #-} -- Implies DeriveFoldable and DeriveFunctor
```
ndmitchell: I don't get it. In this test you remove `DeriveFunctor`, but at the same time add in a hint that DeriveFunctor is not required. I don't get why you have both patches in one code, or how that still passes the test. I suspect I'm missing something.

src/HSE/Util.hs
shayne-fletcher: Dropping the import. Turns out not required.

tests/hint.test
shayne-fletcher: I made two changes:
  - `newtype-deriv.hs`
    - Don't import `DeriveFoldable` or `DeriveFunctor`
    -  If we import them, we'll get new output hinting that they are not required
  - `restricted-extension`
    - Do import `DeriveFoldable` and `DeriveFunctor` (no change)
    - Add the extra warnings now generated to the expected output
    - Intent is to allow this as an incidental test of this PR.

src/HSE/Util.hs
shayne-fletcher: Can't `nubOrd` (no `Ord`). I realize now that I've missed a spot though where we should `nub`.

tests/hint.test
ndmitchell: OK, that makes sense - very good idea

src/HSE/Util.hs
ndmitchell: That concerns me a lot - these extensions could get numerous. An O(n^2) here could end up being noticable. Any chance we send an upstream patch to GHC to derive Ord on these things?

shayne-fletcher: I don't see why that should be hard. I'll take it.

shayne-fletcher:
In an upcoming `ghc-lib-parser-ex`, I'll move most of the implementation there. When an `Ord` instance is available for `Extension`, I'll be sure to take advantage of it. In the meantime, have to add an exclusion for "Avoid restricted function" to `HSE.Util`.

src/HSE/Util.hs
shayne-fletcher: MR : https://gitlab.haskell.org/ghc/ghc/merge_requests/2707.

ndmitchell:
I assume we'll pull this in favour of the code going into ghc-lib-parser-exts?

shayne-fletcher:
Yes. Let's leave this open. We can wait for the just released 8.8.5.2 to arrive in the Stackage nightly, then I'll update this PR. Sound good?

shayne-fletcher:
I propose this is in it's final form now and ready to merge whenever you're comfortable.

shayne-fletcher:
Satisfyingly, in this PR we finally make use of the `Bounded` instance we upstreamed for `Extension` according to the MR, 9 months ago 🎉

shayne-fletcher:
Test failure for 8.4.4 because it didn't get the 8.8.5.2 ghc-lib-parser-ex. Forcing a re-run.

ndmitchell:
You need to update the hlint.cabal to demand `ghc-lib-parser-ex >= 8.8.5.2 && ghc-lib-parser-ex < 8.8.6`. That's why the 8.4 build is failing.

shayne-fletcher:
Ok. In that case we then have to ensure that the release reach the Stackage nightly first. See, I put it out  as 8.8.5.2 so as to avoid touching the cabal file as per our discussions last week. If modifying the cabal file is proving necessary (as it is). better we fall back to using major version bumps going forward when we formally change the interface (by the addition of new functions) no?

shayne-fletcher:
The good news is its made it there (https://www.stackage.org/package/ghc-lib-parser-ex). I'll make the update to the cabal file.

ndmitchell:
No requirement for it to reach Stackage nightly. Touching the .cabal file is absolutely fine and has no impact on Stackage. I don't think touching the Cabal file has any impact at all, so our current plan still works nicely.

shayne-fletcher:
I'm confused. Let's resolve by an email chat later today when I've had some coffee 😄 . In the meantime I'll get this out the door.

hlint.cabal
```diff
@@ -68,7 +68,7 @@ library
         refact >= 0.3,
         aeson >= 1.1.2.0,
         filepattern >= 0.1.1,
-        ghc-lib-parser-ex == 8.8.5.*
+        ghc-lib-parser-ex <= 8.8.5.2 && < 8.8.6
```
ndmitchell:
```suggestion
        ghc-lib-parser-ex >= 8.8.5.2 && < 8.8.6
```

shayne-fletcher: 😊

shayne-fletcher:
> No requirement for it to reach Stackage nightly. Touching the .cabal file is absolutely fine and has no impact on Stackage. I don't think touching the Cabal file has any impact at all, so our current plan still works nicely.

I think a light bulb is beginning to drop. In any case, will clarify by mail later to be sure I've got it.

shayne-fletcher:
The last 8.4 failed with what I think is a transient network error `The command "curl -sSL https://raw.github.com/ndmitchell/neil/master/travis.sh | sh" exited with 1`. Will force a rebuild.

shayne-fletcher:
Argh no, not transient.
```
arning: solver failed to find a solution:
Could not resolve dependencies:
[__0] trying: hlint-2.2.11 (user goal)
[__1] trying: hlint:+gpl
[__2] unknown package: hscolour (dependency of hlint +gpl)
[__2] fail (backjumping, conflict set: hlint, hscolour, hlint:gpl)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: hlint, hlint:gpl, hscolour
Trying configure anyway.
Configuring hlint-2.2.11...
cabal: Encountered missing dependencies:
aeson >=1.1.2.0,
cpphs >=1.20.1,
data-default >=0.3,
file-embed -any,
filepattern >=0.1.1,
ghc-lib-parser ==8.8.*,
ghc-lib-parser-ex >=8.8.5.2 && <8.8.6,
haskell-src-exts >=1.21 && <1.24,
haskell-src-exts-util >=0.2.5,
uniplate >=1.5,
utf8-string -any,
vector -any,
yaml >=0.5.0
```
No idea what's going wrong here 😢

ndmitchell:
I think the problem is that ghc-lib-parser now depends on containers-0.6.0.1. I suggest you reduce that to contains-0.5 or higher, and then I think it might work. You can reduce it without a new release by editing the Cabal metadata on Hackage directly. Once you've done that, I'll trigger a rebuild.

shayne-fletcher:
Edited metadata so that we have `containers >= 0.5.8.1`. Let's see.

PR #877, opened 2020-02-16
Use unLoc instead of reimplementing it

josephcsible:


ndmitchell:
Thanks! Cc @zilu41. Also feel free to comment on other people's PR's at any point - the more reviewers the better.

PR #878, opened 2020-02-16
Remove bad arrow hints

josephcsible:
These hints will break any program that uses them with actual arrows that aren't just functions.


josephcsible:
If you'd rather keep these hints, I have two alternative ideas:

1. Change `Data.Tuple.Extra` to re-export these combinators as-is, instead of redeclaring them with a less general type
2. Add something like `isFunction` to hlint, and use it on these hints

ndmitchell:
I think the right thing to do is probably add a note that these only work on function arrows (or tuple arrows in some places - it's pretty confusing). The observation was that nearly all users of these functions are using them without actually using arrows, and since it's opt in, having wrong by default hints is less problematic.

josephcsible:
Wait a minute. As I think about this more, isn't the net effect of these hints just to replace `import Control.Arrow` with `import Data.Tuple`, without changing the actual code at all? What's the gain by doing that, even in the common case where it is valid?

ndmitchell:
User comprehension - Control.Arrow can do a lot of crazy things, and is nearly always used at one common type. Figuring out that type from the docs of Control.Arrow is a logic puzzle. In contrast, for Data.Tuple.Extra, it's pretty obvious.

ndmitchell:
While I think doing the replacement is unambiguously a good idea, I agree it breaks code, and the note that explains why it breaks code is going to be awfully odd. Removed for now.

PR #879, opened 2020-02-17
Remove more incorrect hints

josephcsible:
`length (zipFrom LT "foobar")` bottoms, whereas `length (zip [LT ..] "foobar")` returns 3.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -885,8 +885,6 @@
     - warn: {lhs: "null (intersect a b)", rhs: "disjoint a b"}
     - warn: {lhs: "[minBound .. maxBound]", rhs: "enumerate"}
     - warn: {lhs: "zipWithFrom (,)", rhs: "zipFrom"}
-    - warn: {lhs: "zip [i..]", rhs: "zipFrom i"}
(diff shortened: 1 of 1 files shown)


ndmitchell:
I put an alternative patch up in e20575341baa9cc9753c3330d58c7adc9be912ca - generally, if the values are Int's, you are fine, and that's where the hints almost always apply anyway, so now they're pretty-much safe (assuming your Int type doesn't overflow or other weirdities which no one expects anyway).

josephcsible:
That's definitely better, but the problem can still occur. Consider:

```haskell
foo :: [(Int8, Char)]
foo = zipWith [125..] "foobar"
```

I wish HLint could see types, so that we could just restrict this to `Integer` and `Natural` and call it a day.

ndmitchell:
Another option would be to adjust zipFrom so that it really does match zip [LT..]

josephcsible:
That may actually be the best way to solve this. Just `zipFrom = zip . enumFrom`.

ndmitchell:
I did exactly that in extra-1.7, so closing this patch and removing the isLitInt.

PR #884, opened 2020-02-21
Add more fish operator hints

zliu41:


data/hlint.yaml
```diff
@@ -443,6 +443,8 @@
     - warn: {lhs: flip (=<<), rhs: (>>=)}
     - hint: {lhs: \x -> f x >>= g, rhs: f Control.Monad.>=> g}
     - hint: {lhs: \x -> f =<< g x, rhs: f Control.Monad.<=< g}
+    - hint: {lhs: (>>= f) . g, rhs: f Control.Monad.<=< g}
```
josephcsible:
```suggestion
    - hint: {lhs: (>>= g) . f, rhs: f Control.Monad.>=> g}
```
Keep the symmetry?

zliu41: I like my version slightly better since it preserves the relative order of `f` and `g`. Imagine the original code being `f1 . f2 . (>>= f3) . f4 . f5`

ndmitchell:
Thanks. Happy either way, but keeping the relative order seems better if we can.

PR #886, opened 2020-02-24
Update to ghc-lib-parser-8.8.3.20200224.

shayne-fletcher:
As before, non-critical, just helping me keep things straight. Change `stack.yaml` to reference the just released `ghc-lib-parser-8.8.3.20200224`.


ndmitchell:
Sure thing!

PR #888, opened 2020-02-26
Use foldr to call functions directly, instead of unnecessarily composing them first

josephcsible:


ndmitchell:
Cool and mind bending :)

PR #889, opened 2020-02-27
Convert Lambda.hs to ghc-lib-parser

googleson78:
Things to look at:

* [ ] There is still an `Input/outputs` test that fails, because `niceLambdaR'` doesn't output the right refactor.
* [ ] I haven't tried running the `refact` tests from my branch, so those might fail too (?)
* [x] I used `PatternSynonyms` somewhere so I had to whitelist it. Should I remove the usage?
  \- No - it's fine. See discussions.
* [x] In `lambdaDecl`: The `PatBind` case that was handled before by converting to a `FunBind` is now unhandled.
  This is because:
  * `FunBind`s have a name, whereas `PatBind`s don't - so we can't generate a `FunBind` from a `PatBind` like previously
  * I don't think that hint will ever trigger for a `PatBind` (from reading the ghc docs on `PatBind` and playing a tiny bit with what asts ghc outputs)

  \- It was decided to leave the `PatBind` case "unhandled".

(thanks for the patience and sorry for the delay!)


src/GHC/Util/View.hs
```diff
@@ -53,3 +55,7 @@ instance View' (Pat GhcPs) PApp_' where
   view' (fromPParen' -> LL _ (ConPatIn (L _ x) (InfixCon lhs rhs))) =
     PApp_' (occNameString . rdrNameOcc $ x) [lhs, rhs]
   view' _ = NoPApp_'
+
+-- A lambda with no guards and no where clauses
+pattern SimpleLambda :: [Pat GhcPs] -> LHsExpr GhcPs -> LHsExpr GhcPs
```
googleson78: I added a pattern synonym instead of following the `View` convention, because it's less boilerplate.
The `View`s here have no recursive calls of `view` so they seem to be equivalent to pattern synonyms.

src/Hint/All.hs
```diff
@@ -44,10 +44,8 @@ data HintBuiltin =
 
 builtin :: HintBuiltin -> Hint
 builtin x = case x of
-    -- Hse.
```
googleson78: :tada:

src/Hint/Lambda.hs
```diff
@@ -86,96 +93,179 @@ yes = map (\f -> dataDir </> f) dataFiles -- (dataDir </>)
...
-    , not $ any isQuasiQuote $ universe x
-    = etaReduce (init ps) x
-etaReduce ps (InfixApp a x (isDol -> True) y) = etaReduce ps (App a x y)
-etaReduce ps x = (ps,x)
+lambdaHint :: DeclHint'
+lambdaHint _ _ x
+    =  concatMap (uncurry lambdaExp') (universeParentBi x)
+    ++ concatMap lambdaDecl' (universe x)
+
+-- TODO: handle PatBinds
```
googleson78: I think this TODO can safely be deleted, as per the comments in the pull request

src/Hint/Lambda.hs
```diff
@@ -86,96 +93,179 @@ yes = map (\f -> dataDir </> f) dataFiles -- (dataDir </>)
...
+          t1 = template finalpats
+          --t2 = template pats2 bod2
+lambdaDecl' _ = []
+
+
+etaReduce' :: [Pat GhcPs] -> LHsExpr GhcPs -> ([Pat GhcPs], LHsExpr GhcPs)
+etaReduce' (unsnoc -> Just (ps, view' -> PVar_' p)) (LL _ (HsApp _ x (view' -> Var_' y)))
+    | p == y
+    , p /= "mr"
+    -- TODO: what is this "mr" thing??
```
googleson78: Why do we not eta-reduce variables named `mr`? :thinking:

src/GHC/Util/View.hs
googleson78: Ah, `PatternSynonyms` is restricted
It's been available since `7.8.1` so it shouldn't be an issue. Is there some other reason to restrict it?

.hlint.yaml
```diff
@@ -10,7 +10,7 @@
 - extensions:
   - default: false
   - name: [DeriveDataTypeable, GeneralizedNewtypeDeriving, NoMonomorphismRestriction, OverloadedStrings]
-  - name: [MultiWayIf, PatternGuards, RecordWildCards, ViewPatterns, TupleSections, LambdaCase]
+  - name: [MultiWayIf, PatternGuards, RecordWildCards, ViewPatterns, PatternSynonyms, TupleSections, LambdaCase]
```
googleson78: I've whitelisted `PatternSynonyms` so I can get CI further along.
If there is some issue with the extension I'll remove this commit and rewrite the code using them to use `View'` instead.

googleson78:
uh-oh.. I guess there is a bit of work to do w.r.t. refactorings

.hlint.yaml
ndmitchell: Pattern synonyms are absolutely fine.

src/GHC/Util/View.hs
ndmitchell: No reason to restrict it. Pattern synonym away!

src/Hint/All.hs
ndmitchell: Woohoo indeed!

src/Hint/Lambda.hs
ndmitchell: Then please delete :)

src/Hint/Lambda.hs
ndmitchell: Because `mr` stands for monomorphism restriction and some people type `foo mr = bar mr` to indicate that this variable can't be eta reduced or you trip into that restriction. From the days before easy ignoring, and I suspect no one knows about it - feel free to delete.

googleson78: Deleted it.

src/Hint/Lambda.hs
googleson78: Done

ndmitchell:
There's still a Haddock error:

```
src/Hint/Lambda.hs:222:18: error:
    parse error on input ‘-- * add brackets to the match, because matches in lambdas require them’
    |
222 |                  -- * add brackets to the match, because matches in lambdas require them
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

googleson78:
I tried to start fixing the refactorings, but I ran into cyclic dependencies (when trying to import and use `toSS'`), so it will take a bit more work to figure that out (and some of this works needs to be done to get rid of the `HSE` dependency anyway). So I tagged them with `@NoRefactor`.
This only leaves 
```
TEST FAILURE IN tests/_serialise_7
...
tests/serialise-six.hs
...
``` 
failing, but I don't know how to ignore that one. (or if I should)

googleson78:
I fixed the failing `Input/output` test in a [different branch](https://github.com/googleson78/hlint/tree/lambda-refactor)  (+ some of the refactor tests that I ignored here), but it requires some other tangential changes, so I haven't pushed it here.

src/Hint/Lambda.hs
```diff
@@ -22,28 +22,30 @@
 <TEST>
 f a = \x -> x + x -- f a x = x + x
 f a = \a -> a + a -- f _ a = a + a
+f (Just a) = \a -> a + a -- f (Just _) a = a + a @NoRefactor
```
googleson78: N.B. (to myself mostly): This test fails refactoring even with the old code

src/Hint/Lambda.hs
```diff
@@ -69,14 +71,18 @@ foo a b c = bar (flux ++ quux) c where flux = a -- foo a b = bar (flux ++ quux)
...
+yes = blah (\ x -> case x of A -> a; B -> b) -- @Note may require `{-# LANGUAGE LambdaCase #-}` adding to the top of the file @NoRefactor
 no = blah (\ x -> case x of A -> a x; B -> b x)
+yes = blah (\ x -> (y, x)) -- (y,) @NoRefactor
 yes = blah (\ x -> (y, x, z+q)) -- (y, , z+q) @NoRefactor
+yes = blah (\ x -> (y, x, y, u, v)) -- (y, , y, u, v) @NoRefactor
+yes = blah (\ x -> (y, x, z+q)) -- @Note may require `{-# LANGUAGE TupleSections #-}` adding to the top of the file @NoRefactor
 yes = blah (\ x -> (y, x, z+x))
 tmp = map (\ x -> runST $ action x)
-yes = map (\f -> dataDir </> f) dataFiles -- (dataDir </>)
+yes = map (\f -> dataDir </> f) dataFiles -- (dataDir </>) @NoRefactor
```
googleson78: There is something weird going on here.

Even though I'm using `suggestN'` which adds an empty list of refactorings, this test still fails with "bad refactoring", because there is **some** refactoring suggested:
```
TEST FAILURE (BAD REFACTORING)
SRC: src/Hint/Lambda.hs:85:0
INPUT: yes = map (\f -> dataDir </> f) dataFiles
Refactor output is expected to contain: (dataDir </>)
Actual: yes = map (\f -> dataDir </> f) dataFiles
```

What's weirder to me is how even before my patch `suggestN` was still being used, yet this test didn't require `@NoRefactor`. There must be some magic that I'm missing..

Perhaps @zliu41 can clue me in?

zliu41: The test pass criteria is that refactoring output contains the suggested replacement (`ideaTo`). If the list of refactorings is empty, then refactoring does nothing and echoes the input, `yes = map (\f -> dataDir </> f) dataFiles`. It doesn't contain the suggested replacement, `(dataDir </>)`, so the test fails.

Previously the list of refactorings contains `(refact $ toSS o)` (https://github.com/ndmitchell/hlint/blob/master/src/Hint/Lambda.hs#L154), so it worked.

ndmitchell:
Given we're down to one failing test, I'm just going to merge and fix up the details after. So close and not willing to be defeated!

PR #893, opened 2020-03-02
Derive MonadFail instead of using error

josephcsible:


ndmitchell:
Very happy to take this as long as it doesn't require any additional CPP due to version differences... Let's see what CI says.

ndmitchell:
`liftIO $ errorIO` would be another possibility,

josephcsible:
If what I just tried doesn't work, then I'll switch to that.

ndmitchell:
Missing import of MonadFail it looks like. I am increasingly feeling `liftIO $ errorIO` is the correct thing to do, rather than fail. We definitely want an error, since it's TODO code, whereas for Test it would be perfectly reasonable to make fail be a test failure rather than a fatal exception.

ndmitchell:
I switched us to errorIO in this case and a lot of others in ce6883a269518fd7ae656512def58714c20b35e4, which I think answers the intent behind this PR, so closing.

PR #894, opened 2020-03-03
Use GHC SrcSpans in Idea

googleson78:
Not sure if this is what `Idea needs to not use HSE SrcSpans` means.

I used adapters in some places (e.g. `Apply.hs`), because from what I can tell you need to get rid of `ParseFlags` and hse parsing there first, to be able to get rid of the HSE SrcSpans there.


src/GHC/Util.hs
```diff
@@ -15,8 +18,14 @@ module GHC.Util (
 
   , parsePragmasIntoDynFlags
   , parseFileGhcLib, parseExpGhcLib, parseImportGhcLib
+  , hseSpanToGHC, hseLocToGHC
```
googleson78: Is this module an ok location for all these things?

src/GHC/Util.hs
```diff
@@ -51,3 +63,82 @@ parseFileGhcLib :: FilePath -> String -> DynFlags -> ParseResult (Located (HsMod
...
+
+toOldeSpan :: SrcSpan -> (String, Int, Int, Int, Int)
+toOldeSpan (RealSrcSpan span) =
+  ( unpackFS $ srcSpanFile span
+  , srcSpanStartLine span
+  , srcSpanStartCol span
+  , srcSpanEndLine span
+  , srcSpanEndCol span
+  )
+-- TODO: the bad locations are all (-1) right now
```
googleson78: This is fine, right?

src/Test/Annotations.hs
```diff
@@ -69,7 +72,8 @@ testAnnotations setting file rpath = do
                         ,"SRC: " ++ showSrcLoc loc
                         ,"INPUT: " ++ inp
                         ,"OUTPUT: " ++ show i]
-                        | i@Idea{..} <- fromRight [] ideas, let SrcLoc{..} = getPointLoc ideaSpan, srcFilename == "" || srcLine == 0 || srcColumn == 0]
+                        | i@Idea{..} <- fromRight [] ideas, let GHC.SrcLoc{..} = GHC.srcSpanStart ideaSpan, srcFilename == "" || srcLine == 0 || srcColumn == 0]
+                        -- TODO: shouldn't these checks be == -1 instead?
```
googleson78: I'm probably entirely missing something here, but the `== 0`s caught my attention!

src/GHC/Util.hs
shayne-fletcher: I'm a bit reluctant to have the GHC modules depend on HSE if we can get away with it. We've managed to avoid that until now. Note also there are the functions `ghcSpanToHSE`, `ghcSrcLocToHSE` in `HSE/All.hs`.

googleson78: I moved it to `HSE.All`

I'm restating the same question about the other functions.

ndmitchell: I'm not overly fussed about where the functions are for now. I imagine all this stuff will move over time.

shayne-fletcher: Thanks for that. Yes, not a big deal but nice that it worked out putting them in `All`. These conversion functions will ultimately be eliminated 😄

src/GHC/Util.hs
shayne-fletcher: IIUC then yes, I think so.

shayne-fletcher:
If we didn't use pattern synonyms here, would it have a profound effect on readability/quantity of code?

src/Test/Annotations.hs
shayne-fletcher: Think Neil will have to comment on those. Another possibility is that there is some machinery is set-up to inject these spans with those values for the specific reason of triggering those matches.

googleson78:
@shayne-fletcher 

Not too much, 4 places will be affected (from quick grepping)

It would either be
* have a function that converts "unhelpful" variants to realsrcspans and call it everywhere as preprocessing
* have {start,end}{Line,Col}(+file) that internally handle the unhelpful variant and directly call them

I went with this because it seemed like it would have the fewest changes. Are you worried that it's unneeded complexity (or something else)?

shayne-fletcher:
@googleson78 To be honest, pattern synonyms confuse the heck out of me and I have to go and relearn them every time I encounter one :)  Maybe I just have to get over it - let's hear from Neil - but if we're introducing them in this program for the first time and not doing so doesn't impact us too hard, then maybe let's not?

src/GHC/Util.hs
```diff
@@ -15,8 +17,13 @@ module GHC.Util (
 
   , parsePragmasIntoDynFlags
   , parseFileGhcLib, parseExpGhcLib, parseImportGhcLib
+  , pattern SrcSpan, srcSpanFilename, srcSpanStartLin, srcSpanStartColumn, srcSpanEndLin, srcSpanEndColumn
+  , pattern SrcLoc, srcFilename, srcLine, srcColumn
+  , showSrcLoc'
   ) where
 
+import Data.List.Extra (drop1)
```
ndmitchell: No need to import this with restricted functions.

src/GHC/Util.hs
```diff
@@ -51,3 +59,72 @@ parseFileGhcLib :: FilePath -> String -> DynFlags -> ParseResult (Located (HsMod
 parseFileGhcLib filename str flags =
   GhclibParserEx.parseFile filename flags
     (if takeExtension filename /= ".lhs" then str else unlit filename str)
+
+{-# COMPLETE SrcSpan #-}
+-- | The \"Lin\" thing is because there is already e.g. 'SrcLoc.srcSpanStartLine'
```
ndmitchell: Typically we've been adding '. That convention has served us well, so I'd follow that.

src/Test/Annotations.hs
ndmitchell: I have no idea. It does seem odd!

src/GHC/Util.hs
googleson78: I tend to prefer it by default because it's way easier for people without an ide to navigate code. (from my past experience)

src/GHC/Util.hs
googleson78: I originally didn't follow it because usually if `X` exists for HSE then `X'` is "the same thing" but for ghc. (while in this case `srcSpanStartLine` exists in ghc, not in HSE)

I guess this makes sense - I changed it.

googleson78:
@shayne-fletcher I should note that they are not being introduced here! (so I guess it's flown under your nose unintentionally :( )

I used one in #889 instead of using the `View` "idiom" -https://github.com/ndmitchell/hlint/pull/889/files#r384997274

src/GHC/Util.hs
ndmitchell: Easier solution is to get an IDE: https://github.com/digital-asset/ghcide

ndmitchell:
Great progress. We can always refine it later.

I believe this change breaks the API of HLint, because Idea is an exported data type. Therefore, it's a 0.1 version bump.

ndmitchell:
I suggest we only release it once _all_ HSE types are eliminated from the API in HLint4.

src/GHC/Util.hs
googleson78: Yes absolutely! `ghcide` has been invaluable for working on `hlint`, comparing my experience to the one I had previously without relying on `ghcide`! :heart:

PR #895, opened 2020-03-04
Expand variables in orig if the substitution has no loc

zliu41:
This fixes a couple of refactor tests, for instance

```haskell
yes = foo $ \(a, b) -> (a, y + b)
```

Before:
```
--serialise
[("/home/ziyangliu/ana/Foo.hs:1:13: Suggestion: Use second\nFound:\n  \\ (a, b) ->
(a, y + b)\nPerhaps:\n  Data.Bifunctor.second ((+) y)\nNote: increases laziness\n",
[Replace {rtype = Expr, pos = SrcSpan {startLine = 1, startCol = 13, endLine = 1,
endCol = 34}, subts = [], orig = "Data.Bifunctor.second (f)"}])]

--refactor
yes = foo $ Data.Bifunctor.second (f)
```

The bad refactoring is caused by `subts = []`, which is in turn caused by the substitution having no loc, because the substitution is `(+) y` which does not appear in the original code.

After (`orig` is expanded):

```
--serialise
[("/home/ziyangliu/ana/Foo.hs:1:13: Suggestion: Use second\nFound:\n  \\ (a, b) ->
(a, y + b)\nPerhaps:\n  Data.Bifunctor.second ((+) y)\nNote: increases laziness\n",
[Replace {rtype = Expr, pos = SrcSpan {startLine = 1, startCol = 13, endLine = 1,
endCol = 34}, subts = [], orig = "Data.Bifunctor.second ((+) y)"}])]

--refactor
yes = foo $ Data.Bifunctor.second ((+) y)
```

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -986,7 +986,7 @@
 # yes = if foo then do stuff; moreStuff; lastOfTheStuff else return () \
 #     -- Control.Monad.when foo $ do stuff ; moreStuff ; lastOfTheStuff @NoRefactor: hlint bug
 # yes = if foo then stuff else return () -- Control.Monad.when foo stuff
-# yes = foo $ \(a, b) -> (a, y + b) -- Data.Bifunctor.second ((+) y) @NoRefactor: hlint bug, subts = []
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks!

PR #897, opened 2020-03-06
Clarify missing refactor error message

andys8:
I didn't know about the external tool `refactor`. This is the error message that currently occurs. I think it might be helpful to clarify `refactor` is an executable and not a command line argument, a hint that can be refactored or other internal stuff.

```shell
hlint src/Log.hs
src/Log.hs:20:35: Warning: Redundant bracket
Found:
  (minLogLevel)
Perhaps:
  minLogLevel

1 hint

hlint src/Log.hs --refactor
hlint: Could not find refactor
Tried with: refactor

CallStack (from HasCallStack):
  error, called at src/HLint.hs:251:20 in hlint-2.2.2-JJ80rphq264GBRgLnOQWMf:HLint
```

### DIFF
--- a/src/Refact.hs
+++ b/src/Refact.hs
@@ -50,7 +50,12 @@ refactorPath rpath = do
             pure $ if versionBranch ver >= [0,1,0,0]
                        then Right exc
                        else Left "Your version of refactor is too old, please upgrade to the latest version"
-        Nothing -> pure $ Left $ unlines [ "Could not find refactor", "Tried with: " ++ excPath ]
+        Nothing -> pure $ Left $ unlines
+                       [ "Could not find 'refactor' executable"
+                       , "Tried to find '" ++ excPath ++ "' on the PATH"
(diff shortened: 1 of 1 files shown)


ndmitchell:
Good idea! Since we are improving the message, perhaps we should say how to cabal install it in the message too?

andys8:
Okay, what do you think of..?
```haskell
[ "Could not find 'refactor' executable"
, "Tried to find '" ++ excPath ++ "' on the PATH"
, "'refactor' is provided by the 'apply-refact' package and has to be installed"
, "<https://github.com/mpickering/apply-refact>"
]
```

I personally am using `stack` instead of `cabal` and this install instruction could confuse me. So I would prefer a link. But feel free to change the message.

Btw. I didn't want to mess up the projects formatting. Is there a specific tool in use?

ndmitchell:
Thanks for the patch, looks good - probably best to avoid cabal vs stack as you have done.

I don't use any specific formatting tool yet, although I might do one day!

PR #899, opened 2020-03-11
Warn unused PatternSynonyms

zliu41:


ndmitchell:
Thanks!

ndmitchell:
For future me, I confirmed that when I imported a pattern synonym and used it, that didn't require the pragma.

PR #900, opened 2020-03-12
Add ParseMode', defaultParseMode'; upgrade ghc-lib-parser-ex-8.8.5.4

shayne-fletcher:
Smallest of starts on parse flags:
- Publish and upgrade to ghc-lib-parser-ex-8.8.5.4 exposing `preludeFixities`, `baseFixities`, `lensFixities` and `otherFixities`;
- Add type `ParseMode'` and function `defaultParseMode'` to `All.hs`.

### DIFF
--- a/.weeder.yaml
+++ b/.weeder.yaml
@@ -3,5 +3,9 @@
 ##########################
 
 - message:
+  - name: Weeds exported
+  - identifier: [ParseMode', defaultParseMode', isAssocLeft, isAssocNone] # Last two weeder bugs?
+
+- message:
   - name: Redundant build-depends entry
   - depends: [ghc-lib, ghc, ghc-boot-th, ghc-boot] # Only some subset is supported
--- a/src/HSE/All.hs
+++ b/src/HSE/All.hs
@@ -9,6 +9,8 @@ module HSE.All(
     ParseError(..), ModuleEx(..),
     parseModuleEx, ghcComments,
     parseExpGhcWithMode, parseImportDeclGhcWithMode, parseDeclGhcWithMode,
+-- Not used yet.
+    ParseMode', defaultParseMode'
     ) where
 
 import Util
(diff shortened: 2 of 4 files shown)


ndmitchell:
I'm not sure if lensFixities or otherFixities should go into ghc-lib-parser-ex - they are somewhat adhoc to meet the needs of HLint, while the prelude and base fixities are both very well specified and reusable. Thoughts?

ndmitchell:
And should fixities be identified by String or RdrName?

shayne-fletcher:
> I'm not sure if lensFixities or otherFixities should go into ghc-lib-parser-ex - they are somewhat adhoc to meet the needs of HLint, while the prelude and base fixities are both very well specified and reusable. Thoughts?

That works for me. I'll take them out. The base and prelude fixities were already there, just not exposed. I'll push 8.8.5.5 which removes lens and others and update the PR.

shayne-fletcher:
> And should fixities be identified by String or RdrName?

I would have said `String` but what is it that prompts you to ask?

shayne-fletcher:
Ok, have pushed 8.8.5.5 which removes lens and other fixities keeping base and prelude. Rebased on master and taken a queue from you on the fields you want to keep in `ParseMode`.

shayne-fletcher:
Picked up in https://github.com/ndmitchell/hlint/pull/902.

shayne-fletcher:
> And I'm getting increasingly suspicious of the Language flag - either we're not honouring that in GHC, or we should be, or something doesn't feel right.

I think what's going on with this is that the language flag has influenced what extensions can be enabled. This hasn't mattered to GHC because it has relied on HSE for all information relating to what extensions to enable/disable.

ndmitchell:
Having been through it I think in fact we've just not been passing the language flag - I just pushed a patch to wire it through where it should be. In truth the language flag enables a few additional default extensions, and since we enable a huge number, it doesn't have much impact on us.

shayne-fletcher:
Argh. Your push trampled on my branch again 😄

PR #902, opened 2020-03-12
Min version ghc-lib-parser-ex 8.8.8.5

shayne-fletcher:
Update ghc-lib-parser-ex to get access to prelude and base fixities.


ndmitchell:
Thanks!

PR #905, opened 2020-03-13
Match hints in spite of left composition

### LINKED ISSUE #890: Composed functions to the left don't match hints
Given the `isLeft (fmap f x) ==> isLeft x` hint, it should trigger in all these 4 cases, but it only triggers in the first three.

```
foo f x = isLeft (fmap f x)
bar f = isLeft . fmap f
baz f x = isLeft $ fmap f x
qux magic f x = magic . isLeft $ fmap f x
```

From #882

zliu41:
Fixes #890

### DIFF
--- a/src/GHC/Util/Unify.hs
+++ b/src/GHC/Util/Unify.hs
@@ -1,17 +1,19 @@
-{-# LANGUAGE PatternGuards, ViewPatterns, FlexibleContexts, ScopedTypeVariables #-}
+{-# LANGUAGE PatternGuards, ViewPatterns, FlexibleContexts, ScopedTypeVariables, TupleSections #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 module GHC.Util.Unify(
     Subst', fromSubst',
     validSubst', substitute',
-    unifyExp'
+    unifyExp
     ) where
 
+import Control.Applicative
 import Control.Monad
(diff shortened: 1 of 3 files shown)


src/GHC/Util/Unify.hs
```diff
@@ -109,6 +110,56 @@ unify' nm root x y
...
+  ((, Just y11) <$> unifyExp' nm False x1 y12)
+    <|> case y12 of
+          (L _ (OpApp _ y121 dot' y122)) | isDot dot' ->
+            unifyComposed' nm x1 (noLoc (OpApp noExt y11 dot y121)) dot' y122
+          _ -> Nothing
+
+-- unifyExp handles the cases where both x and y are HsApp, or y is OpApp. Otherwise,
+-- delegate to unifyExp'. These are the cases where we potentially need to call
+-- unifyComposed' to handle left composition.
+unifyExp :: NameMatch' -> Bool -> LHsExpr GhcPs -> LHsExpr GhcPs -> Maybe (Subst' (LHsExpr GhcPs), Maybe (LHsExpr GhcPs))
```
ndmitchell: Can you add a comment as to what the second component of the result is? When it is Just? If it is Just, what should be done with it?

zliu41: Done!

ndmitchell:
Thanks!

PR #906, opened 2020-03-13
Move lens, other, custom fixities to Fixity

shayne-fletcher:
This PR moves the definitions of  lens, other and custom fixities to `Fixity`. They are now specified as `(String, Fixity)`  lists in the same style as the prelude and base fixities from `GhclibParserEx`.

- Surprisingly, the change caused a test failure:
  -  The test ``no = sort <$> f input `shouldBe` sort <$> x`` before this change produced no hints;
  - After this change, it triggers "Functor law";
  - I changed the test to expect that hint because in fact it looks kosher (`f <$> g <$> x === (f . g) <$> x`) although I'd appreciate it if you could have a look at it - the existence of the test suggests I may be missing something;
- I took the opportunity to prune an entry for a now deleted file from `.hlint.yaml`.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -71,5 +71,3 @@
 - ignore: {name: Use const, within: Config.Yaml}
 # temporary disable while we have unused top binds here
 - ignore: {name : Avoid restricted flags, within: [GHC.Util.HsExpr, Hint.Match]}
-# It is used and on Ubuntu 8.8, GHC errors if it's not enabled.
-- ignore: {name : Unused LANGUAGE pragma, within: [GHC.Util.Language.Haskell.GHC.ExactPrint.Types]}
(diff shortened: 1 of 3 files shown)


ndmitchell:
The test shows a real problem. Given how `shouldBe` is infix 1, the brackets are implicitly `(sort <$> f input) `shouldBe` (sort <$> x)` so it doesn't trigger an fmap hint. However, it seems the shouldBe hint got lost somewhere. My best guess is you need to remove the tick marks around all the function names. in the infix declarations.

shayne-fletcher:
> The test shows a real problem. Given how `shouldBe` is infix 1, the brackets are implicitly `(sort <$> f input) `shouldBe` (sort <$> x)` so it doesn't trigger an fmap hint. However, it seems the shouldBe hint got lost somewhere. My best guess is you need to remove the tick marks around all the function names. in the infix declarations.

Indeed, that was it.

ndmitchell:
Thanks!

PR #911, opened 2020-03-15
Only use `<- pure` where it provides value

josephcsible:


ndmitchell:
What you call an anti-pattern, I call a pattern. The prime pattern is vastly more dangerous - the odds of referring to ghcFlags or ghcFlags' when you meant the other one is super high and it's usually not a type error. The odds of forgetting a prime on the LHS and getting into a hard to diagnose infinite loop are also high. What I really want is `let` and `let rec` to be two different keywords, like in OCaml, but in the absence of that, the `<- pure` pattern seems like a good idea.

ndmitchell:
I wrote up my thoughts in https://neilmitchell.blogspot.com/2020/03/the-pure-pattern.html

josephcsible:
Fair point. Here's two ideas:

1. Make this change for `settings3` and `useSrc`, since they aren't shadowing anything
2. For the rest of the variables, rename them to have more descriptively different names than just a `'`

Do you like either or both of these?

src/HLint.hs
```diff
@@ -153,8 +153,8 @@ runHints args settings cmd@CmdMain{..} = do
     j <- if cmdThreads == 0 then getNumProcessors else pure cmdThreads
     withNumCapabilities j $ do
         let outStrLn = whenNormal . putStrLn
-        ideas <- getIdeas cmd settings
-        ideas <- pure $ if cmdShowAll then ideas else  filter (\i -> ideaSeverity i /= Ignore) ideas
+        ideas' <- getIdeas cmd settings
+        let ideas = if cmdShowAll then ideas' else  filter (\i -> ideaSeverity i /= Ignore) ideas'
```
ndmitchell: If you do want to do this, making it `(if cmdShowAll then id else filter ...) <$> getIdeas cmd settings` looks better.

ndmitchell:
For the first 2, sure, happy to move to `let`.

I reviewed the rest. In 1 case I think the let could be removed without any substantial harm. However, in all the other cases there isn't a good name to rename it to... What's the dislike of pure bind? If you can get a GHC proposal through to make `let` non-recursive and introduce `letrec` I'd very happily adopt that in HLint!

josephcsible:
My dislike of pure bind is that I think it's less efficient, and also slightly less clear. In particular, because it enters the monadic context twice for no reason. For example, it feels like doing `map f . map g` instead of `map (f . g)`, `x <- f <$> getX; g x` instead of `x <- getX; g (f x)`, or I guess most simply, `m >>= return` instead of just `m` (this is what it will actually desugar to, modulo `pure = return`). I'd be more okay with this if I knew GHC had a rewrite rule to make this go away, but I don't think it does. (And as I say that, I realize that this technique will also break `ApplicativeDo`.)

src/HLint.hs
josephcsible: I don't like using `fmap` or `<$>` on the RHS of `<-` for basically the same reason that I don't like `<- pure` (see my other comment below).

josephcsible: I feel like you probably won't like this, but how about `(if cmdShowAll then id else filter ... -> ideas) <- getIdeas cmd`?

josephcsible:
TIL. In that case, I'll edit this PR to just do point 1 (`settings3` and `useSrc`), and leave the rest as-is.

ndmitchell:
http://blog.ezyang.com/2011/05/unraveling-the-mystery-of-the-io-monad/ is a good resource for understanding IO under the hood.

ndmitchell:
Thanks!

PR #912, opened 2020-03-15
Get HSE out of Extensions

shayne-fletcher:
This is terribly hacky. We'll have to iterate on this however, this gets HSE's Extension type out of 'Extensions.hs' and (amazingly) all the tests pass.

~I've temporarily disabled the check  for ScopedTypeVariables (`-- used "ScopedTypeVariables" = hasS isPatTypeSig`) - this is because despite it being whitelisted in .hlint.yaml, it keeps reporting on this module - perhaps you can determine why~

I'm pretty sure this PR incidentally fixes a bug relating to detecting `ScopedTypeVariables` which explains why I've been removing that extension from files that have been deemed to not use it. I've had to add an exclusion and continue using it in `Brackets`. See `.hlint.yaml` for why (can't type the file without it).


shayne-fletcher:
That will have to do for today - that was tough. Hope you find it a good starting point. Would appreciate some collaboration on this one 😄

src/HSE/Util.hs
```diff
@@ -1,30 +1,48 @@
...
     ) where
 
+import GHC.Util.Outputable
+import GHC.LanguageExtensions.Type
+import Data.Maybe
 import qualified Data.Map as Map
 import qualified Language.Haskell.GhclibParserEx.DynFlags as GhclibParserEx
-import Language.Haskell.Exts(Extension(..), parseExtension)
 
+-- Because we don't have Ord or Data.Data.Data we fall back on String
```
shayne-fletcher: No `Ord` so extensions can't be keys, no `Data` so that rules out using `HsExtendInstances` which was my first idea.

shayne-fletcher:
> I've temporarily disabled the check for ScopedTypeVariables (-- used "ScopedTypeVariables" = hasS isPatTypeSig) - this is because despite it being whitelisted in .hlint.yaml

Could relate some how to HSE calling it `PatternSignatures`.

shayne-fletcher:
AppVeyor has failed due to some transient download failure involving aws hosting.

src/Hint/Extensions.hs
```diff
@@ -182,26 +182,37 @@ import Language.Haskell.GhclibParserEx.GHC.Hs.Pat
 import Language.Haskell.GhclibParserEx.GHC.Hs.Expr
 import Language.Haskell.GhclibParserEx.GHC.Hs.Types
 import Language.Haskell.GhclibParserEx.GHC.Hs.Decls
+import qualified Language.Haskell.GhclibParserEx.DynFlags as GhclibParserEx
+import GHC.LanguageExtensions.Type
+
+-- Terrible hack.
```
ndmitchell: Why does this hack exist? GHC must have mechanisms for parsing extension names, so why can't we reuse that?

src/Extension.hs
```diff
@@ -41,3 +47,30 @@ toHseEnabledExtensions :: [Extension] -> [HSE.Extension]
 toHseEnabledExtensions = mapMaybe ((HSE.EnableExtension <$>) . readMaybe . show')
   where
     show' e = if ex == "Cpp" then "CPP" else ex where ex = show e
+
+-- Oprhan instance. Bad.
+instance Ord Extension where
+  compare = compare `on` show
```
ndmitchell: Should this move into ghc-lib-parser-ex? Then it can be deleted at the appropriate time that it gets upstreamed to GHC?

It should also probably use the compare on fromEnum rather than show.

src/Hint/Extensions.hs
```diff
@@ -304,10 +315,9 @@ used PatternGuards = hasS f
     g [L _ BodyStmt{}] = False
     g _ = True
 used StandaloneDeriving = hasS isDerivD
-used PatternSignatures = hasS isPatTypeSig
+used ScopedTypeVariables = hasS isPatTypeSig
```
ndmitchell: This is wrong. There was an extension that allowed you to put pattern signatures in, but if GHC lacks it, then just delete this clause. ScopedTypeVariables is a different and more powerful extension which we can't detect presence of absence of.

ndmitchell:
Overall, it does look pretty solid though - a bit of revision and we should be good to land!

shayne-fletcher:
> I took the removal of some scoped type variables, which we can't detect and which are useful, and deleted them and upstreamed that, so this patch is just the extension stuff without the unrelated pieces.

Can you clarify what you mean by this?

src/Hint/Extensions.hs
shayne-fletcher: I'll look into it.

src/Extension.hs
shayne-fletcher: Yes, let's move it down.

src/Hint/Extensions.hs
shayne-fletcher: Ok. This clause will be deleted.

ndmitchell:
In some places you removed ScopedTypeVariables extensions that weren't useful, and since that added noise, I went through all the instances in existing HLint and deleted the ones that aren't useful and deleted them. That way we can focus this patch on just the extensions, especially since it seems the removal of ScopedTypeVariables was a bug.

ndmitchell:
If we are going to move the Ord instance out to ghc-lib-parser-ex then I suggest we do it first, since otherwise when it does get moved it will break HLint's that don't know about the newer version, requiring a 0.1 bump of something or other.

src/Extension.hs
shayne-fletcher: Done in https://github.com/shayne-fletcher/ghc-lib-parser-ex/pull/31 and https://github.com/ndmitchell/hlint/pull/914

src/Hint/Extensions.hs
shayne-fletcher: Cracked it.

shayne-fletcher:
After the most recent revision, assuming this passes CI, I judge this good to go.

shayne-fletcher:
> If we are going to move the Ord instance out to ghc-lib-parser-ex then I suggest we do it first, since otherwise when it does get moved it will break HLint's that don't know about the newer version, requiring a 0.1 bump of something or other.

That's waiting on your merge (the bump) in https://github.com/ndmitchell/hlint/pull/914.

shayne-fletcher:
If we get into a rebase situation after merging the above, I'll take care of it.

src/Hint/Extensions.hs
```diff
@@ -188,31 +190,41 @@ extensionsHint _ x =
...
     ]
   where
+    filterEnabled :: [String] -> [String]
+    filterEnabled  = filter (not . isPrefixOf "No")
+
+    lookupExt :: String -> Extension
+    lookupExt s =
+      case find (\(FlagSpec n _ _ _) -> n == s) xFlags of
+        Just f -> flagSpecFlag f
+        Nothing -> error ("IMPOSSIBLE: Unrecognized language extension '" ++ s ++ "'")
```
ndmitchell: Is this branch literally impossible to hit? How so?

shayne-fletcher: If it has been compiled it has been given to `parsePragmasIntoDynFlags` which has given it to `getOptions` which does a check that the given extension is valid. If it's not valid, a parse error will result. Therefore we are assured that the extensions listed in the pragmas of the module are kosher.

ndmitchell: Do we have a test that an invalid pragma leads to a parse error "hint"?

shayne-fletcher: I haven't written anything however, if I give it a module
```
{-# LANGUAGE Foo #-}
```
The result is early termination with `Error: Parse error`.

shayne-fletcher: It will have gone down this case in `parseModuleEx`:
```
          Left msg -> do
            -- Parsing GHC flags from dynamic pragmas in the source
            -- has failed. When this happens, it's reported by
            -- exception. It's impossible or at least fiddly getting a
            -- location so we skip that for now. Synthesize a parse
            -- error.
            let loc = GHC.mkSrcLoc (mkFastString file) (1 :: Int) (1 :: Int)
            pure $ Left (ParseError (GHC.mkSrcSpan loc loc) msg "")
```

shayne-fletcher: We can't get to `extensionHint` without a `ParseModuleEx {}` and if we have one of those, we know the extensions expressed by the pragmas have been validated.

ndmitchell: Cool. Can we have slightly more comment on this Impossible case then. And can we have a test in this module with `{-# LANGUAGE InvalidExtension #-} -- @ERROR ParseError` (or however we express it) so the test and the code relying on it are nearby.

ndmitchell: (Assuming that's fairly easy to do, don't sweat if it turns out to be non-obvious!)

shayne-fletcher: It would seem the message originates here:
```
-- | Return either an idea (a parse error) or the module. In IO because might call the C pre processor.
parseModuleApply :: ParseFlags -> [Setting] -> FilePath -> Maybe String -> IO (Either Idea ModuleEx)
parseModuleApply flags s file src = do
    res <- parseModuleEx (parseFlagsAddFixities [x | Infix x <- s] flags) file src
    case res of
      Right r -> pure $ Right r
      Left (ParseError sl msg ctxt) ->
            pure $ Left $ classify [x | SettingClassify x <- s] $ rawIdeaN Error "Parse error" sl ctxt Nothing []
```

shayne-fletcher: Done and done.

ndmitchell:
Thanks!

PR #919, opened 2020-03-18
more specific names for two fusion rules

jvoigtlaender:
I found it odd that `hlint` currently gives such a warning:
```
Fuse foldr/map
Found:
  foldr f z (map g x)
Perhaps:
  foldr (f . g) z x
```
but also this:
```
Use .
Found:
  foldMap f (map g x)
Perhaps:
  foldMap (f . g) x
```
The two situations in code are very similar, and saying `Fuse ....` seems more useful to me than `Use .` also in the second situation.

This PR introduces names for rules such that the second situation will be notified analogously to the first one.


ndmitchell:
Thanks! In general names are auto generated. Most of the time they're reasonable. These aren't very good.

Issue #915, opened 2020-03-16
Fix parse error message


ndmitchell:
```
tests/ignore-parse-error3.hs:1:1: Error: Parse error
Found you should remove it.
```

That shouldn't be. I suspect we don't turn it into a ParseError properly.

PR #921, opened 2020-03-19
Improve the situation with parse error messages

shayne-fletcher:
We now display the actual GHC message rather than "Parse error". As an example, in the case of the recently considered invalid extension case we print,
```
tests/ignore-parse-error3.hs:1:1: Error: Unsupported extension: InvalidExtension

Found:
  {-# LANGUAGE InvalidExtension #-}

1 hint
```
If the error results from parsing dynamic pragmas, we just use the entire module source for the context at this time (rather than "").

Closes https://github.com/ndmitchell/hlint/issues/915.

### DIFF
--- a/src/Apply.hs
+++ b/src/Apply.hs
@@ -87,7 +87,7 @@ parseModuleApply flags s file src = do
     case res of
       Right r -> pure $ Right r
       Left (ParseError sl msg ctxt) ->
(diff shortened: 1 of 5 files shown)


shayne-fletcher:
Another example.

Screen Shot 2020-03-18 at 8 55 42 PM

ndmitchell:
Great! For historical context, the parse errors from HSE were so awful we realised that giving them to the user was more harmful than ignoring them. Fortunately GHC is a lot better at that.

PR #922, opened 2020-03-19
Reduce the SrcSpan of "Redundant do" to only cover "do"

zliu41:
This is so that the entire do block doesn't get highlighted in my editor.

Before:
Screenshot from 2020-03-18 22-09-23

After:
Screenshot from 2020-03-18 22-10-05


ndmitchell:
Thanks! Great idea.

PR #924, opened 2020-03-21
Remove use of HSE extensions entirely

shayne-fletcher:
Remove use of HSE's `Extension`. As far as I can tell, `HSE/All.hs` is now "HSE free" - parse modes and parse flags are expressed only in ghc-lib types.

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -60,7 +60,6 @@ library
         cpphs >= 1.20.1,
         cmdargs >= 0.10,
         yaml >= 0.5.0,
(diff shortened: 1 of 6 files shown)


shayne-fletcher:
In this PR, `haskell-src-exts` is removed from `hlint.cabal`.

shayne-fletcher:
One Travis failure "An error occurred while generating the build script." 🤣

ndmitchell:
Wow! I had no idea we were this close!!! Amazing work.

I'm going to assume this is a Travis-side wobble, so will pass once Mac is back.

PR #926, opened 2020-03-23
Fix refactoring when the hint contains _noParen_

zliu41:
```haskell
yes = f ((,) (2 + 3))
```

Before:
```haskell
yes = f (_noParen_ (2 + 3),)
```

After:
```haskell
yes = f (2 + 3,)
```


src/GHC/Util/Unify.hs
```diff
@@ -58,6 +58,12 @@ validSubst' eq = fmap Subst' . mapM f . groupSort . fromSubst'
     where f (x, y : ys) | all (eq y) ys = Just (x, y)
           f _ = Nothing
 
+-- Remove unnecessary brackets from a Subst'. The first argument is a list of unification variables
+-- for which brackets should be removed from their substitutions.
+removeParens :: [LHsExpr GhcPs] -> Subst' (LHsExpr GhcPs) -> Subst' (LHsExpr GhcPs)
```
ndmitchell:
```suggestion
removeParens :: [String] -> Subst' (LHsExpr GhcPs) -> Subst' (LHsExpr GhcPs)
```

Would be better for the map unsafePrettyPrint to call from the caller side, since this API makes no sense if you don't pass a list of strings here.

ndmitchell:
Thanks! After looking at your patch I decided `_eval_` can die since it's not used anywhere and is somewhat odd as hints go - so I've removed that entirely.

Issue #927, opened 2020-03-24
Upgrade to ghc-8.10.1


shayne-fletcher:
An 8.10 compatible `ghc-lib-parser` will be going to hackage today (`ghc-lib-parser-ex` won't be long behind). Time to get a branch going for the conversion.

shayne-fletcher:
- A heads up that I intend to have a crack at this Saturday; I hope that it'll be a half day job but well, famous last words and all that 😀

ndmitchell:
Sounds great!

ndmitchell:
BTW, be aware of https://github.com/mpickering/apply-refact/issues/53 which is causing widespread CI failure.

shayne-fletcher:
This is looking sweet.
```
Testing
Source annotations .................
Input/outputs .....................................................
Hint names ..
Hint annotations ..

Tests passed (705)
```

I skipped the refactor tests (is there a good way to opt to do that? I fake it at DA by giving a bogus refactor path, here today I've just deleted the refactor exe- if there isn't could one be provided please?)

shayne-fletcher:
PR https://github.com/ndmitchell/hlint/pull/934 up for review.

ndmitchell:
Done! Nice work Shayne!

PR #928, opened 2020-03-24
Upgrade ghc-lib-parser-ex to 8.8.6.0

shayne-fletcher:
ghc-lib-parser-ex 8.8.6.0 moves `DynFlags` to `GHC.Driver.Session`, this PR upgrades hlint to that version and accounts for the move. This is already on hackage - I forgot about stackage and got us an out-of-bounds notice for hlint  (https://github.com/commercialhaskell/stackage/issues/5255). If you land this we should be good again.

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -66,7 +66,7 @@ library
         refact >= 0.3,
         aeson >= 1.1.2.0,
         filepattern >= 0.1.1,
-        ghc-lib-parser-ex >= 8.8.5.8 && < 8.8.6
+        ghc-lib-parser-ex >= 8.8.6.0 && < 8.8.7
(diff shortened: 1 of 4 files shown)


shayne-fletcher:
Very unclear what's going on with the builds. Appveyor failed with no log on the first run, no sign of Travis even registering that its got work to do :-/

shayne-fletcher:
FWIW, I have pushed an 8.10.0.0 to hackage. I presume that too will generate an out-of-bounds due to hlint preventing it from going to stackage. That's of course fine with me while hlint is not upgraded to 8.10 (https://github.com/ndmitchell/hlint/issues/927) - presumably that will have no impact on 8.8.6.0 going in.

ndmitchell:
Thanks! Sure, we should hold ghc-lib-parser-ex until your work to upgrade completes. That seems good.

PR #930, opened 2020-03-26
Fix spans of lambda hints

zliu41:
hlint 2.2.11:
Screenshot from 2020-03-25 17-30-49

hlint head:
Screenshot from 2020-03-25 17-29-13

This PR restores the 2.2.11 behavior.


ndmitchell:
Thanks!

PR #931, opened 2020-03-26
Add StaticPointers to reallyBadExtensions

### LINKED ISSUE #929: Do block not detected
The following code
```
normaliseStatic :: I.Term -> CC I.Term
normaliseStatic v@(I.Def f es) = lift $ do
  static <- isStaticFun . theDef <$> getConstInfo f
  if static then normalise v else pure v
normaliseStatic v = pure v
```
produces
```
    parse error on input `<-'
    Perhaps this statement should be within a 'do' block?
Found:
  > 
```
GHC (8.x.y) seems to compile it OK.

zliu41:
Fixes #929.

### DIFF
--- a/src/Extension.hs
+++ b/src/Extension.hs
@@ -21,6 +21,7 @@ badExtensions =
 
 reallyBadExtensions =
   [ TransformListComp -- steals the group keyword
+  , StaticPointers -- steals the static keyword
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks!

PR #933, opened 2020-03-28
Update ideaFrom and ideaTo for "Redundant do"

zliu41:
In #922, the span of "Redundant do" is reduced to just cover the "do". So it seems a good idea to make this change.

Before:
```
Found: do <stmt>  Perhaps: <stmt>
```

After:
```
Found: do  Perhaps you should remove it.
```

This doesn't affect refactoring, because `ideaRefactoring` is unchanged.

### DIFF
--- a/src/Hint/Monad.hs
+++ b/src/Hint/Monad.hs
@@ -13,7 +13,7 @@ yes = do mapM print a; return b -- mapM_ print a
 yes = do _ <- mapM print a; return b -- mapM_ print a
 no = mapM print a
 no = do foo ; mapM print a
-yes = do (bar+foo) -- (bar+foo)
+yes = do (bar+foo) --
 no = do bar ; foo
 yes = do bar; a <- foo; return a -- do bar; foo
 no = do bar; a <- foo; return b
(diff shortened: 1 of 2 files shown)


src/Hint/Monad.hs
```diff
@@ -40,9 +40,9 @@ folder f a xs = foldM f a xs >>= \_ -> return () -- foldM_ f a xs
 yes = mapM async ds >>= mapM wait >> return () -- mapM async ds >>= mapM_ wait
 main = "wait" ~> do f a $ sleep 10
 main = print do 17 + 25  @NoRefactor: needs -XBlockArguments which isn't available before GHC 8.6
-main = print do 17 -- 17 @NoRefactor
-main = f $ do g a $ sleep 10 -- g a $ sleep 10
-main = do f a $ sleep 10 -- f a $ sleep 10
+main = print do 17 -- @NoRefactor
```
ndmitchell: Why is this one NoRefactor?

ndmitchell:
Great idea! These hints often look like garbage, and scroll on for pages, so well worth fixing.

src/Hint/Monad.hs
zliu41: apply-refact (0.6.0.0) crashes if the BlockArgument pragma is absent.

ndmitchell: Ah, cool - I note the line above even explains it. I copied the explanation on to that line too. Thanks.

PR #934, opened 2020-03-28
Update to ghc-8.10.1

shayne-fletcher:
Switch to the 8.10.1 parse tree.

### DIFF
--- a/src/GHC/Util.hs
+++ b/src/GHC/Util.hs
@@ -38,17 +38,22 @@ import GHC.Util.Unify
 import qualified Language.Haskell.GhclibParserEx.Parse as GhclibParserEx
 import Language.Haskell.GhclibParserEx.GHC.Driver.Session (parsePragmasIntoDynFlags)
 
-import HsSyn
+import GHC.Hs
 import Lexer
 import SrcLoc
 import DynFlags
 import FastString
+import RdrHsSyn
 
 import System.FilePath
 import Language.Preprocessor.Unlit
 
 parseExpGhcLib :: String -> DynFlags -> ParseResult (LHsExpr GhcPs)
-parseExpGhcLib = GhclibParserEx.parseExpression
+parseExpGhcLib s flags =
+  case GhclibParserEx.parseExpression s flags of
+    POk s e ->
+       unP (runECP_P e >>= \e -> return e) s :: ParseResult (LHsExpr GhcPs)
+    PFailed ps -> PFailed ps
 
 parseImportGhcLib :: String -> DynFlags -> ParseResult (LImportDecl GhcPs)
 parseImportGhcLib = GhclibParserEx.parseImport
--- a/src/HSE/All.hs
+++ b/src/HSE/All.hs
@@ -25,15 +25,16 @@ import Extension
 import FastString
 import Prelude
 
-import qualified HsSyn
+import GHC.Hs
 import qualified SrcLoc as GHC
-import qualified ErrUtils
+import ErrUtils
 import qualified Outputable
 import qualified Lexer as GHC
 import GHC.LanguageExtensions.Type
 import qualified ApiAnnotation as GHC
 import qualified BasicTypes as GHC
 import qualified DynFlags as GHC
+import Bag
 
 import GHC.Util (parsePragmasIntoDynFlags, parseFileGhcLib, parseExpGhcLib, parseDeclGhcLib, parseImportGhcLib, baseDynFlags)
 import qualified Language.Haskell.GhclibParserEx.Fixity as GhclibParserEx
@@ -85,7 +86,7 @@ data ParseError = ParseError
 
 -- | Result of 'parseModuleEx', representing a parsed module.
 data ModuleEx = ModuleEx {
-    ghcModule :: GHC.Located (HsSyn.HsModule HsSyn.GhcPs)
+    ghcModule :: GHC.Located (HsModule GhcPs)
   , ghcAnnotations :: GHC.ApiAnns
 }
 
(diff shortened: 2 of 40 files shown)


src/Hint/List.hs
```diff
@@ -132,7 +132,10 @@ moveGuardsForward = reverse . f [] . reverse
               span (if any hasPFieldsDotDot (universeBi x)
                        || any isPFieldWildcard (universeBi x)
                       then const False
-                      else \x -> pvars' p `disjoint` vars_ x) guards
+                      else \x ->
+                       let pvars = pvars' p in
+                       pvars `disjoint` vars_ x && "pun-right-hand-side" `notElem` pvars
```
ndmitchell: What is "pun-right-hand-side"?

src/GHC/Util.hs
```diff
@@ -38,17 +38,22 @@ import GHC.Util.Unify
...
 import DynFlags
 import FastString
+import RdrHsSyn
 
 import System.FilePath
 import Language.Preprocessor.Unlit
 
 parseExpGhcLib :: String -> DynFlags -> ParseResult (LHsExpr GhcPs)
-parseExpGhcLib = GhclibParserEx.parseExpression
+parseExpGhcLib s flags =
```
ndmitchell: Why does this need to change? Do we not have parseExpression in ghc-lib-parser-ex?

src/HSE/All.hs
```diff
@@ -171,8 +172,15 @@ parseModuleEx flags file str = timedIO "Parse" file $ do
                     pure $ Right (ModuleEx a' anns)
                 -- Parse error if GHC parsing fails (see
                 -- https://github.com/ndmitchell/hlint/issues/645).
-                GHC.PFailed _ loc err ->
-                    ghcFailOpParseModuleEx ppstr file str (loc, err)
+                GHC.PFailed s -> do
+                  -- TODO (SF, 2020-03-28): Quick patch here. After the
+                  -- migration to 8.10, this needs looking at more
+                  -- closely.
```
ndmitchell: I concur. Did you see my patch earlier today to the PFailed branch?

tests/json.test
```diff
@@ -28,7 +28,7 @@ RUN tests/json-parse-error.hs --json
 FILE tests/json-parse-error.hs
 @
 OUTPUT
-[{"module":[],"decl":[],"severity":"Error","hint":"Parse error: on input `@'","file":"tests/json-parse-error.hs","startLine":1,"startColumn":1,"endLine":1,"endColumn":2,"from":"> @\n","to":null,"note":[],"refactorings":"[]"}]
+[{"module":[],"decl":[],"severity":"Error","hint":"Parse error: tests/json-parse-error.hs:1:1: error: parse error on input `@'","file":"tests/json-parse-error.hs","startLine":1,"startColumn":1,"endLine":1,"endColumn":2,"from":"> @\n","to":null,"note":[],"refactorings":"[]"}]
```
ndmitchell: This is the parse quick hack above.

tests/lhs.test
```diff
@@ -25,7 +25,8 @@ BUG 331
 module
 \end{code}
 OUTPUT
-tests/lhs-line-numbers2.lhs:10:1: Error: Parse error: possibly incorrect indentation or mismatched brackets
+tests/lhs-line-numbers2.lhs:10:1: Error: Parse error: tests/lhs-line-numbers2.lhs:10:1: error:
+    parse error (possibly incorrect indentation or mismatched brackets)
```
ndmitchell: The parse quick hack.

src/Hint/List.hs
shayne-fletcher: See this code from `RdrHsSyn.hs`:
```
plus_RDR, pun_RDR :: RdrName
plus_RDR = mkUnqual varName (fsLit "+") -- Hack
pun_RDR  = mkUnqual varName (fsLit "pun-right-hand-side")
```
It comes up in the `pvar'` on this test case
```
issue619 = [pkgJobs | Pkg{pkgGpd, pkgJobs} <- pkgs, not $ null $ C.condTestSuites pkgGpd]
```

ndmitchell:
I just introduced and solved a merge commit on your branch, so you will need to pull/merge before rebasing.

src/GHC/Util.hs
shayne-fletcher: We do, but at this time, `parseExpression` its definition is this:
```
#if defined (GHCLIB_API_811) || defined (GHCLIB_API_810)
parseExpression :: String -> DynFlags -> ParseResult RdrHsSyn.ECP
#else
parseExpression :: String -> DynFlags -> ParseResult (LHsExpr GhcPs)
#endif
parseExpression = parse Parser.parseExpression
```
I thought when the dust settles I'll fix it up there so that code won't feature in hlint.

src/HSE/All.hs
shayne-fletcher: No, I'll have a look

tests/json.test
shayne-fletcher: Yeah.

tests/lhs.test
shayne-fletcher: Nod.

src/Hint/List.hs
ndmitchell: That is an amazing hack in the GHC code base. Yuk! But fair enough, makes sense then - can you add a comment just above saying where that magic string is coming from. Any chance we can extract it from placeHolderPunRhs rather than hardcoding it?

src/GHC/Util.hs
ndmitchell: Makes a lot of sense.

src/Hint/List.hs
shayne-fletcher: I put a comment for now. Not following what you are saying yet in 
```
Any chance we can extract it from placeHolderPunRhs rather than hardcoding it?
```

src/HSE/All.hs
shayne-fletcher: Fixed this back to the old-style.

shayne-fletcher:
@ndmitchell Addressed the comments so far in the last two commits. Please cast your eye over once more.

src/GHC/Util.hs
shayne-fletcher: Raised an issue for the next release https://github.com/shayne-fletcher/ghc-lib-parser-ex/issues/37

shayne-fletcher:
Oh, and there are some new test failures but I judge they have come in on your merge from master just now.

ndmitchell:
Woohoo! Let's merge and then if there are any CI errors I'll fix them up in situ.

src/Hint/List.hs
ndmitchell: It's fine with the comment. I can't believe GHC went there...

PR #939, opened 2020-03-29
Improve "Redundant where" hint

zliu41:
Similar to what was done to "redundant do": limiting the span to just cover the "where", and changing the output to "Found: where, Perhaps you should remove it".

I've verified that the span is correct even if there are spaces after the "where".


ndmitchell:
Thanks!

PR #940, opened 2020-03-30
Update to ghc-lib-parser-ex-8.10.0.2

shayne-fletcher:
Solves the configure problem so that hlint works out of the box with no flags and no compromises.

## 8.10.0.2 released 2020-03-30
- Rework cabal flags; allow full configurability with a good default:
  - Have two flags `auto` and `no-ghc-lib`. Default behavior exactly as `hlint` linking `ghc-lib-parser-8.10.*` if not on `ghc-8.10.*` and `ghc-8.10.*` otherwise.

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -67,24 +67,16 @@ library
         aeson >= 1.1.2.0,
         filepattern >= 0.1.1
 
-    -- We currently target the 8.10 parse tree.
     if !flag(ghc-lib) && impl(ghc >= 8.10.0) && impl(ghc < 8.11.0)
       build-depends:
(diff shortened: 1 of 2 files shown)


shayne-fletcher:
There are two test failures not related and pre-existing on master.

shayne-fletcher:
As discussed elsewhere, ghc-8.4.4 support has to be dropped. GHC no longer supports bootstrapping with compiler versions < 8.6.5 .

ndmitchell:
Thanks, will take this on faith and work to get the CI back.

shayne-fletcher:
Trust but verify please 😉

PR #941, opened 2020-03-31
Add BlockArguments to a few tests

zliu41:
Since GHC 8.4 is no longer supported, adding BlockArguments to some tests so that refactoring tests pass.


ndmitchell:
Thanks!

PR #944, opened 2020-04-04
Reduce the span of "use let"

zliu41:
Reduce the span and the output to cover the problematic statements only.

Before:
Screenshot from 2020-04-03 22-06-52

After:
Screenshot from 2020-04-03 22-09-35


ndmitchell:
Looks awesome. I'll just wait for the CI to weigh in then merge it.

PR #945, opened 2020-04-04
Suggest Map.empty over Map.fromList []

### LINKED ISSUE #111: Suggestion for Map.fromList []
HLint doesn't show suggestion for `Map.fromList []`. It could suggest using `Map.empty`.

maartenJacobs:
Suggestion implemented but help is needed with namespaces: `Data.Map` has 3 namespaces, all of which export `fromList` and `empty`. From the tests it seems like `hlint` is doing the right thing suggesting the used namespace, but I'm wondering if there isn't a way to prevent this duplication.

Closes #111

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -758,6 +758,12 @@
     - warn: {lhs: null x  , rhs: "False", side: isTuple x, name: Using null on tuple}
     - warn: {lhs: length x, rhs: "1"    , side: isTuple x, name: Using length on tuple}
 
+    # MAP
+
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks for the patch! In reality these are three totally different functions, so it makes sense they have to be separate. However, the duplication is a bit troublesome - we have similar instances with things like Data.List reexporting things in Prelude, and fmap and `<$>`  being morally the same. For now the approach is just to add duplication, and once we have enough duplication, sit down and figure out how we common it up once more.

PR #946, opened 2020-04-05
Upgrade ghc-lib-parser-ex

shayne-fletcher:
## 8.10.0.4 released 2020-04-04
- Add expression predicates `isWholeFrac`, `isFieldPunUpdate`, `isStrictMatch`, `isMultiIf`, `isProc`, `isTransStmt`;
- Add pattern predicate `isPFieldPun`.

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -76,7 +76,7 @@ library
       build-depends:
           ghc-lib-parser == 8.10.*
     build-depends:
-        ghc-lib-parser-ex >= 8.10.0.3 && < 8.10.1
+        ghc-lib-parser-ex >= 8.10.0.4 && < 8.10.1
 
     if flag(gpl)
         build-depends: hscolour >= 1.21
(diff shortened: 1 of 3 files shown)


shayne-fletcher:
Tests all good locally (refactoring tests skipped). Appveyor failure unrelated I believe.

Screen Shot 2020-04-04 at 9 40 58 PM

shayne-fletcher:
Ran again locally with refactor tests. Unable to reproduce (macOS).

Screen Shot 2020-04-04 at 9 51 58 PM

ndmitchell:
We know not having this breaks, so let's merge!

shayne-fletcher:
I concur that's the right thing to do. I'm also very confident the exhibited issue is unrelated to this. In any case, I'll put a hold on any further work on this front until we get the release out the door!

ndmitchell:
I think now when we don't have any releases out the door is exactly the time to do it! Had we had a release out the door we'd have broken end users. Breaking the odd developer for a small amount of time isn't nearly as bad.

PR #949, opened 2020-04-11
Warn about redundant fmaps with binds

josephcsible:


data/hlint.yaml
```diff
@@ -415,6 +415,12 @@
     - warn: {lhs: liftA, rhs: fmap}
     - hint: {lhs: m >>= return . f, rhs: f <$> m}
     - hint: {lhs: return . f =<< m, rhs: f <$> m}
+    - warn: {lhs: fmap f x >>= g, rhs: x >>= g . f}
+    - warn: {lhs: f <$> x >>= g, rhs: x >>= g . f}
+    - warn: {lhs: x Data.Functor.<&> f >>= g, rhs: x >>= g . f}
+    - warn: {lhs: g =<< fmap f x, rhs: g . f =<< x}
+    - warn: {lhs: g =<< f <$> x, rhs: g . f =<< x}
+    - warn: {lhs: g =<< (x Data.Functor.<&> f), rhs: g . f =<< x}
```
ndmitchell, line 423: I assume this bracket is required? But if not, it will reduce the number of matches.

josephcsible: Yep, it's required. `=<<` is `infixr 1` and `<&>` is `infixl 1`, so there'd be a parse error without the parentheses.

PR #950, opened 2020-04-11
Improve "Redundant $" hint

zliu41:
Same as what was done to "Redundant do" and "Redundant where": reduce the span and suggest "Perhaps you should remove it."

Tested in VSCode:

Screenshot from 2020-04-11 12-44-13


ndmitchell:
Thanks!

PR #954, opened 2020-04-18
Use suggested replacement as refactoring template if any substitution has no loc

zliu41:
Previously `foo = typeOf (undefined :: a)` fails to refactor, because
```
template = typeRep (Proxy :: Proxy a)
substs = [ a -> undefined ]
```
So hlint is telling apply-refact to refactor it into `typeRep (Proxy :: Proxy undefined)`.

The template is not `typeRep (Proxy :: Proxy b)` because the substitution `b -> a` has no loc, so `b` is expanded into `a` in the template (#895).

This can be fixed by: if any substitution in `u` has no loc, set `u` to `mempty`, and use the suggested replacement (`res`) as the refactoring template (`tpl`). In the example above it becomes
```
template = typeRep (Proxy :: Proxy a)
substs = []
```

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -1159,7 +1159,6 @@
 # no = sort <$> f input `shouldBe` sort <$> x
 # sortBy (comparing length) -- sortOn length
 # myJoin = on $ child ^. ChildParentId ==. parent ^. ParentId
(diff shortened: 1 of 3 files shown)


ndmitchell:
This conflicts with the other patch of yours I just merged. Any chance you can rebase it on master?

zliu41:
@ndmitchell Done. Thanks for merging the PRs, and take good care!

PR #957, opened 2020-04-24
Update pass criteria for refactoring tests

zliu41:
Previously:
- If there's no hint, the refactoring output should be identical to the input (ignoring spaces and semicolons)
- If there's a hint, `ideaTo` should be a substring of the refactoring output (ignoring spaces and semicolons)

Now, additionally:
- If `ideaTo == Nothing` (meaning the hint is a parse error), skip the refactoring test.

  It doesn't make sense to test refactoring when there's a parse error.

- If `ideaTo == Just ""` (i.e., the suggestion is "perhaps you should remove it"), the refactoring output should be a _proper subsequence_ of the input (ignoring spaces and semicolons).

  This ensures the "redundant x" hints are tested properly. Previously refactoring tests for "redundant x" hints trivially pass (as long as `refactor` doesn't crash) because `""` is a substring of any string.

### DIFF
--- a/src/Hint/Bracket.hs
+++ b/src/Hint/Bracket.hs
@@ -74,7 +74,7 @@ foo = (case x of y -> z; q -> w) :: Int
 main = do a += b . c; return $ a . b
 
 -- <$> bracket tests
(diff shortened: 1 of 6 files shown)


ndmitchell:
Thanks - I appreciate all the work you've done to improve the tests!

PR #960, opened 2020-05-02
Update ghc-lib, ghc-lib-parser-ex

shayne-fletcher:
## 8.10.0.5 released 2020-05-02
- New modules
  - `Language.Haskell.GhclibParserEx.GHC.Parser`, `Language.Haskell.GhcLibParserEx.GHC.Utils.Outputable` to replace `Language.Haskell.GhclibParserEx.Parse` and `Language.Haskell.GhclibParserEx.Outputable` (which remain for now but deprecated and will be removed in a future release)

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -76,7 +76,7 @@ library
       build-depends:
           ghc-lib-parser == 8.10.*
     build-depends:
-        ghc-lib-parser-ex >= 8.10.0.4 && < 8.10.1
+        ghc-lib-parser-ex >= 8.10.0.5 && < 8.10.1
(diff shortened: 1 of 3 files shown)


ndmitchell:
Given we have an upper bound of 8.10.1 on ghc-lib-parser-ex, will these modules stay around until at least 8.10.2?

shayne-fletcher:
> Given we have an upper bound of 8.10.1 on ghc-lib-parser-ex, will these modules stay around until at least 8.10.2?

Yes. No problem. They can stay as long as we need.

Issue #961, opened 2020-05-02
hlint-3.0 regression: Unrecognized language extension 'Trustworthy'


RyanGlScott:
Consider the following program:

```hs
{-# LANGUAGE Trustworthy #-}
module Bug where
```

On `hlint-2.2.11`, this produces no hints, as expected:

```
$ hlint --version
HLint v2.2.11, (C) Neil Mitchell 2006-2020
$ hlint Bug.hs
No hints
```

With `hlint-3.0`, however, it crashes:

```
$ hlint --version
HLint v3.0, (C) Neil Mitchell 2006-2020
$ hlint Bug.hs
hlint: IMPOSSIBLE: Unrecognized language extension 'Trustworthy'
CallStack (from HasCallStack):
  error, called at src/Hint/Extensions.hs:266:11 in hlint-3.0-b26bf2d7d97a0573cc460418b91c96800f45ab73c13e16368e3465671e6a15bf:Hint.Extensions
```

RyanGlScott:
`hlint-3.0` crashes similarly if you enable `Safe` or `Unsafe`.

ndmitchell:
Eek. Thanks for the report. Looking into it now. Seems to be caused because Trustworthy doesn't have an entry in the GHC extensions data type: https://hackage.haskell.org/package/ghc-lib-parser-8.10.1.20200412/docs/GHC-LanguageExtensions-Type.html#t:Extension. CC @shayne-fletcher

RyanGlScott:
The `Haskell98` and `Haskell2010` extensions also trigger the same crash.

ndmitchell:
Thanks - I should have a fix within about 15 mins.

ndmitchell:
Fix pushed. I'll wait for CI to kick in (15 mins or so) then make a release of HLint 3.0.1. Problem is that these are language pragmas that don't have an associated extension, which we assumed could never happen. I now handle that case in the extension hint, and have added test cases for it.

shayne-fletcher:
Yes, they aren't extensions and thus don't appear in `xFlags`. It turns out there is a dedicated field in `DynFlags` of type `SafeHaskellMode`.
```
-- | The various Safe Haskell modes
data SafeHaskellMode
   = Sf_None          -- ^ inferred unsafe
   | Sf_Unsafe        -- ^ declared and checked
   | Sf_Trustworthy   -- ^ declared and checked
   | Sf_Safe          -- ^ declared and checked
[...]

PR #962, opened 2020-05-02
Rename 'langExts' and 'mkLangExts'

shayne-fletcher:
For your consideration. Previously `langExts` returned any and all strings found in a `LANGUAGE` pragma. Now it weeds out those strings that aren't extensions (in the sense of inhabitants of `Extension`).  The point is that the function is made more faithful to its name. This PR relates to issue https://github.com/ndmitchell/hlint/issues/961.

### DIFF
--- a/src/Apply.hs
+++ b/src/Apply.hs
@@ -71,7 +71,7 @@ applyHintsReal settings hints_ ms = concat $
 removeRequiresExtensionNotes :: ModuleEx -> Idea -> Idea
 removeRequiresExtensionNotes m = \x -> x{ideaNote = filter keep $ ideaNote x}
     where
-        exts = Set.fromList $ concatMap snd $ langExts $ pragmas $ ghcAnnotations m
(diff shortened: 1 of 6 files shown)


ndmitchell:
This seems to regress in the sense that if a language extension isn't "real", we lose the ability to lint any of the extensions next to it. See the test that changed.

shayne-fletcher:
Yes, I agree, that's unfortunate. I wonder if then we might consider renaming the current `langExts` to `potentialLangExts`? (One might then provide `langExts` as being a stricter version (but as we don't have a use-case for it, omit it).

ndmitchell:
Maybe `languagePragmas` would be a better name - they are definitely language pragmas, but as we now know, that doesn't mean they are extensions.

shayne-fletcher:
Yes, I like that.

shayne-fletcher:
Updated accordingly.

PR #964, opened 2020-05-03
Remove mentions of haskell-src-exts

shayne-fletcher:
Dropped mentions of haskell-src-exts.


ndmitchell:
I think that with the number of external contributors keeping an active list of acknowledgements is nigh-on impossible to keep up to date. I may just change that to link to the contributions tab. But thanks for the improvements!

PR #968, opened 2020-05-03
No further processing if parse errors detected

shayne-fletcher:
Do not proceed with any further analysis if a parse tree has been determined to contain errors.


ndmitchell:
Thanks!

Issue #971, opened 2020-05-04
hlint v3.0.4: Parse error for `NoStarIsType` extension


jinwoo:
When hlint v3.0.4 parses an import statement like below, it gives a parse error:

```
{-# LANGUAGE NoStarIsType #-}
...

import GHC.TypeLits(KnownNat, type (+), type (*))
...
```

```
$ hlint Foo.hs 
Foo.hs:126:47: Error: Parse error: on input `*'
Found:
    import GHC.Generics(Generic, Generic1)
    import GHC.Stack
  > import GHC.TypeLits(KnownNat, type (+), type (*))
    import qualified GHC.TypeLits as T
    import Test.QuickCheck
  

1 hint
```

Explicitly passing `-XNoStarIsType` to hlint doesn't help either.

It used to work fine with v2.2.11.

shayne-fletcher:
I'm a little confused by this.

In so far as I know, one can't disable a language pragma at source (happy to be told I'm wrong) so,
```haskell
{-# LANGUAGE NoStarIsType #-}
```
Looks wrong to me. 

Now suppose this program,
```haskell
import GHC.TypeLits(KnownNat, type (+), type (*))

main = undefined
```
Then compiling with `ghc -c -XNoStarIsType`, I get
```
/Users/shaynefletcher/Quux.hs:2:36: error:
    Illegal keyword 'type' (use ExplicitNamespaces to enable)
  |
2 | import GHC.TypeLits(KnownNat, type (+), type (*))
  |                                    ^^^

/Users/shaynefletcher/Quux.hs:2:46: error:
    Illegal keyword 'type' (use ExplicitNamespaces to enable)
  |
2 | import GHC.TypeLits(KnownNat, type (+), type (*))
  |                                              ^^^
```
If I modify the program to read,
```haskell
{-# LANGUAGE ExplicitNamespaces #-}

import GHC.TypeLits(KnownNat, type (+), type (*))

main = undefined
```
and compile with `ghc -XStarIsType -c ~/Quux.hs`, I get
```
/Users/shaynefletcher/Quux.hs:3:47: error: parse error on input ‘*’
  |
3 | import GHC.TypeLits(KnownNat, type (+), type (*))
  |           
```
and with `ghc -c ~/Quux.hs` (or equivalently) `ghc -c -XNoStarIsType ~/Quux.hs`
all is well (that is, I get a successful parse).

[...]

PR #972, opened 2020-05-05
Don't default enable `StarIsType`

shayne-fletcher:
Add `StarIsType` to "bad extensions".  See https://github.com/ndmitchell/hlint/issues/971 for details as to why that might be the right thing to do.

### DIFF
--- a/src/Extension.hs
+++ b/src/Extension.hs
@@ -16,6 +16,7 @@ badExtensions =
   , UnboxedTuples, UnboxedSums -- breaks (#) lens operator
   , QuasiQuotes -- breaks [x| ...], making whitespace free list comps break
   , {- DoRec , -} RecursiveDo -- breaks rec
+  , StarIsType  -- See https://github.com/ndmitchell/hlint/issues/971.
   ]
 
 reallyBadExtensions =


shayne-fletcher:
This may or may not be desired, I'll leave that to you. In any case, it has been identified that there's a logic error in `ghc-lib-parser-ex` such that if an extension is enabled by default but explicitly disabled via pragma, the disabling does not have an effect.  Preparing a patch for that.

shayne-fletcher:
Given `StarIsType` is being removed (https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0143-remove-star-kind.rst), it probably makes sense to do this.

shayne-fletcher:
Patch that fixes the `ghc-lib-parser-ex` problem in https://github.com/ndmitchell/hlint/pull/973.

shayne-fletcher:
On the other hand, as @justinwoo points out, https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-StarIsType indicates that the extension today is enabled by default so perhaps this patch shouldn't be applied. In any case, the ghc-lib-parser-ex patch in https://github.com/ndmitchell/hlint/pull/973 will fix issue https://github.com/ndmitchell/hlint/issues/971.

justinwoo:
cc @jinwoo ^

ndmitchell:
Rejecting this patch - by default GHC enables StarIsType. We should remove this extension when GHC changes the default. Thanks for the great work fixing HLint by other mechanisms!

PR #974, opened 2020-05-05
Respect disabling extensions from the command line

shayne-fletcher:
Support `-XNoFoo`. Fixes https://github.com/ndmitchell/hlint/issues/971.

Tested on this program with `stack run -- -XNoStarIsType ~/Quux.hs` where Quux.hs contains
```haskell
{-# LANGUAGE ExplicitNamespaces #-}

import GHC.TypeLits(KnownNat, type (+), type (*))
```
There were two test failures : one relates to the added test in `Extension.hs` missing the `ExplicitNamespaces` extension, I fixed it incidentally. The other I think is unrelated to this work and probably relates to a work in progress by @ndmitchell

### DIFF
--- a/src/CmdLine.hs
+++ b/src/CmdLine.hs
@@ -230,14 +230,14 @@ cmdHintFiles cmd = do
     where
         ancestors = init . map joinPath . reverse . inits . splitPath
 
-cmdExtensions :: Cmd -> (Maybe Language, [Extension])
+cmdExtensions :: Cmd -> (Maybe Language, ([Extension], [Extension]))
 cmdExtensions = getExtensions . cmdLanguage
 
 
 cmdCpp :: Cmd -> CppFlags
 cmdCpp cmd
     | cmdCppSimple cmd = CppSimple
-    | Cpp `elem` snd (cmdExtensions cmd) = Cpphs defaultCpphsOptions
+    | Cpp `elem` (fst . snd) (cmdExtensions cmd) = Cpphs defaultCpphsOptions
         {boolopts=defaultBoolOptions{hashline=False, stripC89=True, ansi=cmdCppAnsi cmd}
         ,includes = cmdCppInclude cmd
         ,preInclude = cmdCppFile cmd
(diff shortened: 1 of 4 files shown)


shayne-fletcher:
The expected failing test I talk about above is this:
```
TEST FAILURE IN tests/_cpp_2
DIFFER ON LINE: 1
GOT : Warning: Can't find file "Any/File.h" in directories
WANT: No hints
FULL OUTPUT FOR GOT:
Warning: Can't find file "Any/File.h" in directories
	tests
	.
  Asked for by: tests/cpp-none.hs  at line 1 col 1
No hints
```

src/CmdLine.hs
```diff
@@ -314,14 +314,15 @@ getModule path exts x | not (any isSpace x) && all isMod xs = f exts
 getModule _ _ _ = pure Nothing
 
 
-getExtensions :: [String] -> (Maybe Language, [Extension])
-getExtensions args = (lang, foldl f (if null langs then defaultExtensions else []) exts)
+getExtensions :: [String] -> (Maybe Language, ([Extension], [Extension]))
```
shayne-fletcher, line 317: This should be refactored when we have a moment.

shayne-fletcher:
Also, @ndmitchell , sadly technically I believe this patch affects the API.

shayne-fletcher:
Also, since HLint is enabling `ExplicitNamespaces` by default it's sufficient to just `hlint -XNoStarIsType` to get the above program to compile (and done this way, there's no warning about using a restricted extension).

zliu41:
I created #975, and then saw this PR which appears to be related, but it doesn't fix the problem in #975. @shayne-fletcher could you take a look at my PR?

ndmitchell:
Thanks - definitely a step in the right direction! I'll take it over.

@zliu41 - your bug is in a very similar area, but I think different.

ndmitchell:
Thanks Shayne - nice work. Sadly it does change the API, but there's nothing that can be done about that. The CPP failure was a bug, but I've patched that in 011dd628bf87136c4bd126aa63bc61745a3772f8.

PR #975, opened 2020-05-06
Remove filterEnabled from extensionsHint

zliu41:
I'm not sure why `filterEnabled` is needed; removing it doesn't break any existing test, so this PR removes it.

`filterEnabled` causes the following problem:

```haskell
{-# LANGUAGE LambdaCase, MultiWayIf, NoRebindableSyntax #-}
foo = \case True -> 3
```

```
Found:
  {-# LANGUAGE LambdaCase, MultiWayIf, NoRebindableSyntax #-}
Perhaps:
  {-# LANGUAGE LambdaCase #-}
```

The hint incorrectly removes `NoRebindableSyntax`.


shayne-fletcher:
I agree `filterEnabled` should be removed.

ndmitchell:
Less code. More tests. Can't really argue!

PR #984, opened 2020-05-11
Reduce the span of redundant bracket hints

zliu41:
The span of the redundant bracket hints seems longer than necessary:

Screenshot from 2020-05-10 22-23-18

Before:

```
Foo.hs:1:11-22: Suggestion: Redundant bracket
Found:
  y ++ (foo x)
Perhaps:
  y ++ foo x
```

After:

```
Foo.hs:1:16-22: Suggestion: Redundant bracket
Found:
  (foo x)
Perhaps:
  foo x
```

### DIFF
--- a/src/Hint/Bracket.hs
+++ b/src/Hint/Bracket.hs
@@ -8,24 +8,24 @@ list bracket.
 
 <TEST>
 -- expression bracket reduction
-yes = (f x) x -- @Suggestion f x x
+yes = (f x) x -- @Suggestion f x
 no = f (x x)
 yes = (foo) -- foo
 yes = (foo bar) -- @Suggestion foo bar
 yes = foo (bar) -- @Warning bar
-yes = foo ((x x)) -- @Suggestion (x x)
-yes = (f x) ||| y -- @Suggestion f x ||| y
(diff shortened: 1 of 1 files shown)


ndmitchell:
I can see why this is a better span if you are seeing it in the context of the IDE. But when seeing it at the command line, the former message gives enough context to understand why it's a redundant bracket. I'm not sure what to do.

zliu41:
Yeah ideally this would be configurable like verbosity levels. To me even the new span is too long - I'd prefer to see the squiggly line only below the opening bracket - I find long squiggly lines distracting :slightly_smiling_face:

That being said I think it's ok to keep it as it is, because I almost never want to keep redundant brackets (compared to "redundant do" which I often want to keep for consistency reasons, e.g., when writing hspec tests, so it's more important that its span is minimal).

ndmitchell:
I'd be fine to add a `ideaFromIDE` to https://hackage.haskell.org/package/hlint-3.1/docs/Language-Haskell-HLint.html#t:Idea which would usually be set to `ideaFrom` but in some cases can get a better location when its viewed in context. However, not a massive fan of that name, and if this is the only case where it's relevant, doesn't seem worth the complexity.

PR #986, opened 2020-05-11
Remove ticks from Idea

shayne-fletcher:
Remove more ticks.


ndmitchell:
What' a' great' idea'

PR #987, opened 2020-05-11
Remove ticks from Refact

shayne-fletcher:
Remove yet more ticks

### DIFF
--- a/src/Refact.hs
+++ b/src/Refact.hs
@@ -1,9 +1,8 @@
 {-# LANGUAGE LambdaCase #-}
 
 module Refact
-    ( toRefactSrcSpan'
-    , toSS'
-    , toSrcSpan'
+    ( toRefactSrcSpan
+    , toSS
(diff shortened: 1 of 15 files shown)


src/Refact.hs
```diff
@@ -29,13 +28,10 @@ toRefactSrcSpan' = \case
     GHC.UnhelpfulSpan _ ->
         R.SrcSpan (-1) (-1) (-1) (-1)
 
--- | Don't crash in case ghc gives us a \"fake\" span,
+-- | Dont crash in case ghc gives us a \"fake\" span,
```
ndmitchell, line 31: Do we need to remove 's from `Don't`?

shayne-fletcher: oops

ndmitchell:
T'h'a'n'k's

shayne-fletcher:
You're welcome
      ^
Plenty more to do 😉

PR #997, opened 2020-05-14
Remove ticks from RdrName

shayne-fletcher:
Remove ticks from `RdrName` and take the chance to reduce some repetition by consistent use of `rdrNameStr` and newly added `occNameStr`.

### DIFF
--- a/src/Config/Compute.hs
+++ b/src/Config/Compute.hs
@@ -72,7 +72,7 @@ findExp name vs bod = [SettingMatchExp $
         rhs = apps' $ map noLoc $ HsVar noExtField (noLoc name) : map snd rep
 
         rep = zip vs $ map (mkVar . pure) ['a'..]
-        f (HsVar _ x) | Just y <- lookup (occNameString $ occName $ unLoc x) rep = y
+        f (HsVar _ x) | Just y <- lookup (rdrNameStr x) rep = y
         f (OpApp _ x dol y) | isDol dol = HsApp noExtField x $ noLoc $ HsPar noExtField y
         f x = x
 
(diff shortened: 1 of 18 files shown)


src/Config/Compute.hs
```diff
@@ -72,7 +72,7 @@ findExp name vs bod = [SettingMatchExp $
         rhs = apps' $ map noLoc $ HsVar noExtField (noLoc name) : map snd rep
 
         rep = zip vs $ map (mkVar . pure) ['a'..]
-        f (HsVar _ x) | Just y <- lookup (occNameString $ occName $ unLoc x) rep = y
+        f (HsVar _ x) | Just y <- lookup (rdrNameStr x) rep = y
```
ndmitchell, line 75: Could we get some hlint rules for these in the HLint local hints thing? Otherwise I'll reintroduce them :)

src/Config/Haskell.hs
```diff
@@ -33,8 +32,8 @@ readPragma :: AnnDecl GhcPs -> Maybe Classify
 readPragma (HsAnnotation _ _ provenance expr) = f expr
     where
         name = case provenance of
-            ValueAnnProvenance (L _ x) -> occNameString $ occName x
-            TypeAnnProvenance (L _ x) -> occNameString $ occName x
+            ValueAnnProvenance (L _ x) -> occNameStr x
```
ndmitchell, line 35: Ditto for an HLint hint.

ndmitchell:
Merging, but adding some HLint rules would be really useful so people use the right utilities. We have a few custom hints in https://github.com/ndmitchell/hlint/blob/master/.hlint.yaml, but I think most of them are HSE, so totally useless.

PR #1006, opened 2020-05-16
Remove ticks from Match

shayne-fletcher:
Pacman all the ticks!


ndmitchell:
Can there be any more ticks left?''

shayne-fletcher:
There's a few pockets yet. Not so much.

PR #1010, opened 2020-05-18
Update ghc-lib-parser, ghc-lib-parser-ex

shayne-fletcher:
Not essential (hlint would pick up the ghc-lib-parser change anyway and the ghc-lib-parser-ex change doesn't introduce anything new) but one of those helps me keep everything sane updates.

## 8.10.0.10 released 2020-05-18
- Upgrade to `ghc-lib-parser-8.10.1.20200518`


ndmitchell:
Important to keep you sane!

PR #1011, opened 2020-05-19
Add new hint "Newtype field name prefixed"

EncodePanda:
New hint, turned off by default, that checks if field name in `newtype` constructor is prefixed with `un` followed by the newtype Name.

Example below:
```
➜  bat Foo.hs 
───────┬───────────────────────────────────────────────────────────────────
       │ File: Foo.hs
───────┼───────────────────────────────────────────────────────────────────
   1   │ module Foo where
   2   │ 
   3   │ 
   4   │ newtype Foo = Foo { bar :: Bar }
───────┴───────────────────────────────────────────────────────────────────
➜  hlint Foo.hs 
No hints
➜  touch .hlint.yaml
➜  echo '- warn: {name: Newtype field name prefixed}' > .hlint.yaml
➜  bat .hlint.yaml 
───────┬───────────────────────────────────────────────────────────────────
       │ File: .hlint.yaml
───────┼───────────────────────────────────────────────────────────────────
   1   │ - warn: {name: Newtype field name prefixed}
───────┴───────────────────────────────────────────────────────────────────
➜  hlint Foo.hs                                                    
Foo.hs:4:1-32: Warning: Newtype field name prefixed
Found:
  newtype Foo = Foo {bar :: Bar}
Perhaps:
  newtype Foo = Foo {unFoo :: Bar}
Note: Rename field to unFoo

1 hint

```

### DIFF
--- a/src/Hint/NewType.hs
+++ b/src/Hint/NewType.hs
@@ -28,6 +28,12 @@ newtype Foo = Foo Int deriving (Show, Eq) --
 newtype Foo = Foo { unFoo :: Int } deriving (Show, Eq) --
 newtype Foo = Foo Int deriving stock Show
 </TEST>
+
+   Suggest constructor field name (of a newtype) to be 'un' followed by newtype name
+
+<TEST>
+newtype Foo = Foo { i :: Int } -- newtype Foo = Foo { unFoo :: Int } @NoRefactor
+</TEST>
 -}
 module Hint.NewType (newtypeHint) where
 
--- a/src/Hint/NewType.hs
+++ b/src/Hint/NewType.hs
@@ -25,22 +25,28 @@ data A = A Int#  @NoRefactor
 {-# LANGUAGE UnboxedTuples #-}; data WithAnn x = WithAnn {getWithAnn :: (# Ann, x #)}
 data A = A () -- newtype A = A () @NoRefactor
 newtype Foo = Foo Int deriving (Show, Eq) --
-newtype Foo = Foo { getFoo :: Int } deriving (Show, Eq) --
+newtype Foo = Foo { unFoo :: Int } deriving (Show, Eq) --
 newtype Foo = Foo Int deriving stock Show
 </TEST>
 -}
 module Hint.NewType (newtypeHint) where
 
-import Hint.Type (Idea, DeclHint, Note(DecreasesLaziness), ideaNote, ignoreNoSuggestion, suggestN)
+import Hint.Type (Idea, DeclHint, Note(..), ignore, ideaNote, ignoreNoSuggestion, suggestN)
 
 import Data.List (isSuffixOf)
+import Data.Generics.Uniplate.Operations
 import GHC.Hs.Decls
 import GHC.Hs
-import Outputable
+import Language.Haskell.GhclibParserEx.GHC.Utils.Outputable
+import Outputable hiding ((<>))
+import OccName
 import SrcLoc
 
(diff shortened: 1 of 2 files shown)


josephcsible:
Can you make this take a configurable list of acceptable prefixes, instead of hardcoding just `un`? In particular, a lot of idiomatic Haskell uses `get` and `run` as well as `un`.

EncodePanda:
@josephcsible Sure, I was hoping to extend it with an additional set of features (internally we use `un` for all newtypes except for wrapping monad transformers where we use `run`).

But I was unaware you can parameterize hints. Could you show me an example of how you can provide value to a hint (I presume you do that in `.hlint.yaml`)

josephcsible:
As I actually think about how to do that, it looks like support for parameters like that isn't obvious. I'll defer to @ndmitchell for the answer to that.

EncodePanda:
@ndmitchell `All checks have passed` :)

googleson78:
+1 for "configurability" - e.g. base predominantly uses `get*` (`getSum`/`getProduct` etc)

EncodePanda:
> +1 for "configurability" - e.g. base predominantly uses `get*` (`getSum`/`getProduct` etc)

@googleson78 Sure, this can get in as a separate contribution, right?

src/Hint/NewType.hs
```diff
@@ -138,3 +150,41 @@ dropConsBang x = x
 isUnboxedTuple :: HsType GhcPs -> Bool
 isUnboxedTuple (HsTupleTy _ HsUnboxedTuple _) = True
 isUnboxedTuple _ = False
+
+data WarnNewTypeUn = WarnNewTypeUn
```
ndmitchell, line 154: Why not a Maybe? Or even desugar the whole thing away, and have a warnNewTypeUn :: LHsDecl GhcPs -> String -> Idea`?

src/Hint/NewType.hs
```diff
@@ -138,3 +150,41 @@ dropConsBang x = x
...
+    , unNewFieldName :: String
+    }
+    | NoWarn
+
+newtypeUn :: LHsDecl GhcPs -> [Idea]
+newtypeUn old
+  | WarnNewTypeUn{unNewDecl, unNewFieldName} <- newTypeWithoutUn old
+  = let
+      note fieldName = [Note $ "Rename field to " <> unNewFieldName]
+      hintName = "Newtype field name prefixed"
```
ndmitchell, line 165: From this name, I wouldn't have guessed what hint this referred to. Can we come upw ith a better name?

src/Hint/NewType.hs
```diff
@@ -138,3 +150,41 @@ dropConsBang x = x
...
+    { unNewDecl      :: LHsDecl GhcPs
+    , unNewFieldName :: String
+    }
+    | NoWarn
+
+newtypeUn :: LHsDecl GhcPs -> [Idea]
+newtypeUn old
+  | WarnNewTypeUn{unNewDecl, unNewFieldName} <- newTypeWithoutUn old
+  = let
+      note fieldName = [Note $ "Rename field to " <> unNewFieldName]
```
ndmitchell, line 164: What's the purpose of this note? The node usually refers to how the hint isn't quite correct, not just a description of what the hint did. My guess is this hint doesn't need a name.

ndmitchell:
Please reopen a fresh or discuss on the ticket if people still are interested

PR #1012, opened 2020-05-21
Added package, group, rules for CodeWorld

jvoigtlaender:
I'm not sure how "widely used" a package has to be to warrant its own set of hlint rules. But I found some rules useful in providing style feedback to students programming on/with the CodeWorld website (https://github.com/google/codeworld, https://code.world/haskell#), which is based on the codeworld-api package (https://hackage.haskell.org/package/codeworld-api). Maybe they would be useful to have in hlint for others as well.

So here they are, in an off-by-default rules group. There are certainly others (since the CodeWorld primitives satisfy many algebraic properties) that could be added in due course.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -50,6 +50,11 @@
     - import Data.Attoparsec.Text
     - import Data.Attoparsec.ByteString
 
+- package:
+    name: codeworld-api
+    modules:
+    - import CodeWorld
+
 - group:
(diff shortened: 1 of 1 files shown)


ndmitchell:
The bar isn't that high, but I have a soft spot for CodeWorld, since I'm a big fan of the initiative. Good luck and happy to merge this code.

PR #1013, opened 2020-05-23
Unqualify Control.Applicative.<$>

zliu41:
`<$>` has been in Prelude since base 4.8, so the "Control.Applicative." can be safely removed.


ndmitchell:
Thanks!

PR #1016, opened 2020-05-25
Check scopes of restricted functions

### LINKED ISSUE #861: Can't restrict head without restricting NE.head
If I restrict the `head` function, my intention is to restrict only the bad one from `Prelude`. However, if I do `import qualified Data.List.NonEmpty as NE` and then use `NE.head`, hlint flags that one too.

zliu41:
Fixes #861. This allows qualified restricted functions such as `{name: Data.List.nub, within: []}`. Previously this doesn't match anything, including `nub` and `Data.List.nub`.

If this is merged, a release would be appreciated. We would like to ban `Data.Map.fromList` but not `Ddata.Set.fromList`, and this patch allows us to do so.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -33,7 +33,10 @@
   - {name: unsafeInterleaveIO, within: Parallel}
   - {name: unsafePerformIO, within: [Util.exitMessageImpure, Test.Util.ref, Timing]}
   - {name: unsafeCoerce, within: [Util.gzip, GHC.Util.Refact.Utils]}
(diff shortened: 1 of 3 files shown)


ndmitchell:
Thanks! This code was a bit of a bombshell before, and definitely looks like an improvement now.

ndmitchell:
I'll make a release once the CI confirms everything is good after the merge.

ndmitchell:
3.1.3 now out.

PR #1017, opened 2020-05-25
Get rid of a use of head

josephcsible:


ndmitchell:
Thanks for removing `head`. That bit looks good.

For the use of `foldr`, that seems to confuse the code rather than make it clear. But worse, I think the first should be using a strict accumulator for the length, and the second probably strict in Dupe, so they both likely have space leaks, and moving to the foldr makes them harder to get rid of. I'd stick with explicit recursion, but if you want to avoid the space leaks as well, that would be great.

ndmitchell:
You unluckily rebased just as I had broken the CI, so I've merged it anyway - the one break in the CI was my fault. Thanks for the improvements!

josephcsible:
I just realized that when I squashed the commits, I kept the wrong commit message. Whoops, too late to fix now though.

ndmitchell:
Never mind, I'm not that fussy about the history - it's close enough!

PR #1019, opened 2020-05-29
more rules suggesting even/odd

jvoigtlaender:
There are already these rules: https://github.com/ndmitchell/hlint/blob/f1b8d330cac9683dd5337a38e730a5f9915a97bf/data/hlint.yaml#L616-L619
This pull request adds the analogous rules for `mod` in place of `rem`.


ndmitchell:
Thanks! Given I'd have used mod rather than REM, this looks great.

PR #1020, opened 2020-05-29
more rules for CodeWorld

jvoigtlaender:
A few more rule suggestions for `codeworld-api`.

Some of these were generated automatically via hlint from the CodeWorld sources.


ndmitchell:
Thanks!

PR #1021, opened 2020-05-29
in teaching group, suggest use of map for very simple list comprehension?

jvoigtlaender:
Actually, I was a bit surprised because I thought to remember that in the past hlint did on its own suggest replacing something like `[f x | x <- xs]` by `map f xs`. But maybe I was misremembering, or missing anything?


zliu41:
You should replace `xs` with a single letter variable - multiple letter variables are not unified.

jvoigtlaender:
Thanks, fixed.

ndmitchell:
Thanks! I think we used to suggest that, but it wasn't always simpler, if you write convoluted things with a list comp. I might not be averse to adding it back under certain conditions (e.g. the `f` is an atomic expression, so you don't need additional brackets in the first argument to map).

Issue #1001, opened 2020-05-16
Fix/rewrite Scope.hs


zliu41:
The way Scope.hs behaves is fairly erratic. For example:

```haskell
-- (correctly) suggests mapM
import qualified Control.Monad
x = flip Control.Monad.forM
```

```haskell
-- no hint
-- Should suggest: Control.Monad.mapM or Prelude.mapM
import qualified Control.Monad
import qualified Prelude
x = flip Control.Monad.forM
```

```haskell
-- suggests mapM
-- Should suggest: Control.Monad.mapM or Prelude.mapM
import qualified Control.Monad
import qualified Prelude
import Data.Int
x = flip Control.Monad.forM
```

I guess a name should be assumed, if there's no other source of information, to come from Prelude. With this assumption, in the second case it should be able to suggest `Prelude.mapM`, and in the third case it should know that `Data.Int` is irrelevant.

ndmitchell:
Agreed! Scope needs a significant overhaul and performance improvements too. There are increasingly more tickets which are scope bugs. I'll create a label for them.

ndmitchell:
Plan of attack is probably first to figure out what the semantics of the "perfect" scope module are, then how to implement that efficiently, then finally implement it.

ndmitchell:
Here's an initial outline of the Scope problem. Throughout, I use name to mean things like `foo` and qname to mean things like `Foo.bar`.

## Hint Scope

* A hint scope is responsible for mapping a name or qname to many possible qnames it might represent. For example, `fmap` in a hint might be `Prelude.fmap` or `Data.Functor.fmap`.
* For some names, e.g. `map`, we may choose to map them to `Prelude.fmap` and `Prelude.map`, on the basis that any `fmap` hint also applies to `map` (not sure yet).
* A hint scope is constructed from information in a `.yaml` file, which can be anything we want, but a sequence of imports plus additional information is likely convenient.

## Module Scope

* Like a hint scope, a module scope is responsible for mapping a name or qname to many possible qnames it might represent.
* A module scope is constructed from information in a `.hs` file, notably the list of imports and the presence/absence of the `ImplicitPrelude` extension.

## Operations

* Given a name/qname in a hint, and a name/qname in a module, we need to know if these two refer to the same identifier. In principle that means computing all possible qnames on both sides and seeing if they intersect.
[...]

PR #1023, opened 2020-05-31
Replace an occurrence of disjoint with disjointOrd

zliu41:
This has a noticeable speedup on `cabal/cabal-install/Distribution/Client/Setup.hs` (one of the largest .hs files in the repo):

```
(hlint Setup.hs 20 times)
Before:
real    0m42.864s
user    0m42.619s
sys     0m0.269s

After, 12% speedup:
real    0m37.755s
user    0m37.516s
sys     0m0.257s
```

The speedup quickly diminishes when the file is smaller, since `disjoint` and `disjointOrd` have similar runtime if one of the lists is small (`disjoint` could even be a bit faster).

```
(hlint the whole cabal repo 20 times)
Before:
real    24m35.506s
user    24m28.125s
sys     0m8.527s

After, 4% speedup:
real    23m37.180s
user    23m29.249s
sys     0m8.867s
```

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -62,7 +62,7 @@ library
         yaml >= 0.5.0,
         uniplate >= 1.5,
         ansi-terminal >= 0.6.2,
-        extra >= 1.7.1,
+        extra >= 1.7.3,
         refact >= 0.3,
         aeson >= 1.1.2.0,
         filepattern >= 0.1.1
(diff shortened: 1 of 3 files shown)


ndmitchell:
Awesome! I think the big win will come from redesigning scopes in #1001, but a very welcome improvement.

Issue #504, opened 2018-06-26
Test and fix refactorings


ndmitchell:
To get the refactorings working properly it is necessary for:

* Neil to actually start using it (so it gets documented)
* Add automated tests to all operations (so they are correct)

This ticket is expected to be the overarching ticket for many refactoring tickets.

ndmitchell:
@zliu41 - it feels like this is mostly done now? I might not be using it, but I think I will be when support comes to my IDE, and you are certainly cleaning it up pretty well with tests. Time to close this ticket, or something else you think would be good to include?

zliu41:
@ndmitchell Looks like the reasons for the remaining @NoRefactors are
1. not supported (e.g., reduce duplication, eta reduce)
2. apply-refact bugs
3. apply-refact requires GHC extensions
4. scoping issues (e.g., should suggest `CM.forM` but suggest `forM`)
5. other hlint bugs (e.g., `f (Just a) = \a -> a + a`)

I think 1 should not prevent this ticket from being closed, but it would be nice to document which hints are not supported. 2 to 5 should probably be fixed before closing this ticket. Among these, 3 and 5 should be easy; 4 is tricky; 2 is not difficult but depends on apply-refactor maintainers [being responsive](https://github.com/mpickering/apply-refact/pull/51). (maybe you can have them add you  as a maintiner?)

ndmitchell:
Thanks for the summary. On the basis of that I revised the README at https://github.com/ndmitchell/hlint/blob/master/README.md#automatically-applying-hints.

1. I think 1 is fine, and is a documented invariant of HLint - not everything can get automatic replacements.
2. The apply-refact bugs are problematic, and I would actually suggest they add _you_ as the maintainer since you are way more active in this area than I am and have shown great taste and development in your patches to HLint. I'll make some enquiries.
3. I see you've already got this one in hand.
5. We can fix these up as they happen.

So it sounds like we're well on our way to getting there.

ndmitchell:
I think the refactorings are sufficiently in control that there is no longer any need for an overarching ticket. We have testing and it works very well. Thanks for all your hard work @zliu41

PR #1027, opened 2020-06-01
Pass enabled and disabled extensions to apply-refact

zliu41:
Related: #504

For testing, enabled = defaultExtensions, disabled = [].

This uncovered a bug in apply-refact, see https://github.com/mpickering/apply-refact/issues/58

### DIFF
--- a/src/Extension.hs
+++ b/src/Extension.hs
@@ -24,6 +24,9 @@ reallyBadExtensions =
   {- , XmlSyntax , RegularPatterns -} -- steals a-b and < operators
   , AlternativeLayoutRule -- Does not play well with 'MultiWayIf'
   , NegativeLiterals -- Was not enabled by HSE and enabling breaks tests.
+  , StarIsType -- conflicts with TypeOperators. StarIsType is currently enabled by default,
+               -- so adding it here has no effect except avoiding passing it to apply-refact.
+               -- See https://github.com/mpickering/apply-refact/issues/58
   ]
 
 -- | Extensions we turn on by default when parsing. Aim to parse as
--- a/src/HLint.hs
+++ b/src/HLint.hs
@@ -9,7 +9,6 @@ import Control.Monad.Extra
 import Control.Exception.Extra
 import Control.Concurrent.Extra
 import System.Console.CmdArgs.Verbosity
-import GHC.LanguageExtensions.Type
 import GHC.Util.DynFlags
 import Data.List.Extra
 import GHC.Conc
--- a/src/HLint.hs
+++ b/src/HLint.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# OPTIONS_GHC -Wno-incomplete-patterns #-}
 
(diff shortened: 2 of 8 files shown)


src/HLint.hs
```diff
@@ -184,8 +186,8 @@ getIdeas cmd@CmdMain{..} settings = do
 
 -- #746: run refactor even if no hint, which ensures consistent output
 -- whether there are hints or not.
-handleRefactoring :: [Idea] -> [String] -> Cmd -> IO ()
-handleRefactoring ideas files cmd@CmdMain{..} =
+handleRefactoring :: [Idea] -> [String] -> Cmd -> [Extension] -> [Extension] -> IO ()
+handleRefactoring ideas files cmd@CmdMain{..} ys ns =
```
ndmitchell, line 190: Why not have handleRefactoring just do:

```
let ParseFlags{enabledExtensions, disabledExtensions} = cmdParseFlags cmd
```

Everything else just passes around the `CmdMain` flags and generates the parse flags from it when they need it. I appreciate that results in calling `cmdParseFlags` twice (more expensive), but that seems better than relying on the invariant that the flags match the command.

src/HLint.hs
```diff
@@ -163,18 +166,17 @@ runHints args settings cmd@CmdMain{..} = do
...
             usecolour <- cmdUseColour cmd
             showItem <- if usecolour then showANSI else pure show
             mapM_ (outStrLn . showItem) ideas
             handleReporting ideas cmd
         pure ideas
 
-getIdeas :: Cmd -> [Setting] -> IO [Idea]
-getIdeas cmd@CmdMain{..} settings = do
+getIdeas :: Cmd -> ParseFlags -> [Setting] -> IO [Idea]
+getIdeas CmdMain{..} flags settings = do
```
ndmitchell, line 178: I wouldn't bother passing flags here. Recomputing them is nice and easy.

src/Refact.hs
```diff
@@ -53,9 +54,10 @@ refactorPath rpath = do
                        , "<https://github.com/mpickering/apply-refact>"
                        ]
 
-runRefactoring :: FilePath -> FilePath -> FilePath -> String -> IO ExitCode
-runRefactoring rpath fin hints opts =  do
+runRefactoring :: FilePath -> FilePath -> FilePath -> [Extension] -> [Extension] -> String -> IO ExitCode
+runRefactoring rpath fin hints ys ns opts =  do
```
ndmitchell, line 58: Perhaps `enabled disabled` as the names? I appreciate that `ys`/`ns` was meant to signify that, but it wasn't immediately obvious to me.

src/Extension.hs
```diff
@@ -24,6 +24,9 @@ reallyBadExtensions =
   {- , XmlSyntax , RegularPatterns -} -- steals a-b and < operators
   , AlternativeLayoutRule -- Does not play well with 'MultiWayIf'
   , NegativeLiterals -- Was not enabled by HSE and enabling breaks tests.
+  , StarIsType -- conflicts with TypeOperators. StarIsType is currently enabled by default,
+               -- so adding it here has no effect except avoiding passing it to apply-refact.
+               -- See https://github.com/mpickering/apply-refact/issues/58
```
ndmitchell, line 29: Great detective work!

ndmitchell:
Amazing work!

zliu41:
Thanks!

PR #1028, opened 2020-06-02
Add/modify sequenceA hints

zliu41:
- When a rule LHS contains `sequence` or `sequence_`, add the corresponding `sequenceA` or `sequenceA_` version
- ~When a rule RHS contains `sequence` or `sequence_`, suggest `sequenceA` or `sequenceA_` instead~
- Removed a duplicate hint.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -241,7 +241,6 @@
 
     - warn: {lhs: sequenceA (map f x), rhs: traverse f x}
     - warn: {lhs: sequenceA (fmap f x), rhs: traverse f x}
-    - warn: {lhs: sequence (fmap f x), rhs: traverse f x}
     - warn: {lhs: sequenceA_ (map f x), rhs: traverse_ f x}
     - warn: {lhs: sequenceA_ (fmap f x), rhs: traverse_ f x}
     - warn: {lhs: foldMap id, rhs: fold}
@@ -264,7 +263,7 @@
 
     # FOLDS
 
-    - warn: {lhs: foldr  (>>) (return ()), rhs: sequence_}
+    - warn: {lhs: foldr  (>>) (return ()), rhs: sequenceA_}
     - warn: {lhs: foldr  (&&) True, rhs: and}
     - warn: {lhs: foldl  (&&) True, rhs: and, note: IncreasesLaziness}
     - warn: {lhs: foldr1 (&&) , rhs: and, note: "RemovesError on `[]`"}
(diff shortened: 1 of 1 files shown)


data/hlint.yaml
```diff
@@ -485,14 +484,18 @@
...
     - warn: {lhs: mapM f (replicate n x), rhs: Control.Monad.replicateM n (f x)}
     - warn: {lhs: mapM_ f (replicate n x), rhs: Control.Monad.replicateM_ n (f x)}
     - warn: {lhs: mapM f (map g x), rhs: mapM (f . g) x, name: Fuse mapM/map}
     - warn: {lhs: mapM_ f (map g x), rhs: mapM_ (f . g) x, name: Fuse mapM_/map}
     - warn: {lhs: traverse f (map g x), rhs: traverse (f . g) x, name: Fuse traverse/map}
     - warn: {lhs: traverse_ f (map g x), rhs: traverse_ (f . g) x, name: Fuse traverse_/map}
-    - warn: {lhs: mapM id, rhs: sequence}
-    - warn: {lhs: mapM_ id, rhs: sequence_}
+    - warn: {lhs: mapM id, rhs: sequenceA}
+    - warn: {lhs: mapM_ id, rhs: sequenceA_}
```
ndmitchell, line 498: mapM has a Monad constraint, so does sequence, but sequenceA doesn't - so the original replacement seems right.

data/hlint.yaml
```diff
@@ -264,7 +263,7 @@
 
     # FOLDS
 
-    - warn: {lhs: foldr  (>>) (return ()), rhs: sequence_}
+    - warn: {lhs: foldr  (>>) (return ()), rhs: sequenceA_}
```
ndmitchell, line 266: `>>` induces a Monad constraint, so sequence_ seems the more correct thing to suggest.

data/hlint.yaml
```diff
@@ -241,7 +241,6 @@
 
     - warn: {lhs: sequenceA (map f x), rhs: traverse f x}
     - warn: {lhs: sequenceA (fmap f x), rhs: traverse f x}
-    - warn: {lhs: sequence (fmap f x), rhs: traverse f x}
```
ndmitchell, line 244: Good catch

data/hlint.yaml
zliu41: That shouldn't be a problem since Applicative is less restrictive. The idea is to suggest the more general variant.

zliu41: At work we have rules like `- warn: {lhs: sequence, rhs: sequenceA}` to ban `sequence` (and also `mapM`, `forM`, `return` and so on). My guess is most people prefer their `Applicative` counterparts.

ndmitchell: I tend to reach for sequence still, since it's a shorter name, and `sequenceA` is visually ugly. I think it's probably good to have a group of rules (off-by-default) that do the `sequence => sequenceA`, that we might turn on by default one day, but not sure we want to be changing the type people use just yet (in contrast, if they use `traverse id`, that should definitely be `sequenceA`).

Currently I think return => pure is about as brave as I'm ready to be in HLint, and even then, I'm treading slowly (it's in the future group for now)

zliu41: Ok, makes sense! I've reverted those changes.

ndmitchell:
Thanks!

PR #1029, opened 2020-06-02
Fix redundant lambda refactoring when there are generated wildcard patterns

zliu41:
Fixes refactoring for `f (Just a) = \a -> a + a`. Previously refactoring output is `f (Just a) a = a + a`, now `f (Just _) a = a + a`.

The approach is to avoid replacing a pattern if it contains any wildcard, in this case `Just _`.

Care must be taken to not use any substitution variable that occurs in patterns with wildcards. For example, if the pattern is `Foo a b _`, then `a` and `b` cannot be used as substitution variables, and must be removed from `['a'..'z']`. A test case has been added to cover this.

### DIFF
--- a/src/Hint/Lambda.hs
+++ b/src/Hint/Lambda.hs
@@ -290,7 +290,7 @@ mkOrigPats pats = (zipWith munge subtsVars pats', subtsVars)
     -- Returns (chars in the pattern if the pattern contains wildcards, (whether the pattern contains wildcards, the pattern))
     f :: LPat GhcPs -> (Set Char, (Bool, LPat GhcPs))
     f p
-      | any isWildPat (universeBi p) =
+      | any isWildPat (universe p) =
           let used = Set.fromList [c | (L _ (VarPat _ (rdrNameStr -> [c]))) <- universe p]
            in (used, (True, p))
       | otherwise = (mempty, (False, p))
--- a/src/Hint/Lambda.hs
+++ b/src/Hint/Lambda.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE ViewPatterns, PatternGuards #-}
+{-# LANGUAGE LambdaCase, PatternGuards, ViewPatterns #-}
 
 {-
     Concept:
@@ -22,7 +22,8 @@
 <TEST>
 f a = \x -> x + x -- f a x = x + x
 f a = \a -> a + a -- f _ a = a + a
-f (Just a) = \a -> a + a -- f (Just _) a = a + a @NoRefactor
+f (Just a) = \a -> a + a -- f (Just _) a = a + a
+f (Foo a b c) = \c -> c + c -- f (Foo a b _) c = c + c
 f a = \x -> x + x where _ = test
 f (test -> a) = \x -> x + x
 f = \x -> x + x -- f x = x + x
@@ -100,6 +101,7 @@ module Hint.Lambda(lambdaHint) where
 import Hint.Type (DeclHint, Idea, Note(RequiresExtension), suggest, warn, toSS, suggestN, ideaNote)
 import Util
 import Data.List.Extra
+import Data.Set (Set)
 import qualified Data.Set as Set
 import Refact.Types hiding (RType(Match))
 import Data.Generics.Uniplate.DataOnly (universe, universeBi, transformBi)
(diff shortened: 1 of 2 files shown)


src/Hint/Lambda.hs
```diff
@@ -198,9 +202,11 @@ lambdaExp p o@(SimpleLambda origPats origBody)
     where
       (pats, body) = fromLambda o
 
-      template = unsafePrettyPrint $ lambda (zipWith munge ['a'..'z'] pats) varBody
+      (template, subtsVars) =
```
ndmitchell, line 205: Why not the simpler:

```
let (origPats, vars) = mkOrigPats pats
    template = unsafePrettyPrint $ lambda origPats varBody
```

src/Hint/Lambda.hs
```diff
@@ -149,11 +151,13 @@ lambdaDecl
 
           (finalpats, body) = fromLambda . lambda pats $ origBody
           (pats2, bod2) = etaReduce pats origBody
-          template fps = unsafePrettyPrint $ reform (zipWith munge ['a'..'z'] fps) varBody
-          subts fps b = ("body", toSS b) : zipWith (\x y -> ([x],y)) ['a'..'z'] (map toSS fps)
+          template fps =
```
ndmitchell, line 154: Why make template a function and then only apply it at finalpats? If it's just going to be finalpats, why not `(origPats, substsVars) = mkOrigPats finalpats` at the top level and avoid the nested lets?

src/Hint/Lambda.hs
```diff
@@ -149,11 +151,13 @@ lambdaDecl
...
+            let (origPats, subtsVars) = mkOrigPats fps
+             in (unsafePrettyPrint (reform origPats varBody), subtsVars)
+          subts fps b = ("body", toSS b) : zipWith (\x y -> ([x],y)) subtsVars (map toSS fps)
           s1 = subts finalpats body
           --s2 = subts pats2 bod2
-          t1 = template finalpats
+          (t1, subtsVars) = template finalpats
```
ndmitchell, line 160: Why not inline template here?

src/Hint/Lambda.hs
```diff
@@ -149,11 +151,13 @@ lambdaDecl
...
           (finalpats, body) = fromLambda . lambda pats $ origBody
           (pats2, bod2) = etaReduce pats origBody
-          template fps = unsafePrettyPrint $ reform (zipWith munge ['a'..'z'] fps) varBody
-          subts fps b = ("body", toSS b) : zipWith (\x y -> ([x],y)) ['a'..'z'] (map toSS fps)
+          template fps =
+            let (origPats, subtsVars) = mkOrigPats fps
+             in (unsafePrettyPrint (reform origPats varBody), subtsVars)
+          subts fps b = ("body", toSS b) : zipWith (\x y -> ([x],y)) subtsVars (map toSS fps)
           s1 = subts finalpats body
           --s2 = subts pats2 bod2
```
ndmitchell, line 159: Delete these comments if they don't mean anything (I imagine they were from when we had both HSE and GHC?)

src/Hint/Lambda.hs
```diff
@@ -268,7 +274,31 @@ fromLambda (SimpleLambda ps1 (fromLambda . fromParen -> (ps2,x))) = (transformBi
...
--- | Replaces all non-wildcard patterns with a variable pattern with the given identifier.
-munge :: Char -> LPat GhcPs -> LPat GhcPs
-munge ident p@(L _ (WildPat _)) = p
-munge ident (L ploc p) = L ploc (VarPat noExtField (L ploc $ mkRdrUnqual $ mkVarOcc [ident]))
+-- | For each pattern, if it does not contain wildcards, replace it with a variable pattern.
+--
+-- The second component of the result is a list of substitution variables, which is ['a'..'z'],
+-- excluding variables that occur in patterns with wildcards. For example, if there is a pattern
+-- 'Foo a b _', then 'a' and 'b' are removed.
+mkOrigPats :: [LPat GhcPs] -> ([LPat GhcPs], String)
```
ndmitchell, line 282: Better to use `[Char]` when that is the moral underlying type, rather than String.

```suggestion
mkOrigPats :: [LPat GhcPs] -> ([LPat GhcPs], [Char])
```

src/Hint/Lambda.hs
```diff
@@ -268,7 +274,31 @@ fromLambda (SimpleLambda ps1 (fromLambda . fromParen -> (ps2,x))) = (transformBi
...
+mkOrigPats pats = (zipWith munge subtsVars pats', subtsVars)
+  where
+    (Set.unions -> used, pats') = unzip (fmap f pats)
```
ndmitchell, line 285: |Is this fmap just a map? If so, use a map?

src/Hint/Lambda.hs
zliu41: They are supposed to be un-commented when eta-reduce refactoring is supported. But there's some challenge in refactoring eta-reduce and it's probably not happening soon, so I'll delete them.

src/Hint/Lambda.hs
zliu41: I thought this would cause the CI to fail since it triggers the "Use String" hint - I'll see if CI cares or not.

zliu41: It does - I disabled it in .hlint.yaml.

ndmitchell:
Thanks! I think that the use String for [Char] warning isn't universally true, so I'm going to make that ignore by default for everyone.

ndmitchell:
Now removed the exclusion and the hint (ignore by default).

PR #1032, opened 2020-06-03
Fix refactoring for "Use :"

zliu41:
Given 
```haskell
yes = [if a then b else c] ++ xs
```

Refactoring output was
```haskell
if a then b else c : xs
```

Now
```haskell
(if a then b else c) : xs
```


ndmitchell:
Thanks!

PR #1035, opened 2020-06-06
Fix refactoring for "Use fewer imports" with three or more imports

zliu41:
Fixes https://github.com/mpickering/apply-refact/issues/52.

As far as I can tell there's no more known HLint refactoring bug. There are still a few issues with apply-refact that need to be addressed.


ndmitchell:
Wow! Great work!

ndmitchell:
Let me know when a release of HLint would be good, and we can synchronise with apply-refact, and advertise it as refactoring come of age.

zliu41:
I should be ready to release a new version of apply-refact in a week or so. There are a few more bugs to fix, and I'd also like to support some GHC versions that are not GHC 8.10 (at the very least 8.8). I suppose not many people are on GHC 8.10 yet. I don't think there's a way to migrate it to ghc-lib because ghc-exactprint hasn't been migrated. So adding a bunch of CPP for now.

ndmitchell:
Migrating ghc-exactprint to ghc-lib should be a case of recompiling and mostly changing some .cabal flag magic. The level of compatibility is nearly complete. That said, the .cabal flag magic is a bit grim.

PR #1036, opened 2020-06-06
Disable refactoring for "Use camelCase"

zliu41:
Related: https://github.com/mpickering/apply-refact/issues/39

Currently refactoring doesn't work at all for "Use camelCase". It seems tricky to fix - there appears to be bugs in both HLint and apply-refact related to this.

Besides, the current approach relies on `unsafePrettyPrint`  to print the entire `LHsDecl`, which can be quite large, e.g., the entire function definition. This is undesirable - it's better to use `unsafePrettyPrint` only for very simple printing. So I'd suggest disabling it until a better approach is found.

### DIFF
--- a/src/Hint/Naming.hs
+++ b/src/Hint/Naming.hs
@@ -70,7 +70,8 @@ naming seen originalDecl =
     [ suggest "Use camelCase"
                (shorten originalDecl)
                (shorten replacedDecl)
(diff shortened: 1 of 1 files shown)


ndmitchell:
Even more than that, it changes an identifier, so an automatic replacement is never going to work properly - you have to change the usages too. Best not replace at all if you can't fix things up.

PR #1037, opened 2020-06-09
Do not require @NoRefactor for hints that don't support refactoring

zliu41:
The goals are (1) document exactly which hints don't support refactoring, and (2) no longer require `@NoRefactor` for those hints (which is the vast majority of `@NoRefactor`); use `@NoRefactor` only for other issues like bugs.

If a test case has a hint where `ideaRefactorings = []`, the hint must be added to the list of hints that don't support refactorings, and no `@NoRefactoring` is required (I would also like to check the other direction, namely every hint in the list indeed has an empty `ideaRefactoring`, but this is currently not possible since some hints such as "Redundant return" sometimes support refactoring and sometimes don't).

This uncovers two previously unknown hlint refactoring bugs and one apply-refact bug.

### DIFF
--- a/src/Hint/Bracket.hs
+++ b/src/Hint/Bracket.hs
@@ -40,7 +40,7 @@ issue909 = foo (\((x :: z) -> y) -> 9 + x * 7)
 issue909 = foo (\((x : z) -> y) -> 9 + x * 7) -- \(x : z -> y) -> 9 + x * 7
 issue909 = let ((x:: y) -> z) = q in q
 issue909 = do {((x :: y) -> z) <- e; return 1}
-issue970 = (f x +) (g x) -- f x + (g x) @NoRefactor
+issue970 = (f x +) (g x) -- f x + (g x)
 issue969 = (Just \x -> x || x) *> Just True
 
 -- type bracket reduction
@@ -63,7 +63,7 @@ yes = white $ keysymbol -- white keysymbol
 yes = operator foo $ operator -- operator foo operator
 no = operator foo $ operator bar
 yes = return $ Record{a=b}
-no = f $ [1,2..5] -- f [1,2..5] @NoRefactor: apply-refact bug; see apply-refact #51
+no = f $ [1,2..5] -- f [1,2..5] @NoRefactor: apply-refact bug; fixed in apply-refact #51
 
 -- $/bracket rotation tests
 yes = (b $ c d) ++ e -- b (c d) ++ e
--- a/src/Hint/Duplicate.hs
+++ b/src/Hint/Duplicate.hs
@@ -7,15 +7,15 @@ If you have n the same, error out
 
 <TEST>
 main = do a; a; a; a
-main = do a; a; a; a; a; a -- ??? @NoRefactor: refactoring not supported for duplication hints.
-main = do a; a; a; a; a; a; a -- ??? @NoRefactor
-main = do (do b; a; a; a); do (do c; a; a; a) -- ??? @NoRefactor
-main = do a; a; a; b; a; a; a -- ??? @NoRefactor
+main = do a; a; a; a; a; a -- ???
+main = do a; a; a; a; a; a; a -- ???
+main = do (do b; a; a; a); do (do c; a; a; a) -- ???
+main = do a; a; a; b; a; a; a -- ???
 main = do a; a; a; b; a; a
-foo = a where {a = 1; b = 2; c = 3}; bar = a where {a = 1; b = 2; c = 3} -- ??? @NoRefactor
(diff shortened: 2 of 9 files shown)


ndmitchell:
I'm not a massive fan of centralising a list of all the names that don't support refactoring. We've been fairly lose with the names and not assigning them much identity thus-far, and that gives us a certain level of freedom we lose with this patch. Isn't `null . ideaRefactoring` enough of a hint that in this case the hint isn't supported by this hint?

Looking at the type, why is `ideaRefactoring` a list and not a Maybe? I wouldn't be averse to making it a type:

```
data Refactor = Refactor (Refactoring R.SrcSpan) | NoRefactor
```

And then if what you're looking to do is grep for no refactorings you could grep for `NoRefactor`?

ndmitchell:
I think the general goal of using `@NoRefactor` type annotations for bugs rather than not-yet-implemented is a good one.

zliu41:
Adding a `NoRefactor` constructor sounds good to me.

> Looking at the type, why is ideaRefactoring a list and not a Maybe?

Some hints require more than one `Refactoring`s, for example
```haskell
import A (foo)
import A (bar)
import A (baz)
```
requires one `Replace` and two `Delete`s. In this particular case it may be possible to use a single `Refactoring`, but I imagine in some cases you may want to change both an expression and a type, and that definitely needs two `Refactoring`s.

ndmitchell:
OK, cool, that makes a lot of sense why its a list.

zliu41:
@ndmitchell It occurred to me that we don't really have much freedom with hint names anyway, since hint names are needed in order to enable and disable hints, and so changing a name could break many people's `.hlint.yaml`.

Given that circumstance, it seems reasonable to put the names of all hints that don't support refactoring in one place, which is convenient as a quick reference. What do you think?

shayne-fletcher:
I'm reminded that a feature request comes my way every so often for the ability to "list all hints". Perhaps together with this  we might implement an option to generate a report detailing hints, enabled/disabled status, refactor capable and whatever else we can think of?

zliu41:
@shayne-fletcher very good point. I only recently became aware of the "Use module export list" hint (disabled by default). I had no idea before that it existed.

One option is to generate the list of hints when running tests, provided each hint has at least one test case. We can even include an example for each hint, which can be the first test case for that hint.

zliu41:
I tried generating a summary of builtin hints as a side effect of running the tests, which looks like

```
* Hint: Avoid lambda (Suggestion)
  Sample input:
    f = foo (\y -> g x . h $ y)
  Found:
    \ y -> g x . h $ y
  Perhaps:
    g x . h
  Support refactoring: Sometimes

* Hint: Avoid lambda (Warning)
  Sample input:
    f = foo (\x y -> fun x y)
  Found:
    \ x y -> fun x y
  Perhaps:
    fun
  Support refactoring: Sometimes

* Hint: Avoid lambda using `infix` (Suggestion)
  Sample input:
    foo x = bar (\ d -> search d table)
  Found:
    (\ d -> search d table)
  Perhaps:
    (`search` table)
  Support refactoring: Yes

* Hint: Avoid restricted function (Warning)
  Sample input:
    foo = unsafePerformIO
  Found:
    unsafePerformIO
  Perhaps:
  Support refactoring: No

* Hint: Collapse lambdas (Suggestion)
  Sample input:
    f = foo (\x -> \y -> x x y y)
  Found:
    \ x -> \ y -> x x y y
  Perhaps:
    \ x y -> x x y y
  Support refactoring: Yes

* Hint: Eta reduce (Warning)
  Sample input:
    fun x y z = f x y z
  Found:
    fun x y z = f x y z
  Perhaps:
    fun = f
  Support refactoring: No

...
```

To make sure this is up to date, we can fail the test if the output is different from the existing summary, and add a `--fix-summary` flag to overwrite it.

shayne-fletcher:
That's awesome @zliu41! How about generating in markup or html and then we could probably arrange for a link to it from the README? Perhaps generating an index for it along the way?

ndmitchell:
Awesome. I think that will be an easy way to get what everyone wants! My suggestion would be we add `--generate-summary` flag to `hlint test` and make it write to a file. We check that file into the repo, which means we can serve it up as a web page on GitHub. Given that GitHub doesn't like serving up HTML that much, it suggests a .md file might be the best choice.

If we check that file into the repo, and regenerate it during the test, then if the file changes it will fail CI anyway. That way we'll have a test.

shayne-fletcher:
Really looking forward to seeing this @zliu41 !

zliu41:
SGTM! Will open another PR.

PR #1040, opened 2020-06-10
Fix refactoring for "Avoid lambda"

zliu41:
Fixes a bug where `x = f (\z -> y + z)` is refactored into `x = f (y)`.

Also changed the span of the hint to `(\z -> y + z)` instead of `\z -> y + z`. Otherwise the refactoring output would be `x = f ((y +))`.


ndmitchell:
Thanks

PR #1041, opened 2020-06-10
Make pragma ordering consistent between hint and refactoring

zliu41:
Puts `LANGUAGE` before `OPTIONS_GHC` in both hints and refactorings. Fixes a test case, which failed because `{-# OPTIONS_GHC -cpp -w #-}` is refactored into 
```
{-# OPTIONS_GHC -w #-}
{-# LANGUAGE CPP #-} 
```


ndmitchell:
Thanks! Good catch

PR #1042, opened 2020-06-10
Fix redundant lambda refactoring

zliu41:
There's one thing missing in #1029: given `f (Just a) = \a -> a + a`, the function name `f` must also be removed from subts vars. This ensures `a = \x -> x` won't be refactored into `x x = x`.


ndmitchell:
Thanks!

PR #1043, opened 2020-06-10
Fix refactoring for "use uncurry"

zliu41:
Fixes a bug where `fst foo .= snd foo` is refactored into `uncurry .= foo`. Fixes https://github.com/mpickering/apply-refact/issues/7.

I spoke too soon when I said "no more known HLint refactoring bug" - I keep finding new ones :grinning: Hopefully this is the last.

### DIFF
--- a/src/GHC/Util/HsExpr.hs
+++ b/src/GHC/Util/HsExpr.hs
@@ -30,7 +30,6 @@ import GHC.Util.View
 import Control.Applicative
 import Control.Monad.Trans.State
 
-import Data.Char
 import Data.Data
 import Data.Generics.Uniplate.DataOnly
 import Data.List.Extra
--- a/src/GHC/Util/HsExpr.hs
+++ b/src/GHC/Util/HsExpr.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE ViewPatterns, MultiParamTypeClasses , FlexibleInstances, FlexibleContexts #-}
+{-# LANGUAGE ViewPatterns, MultiParamTypeClasses, FlexibleInstances, FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE TupleSections #-}
 
 module GHC.Util.HsExpr (
@@ -28,6 +29,7 @@ import GHC.Util.FreeVars
 import Control.Applicative
 import Control.Monad.Trans.State
 
+import Data.Char
 import Data.Data
 import Data.Generics.Uniplate.DataOnly
 import Data.List.Extra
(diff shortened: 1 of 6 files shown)


src/GHC/Util/HsExpr.hs
```diff
@@ -290,13 +292,21 @@ descendBracketOld op x = (descendIndex g1 x, descendIndex g2 x)
...
+      | needBracketOld i x y = (addParen y, addParen z)
+      -- https://github.com/mpickering/apply-refact/issues/7
+      | isOp y = (y, addParen z)
     f _ y z                 = (y, z)
 
     f1 = ((fst .) .) . f
     f2 = ((snd .) .) . f
 
+    isOp = \case
+      L _ (HsVar _ var) -> (not . all isAlphaNum) (rdrNameStr var)
```
ndmitchell, line 307: I bet this classifies `foo'` as an operator, which I imagine is wrong? And probably `_foo` too?

ndmitchell: I see `isOp` in src\Hint\Extensions.hs that looks better. Perhaps put that somewhere general and reuse in both cases?

shayne-fletcher: This one?
```
    isOp :: LIdP GhcPs -> Bool
    isOp name = case rdrNameStr name of
      (c:_) -> not $ isAlpha c || c == '_'
      _ -> False
```
> Perhaps put that somewhere general and reuse in both cases?

Pull it down to `ghc-lib-parser-ex` perhaps?

shayne-fletcher: In any case, it's got an interesting signature. It's not wrong but I more obviously expressed as `isOp :: Located RdrName -> Bool`.

ndmitchell: Why not just RdrName? And does GHC itself already have something?

shayne-fletcher: > Why not just RdrName?

That'd work.

> And does GHC itself already have something?

That's an excellent question! `ghc-exactprint` offers this:
```haskell
isSymbolRdrName :: GHC.RdrName -> Bool
isSymbolRdrName n = GHC.isSymOcc $ GHC.rdrNameOcc n
```
where,
```haskell
isSymOcc :: Name -> Bool
isSymOcc n
  = case nameBase n of
      []    -> True  -- Empty name; weird
      (c:_) -> startsVarSym c
                   -- c.f. OccName.startsVarSym in GHC itself
```
So, what do you think? Shall we harvest the ghc-exactprint definition?

shayne-fletcher: TIL
```haskell

-----------------------------------------------------------------------------
-- |
-- Module      :  GHC.Lexeme
-- Copyright   :  (c) The GHC Team
--
-- Maintainer  :  ghc-devs@haskell.org
-- Portability :  portable
--
-- Functions to evaluate whether or not a string is a valid identifier.
--
module GHC.Lexeme (
          -- * Lexical characteristics of Haskell names
        startsVarSym, startsVarId, startsConSym, startsConId,
        startsVarSymASCII, isVarSymChar, okSymChar
  ) where
```
Nice!

shayne-fletcher: Fyi, `isSymOcc` (via `okSymChar` via `startsVarSym`) correctly rules out `_foo`.

ndmitchell:
Thanks! I'm merging now, and I'll fix up the isRdrSymbolName "conflict" with the stuff Shayne added locally and push.

shayne-fletcher:
No actual conflict though right? Just the opportunity to use the `Name.Reader` version IIUC?

ndmitchell:
I think so, not really just, but definitely making that change!

PR #1046, opened 2020-06-11
HSE.All -> All

shayne-fletcher:
A last(?) vestige removed? Move `src/HSE/All.hs` to `src/All.hs`.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -62,8 +62,6 @@
 
 # doesn't fit with the other statements
 - ignore: {name: Use let, within: [HLint, Test.All]}
-# would be a lot of is/from exports
-- ignore: {name: Use explicit module export list, within: HSE.Util}
 # this const has meaingful argument names
 - ignore: {name: Use const, within: Config.Yaml}
 # TEMPORARY: this lint is deleted on HEAD
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -109,6 +109,7 @@ library
         Config.Type
         Config.Yaml
 
+        GHC.All
         GHC.Util
         GHC.Util.ApiAnnotation
         GHC.Util.View
@@ -121,7 +122,6 @@ library
         GHC.Util.Scope
         GHC.Util.Unify
 
-        HSE.All
         Hint.All
         Hint.Bracket
         Hint.Comment
--- a/src/Apply.hs
+++ b/src/Apply.hs
@@ -3,7 +3,7 @@ module Apply(applyHints, applyHintFile, applyHintFiles) where
 
 import Control.Applicative
 import Data.Monoid
-import HSE.All
+import GHC.All
 import Hint.All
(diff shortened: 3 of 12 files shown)


ndmitchell:
Not a fan of moving this to All. Before we had Hint.All and HSE.All (which was really just all the AST stuff). Probably should be GHC.All? Or maybe GHC.Parser?

shayne-fletcher:
> Not a fan of moving this to All. Before we had Hint.All and HSE.All (which was really just all the AST stuff). Probably should be GHC.All? Or maybe GHC.Parser?

I'm easy @ndmitchell. I'll ask you to make the call and I'll make it happen!

ndmitchell:
Does GHC.Parser seem right to you?

shayne-fletcher:
> Does GHC.Parser seem right to you?

To me that choice sparks dissonance with the actual `GHC.Parser` module. Looking at the contents of `All`, I'd be inclined to advocate `Driver` actually.

shayne-fletcher:
> > Does GHC.Parser seem right to you?
> 
> To me that choice sparks dissonance with the actual `GHC.Parser` module. Looking at the contents of `All`, I'd be inclined to advocate `Driver` actually (but even then if that were `GHC.Driver` we have overloading with GHC actual hence I literally mean `Driver`).

Hmmm. Maybe therein is the path out? Recognizing that we've been in states where there are parallel parse trees, perhaps `Driver.GHC`?

shayne-fletcher:
In any case, if between the suggestions of `GHC.Parser` and `GHC.All`, I would vote for `GHC.All`.

ndmitchell:
It seems either it should go somewhere in the GHC.* tree. Or, alternatively, we need a new AST.* tree and things should remain in GHC.* if they are GHC utility wrappers, and move to AST.* if (some condition I can't articulate)

shayne-fletcher:
> It seems either it should go somewhere in the GHC.* tree. Or, alternatively, we need a new AST.* tree and things should remain in GHC.* if they are GHC utility wrappers, and move to AST.* if (some condition I can't articulate)

😁 

I say, let's go with `GHC.All`. It's a mapping of the status quo. When/if the need proves itself to do more, we can cross those bridges as the need arises.

ndmitchell:
Makes sense!

shayne-fletcher:
> if they are GHC utility wrappers, and move to AST.* i

I think there is very little code now that matches the tag "GHC utility". Really what resides under `GHC` are those core algorithms like bracket, unify, pretty lambda etc. that are now expressed in terms of GHC. If we were to find a better name for those I think we could eliminate `GHC.Util` entirely. Something to think about going forward.

shayne-fletcher:
> Makes sense!

I'll get on it and update this PR accordingly. Stand by.

ndmitchell:
Makes a lot of sense - thanks!

PR #1047, opened 2020-06-12
two new hints: Redundant take (in context of zip)

jvoigtlaender:
I found this pattern in student code (in the context of dealing with infinite lists, so either `x` or `y` or both were infinite lists, of which finite prefixes were taken, then zipped).

The rules might be of interest for general use, so I added them to the `default` group. But I could alternatively change the pull request to add them to the `teaching` group instead.

Moreover, analogous rules for `zipWith` instead of `zip` would be conceivably reasonable as well.


ndmitchell:
Thanks!

Issue #144, opened 2015-07-13
Add web page link


ndmitchell:
Give a page you can paste in text and get back HLint suggestions. Either just link to an existing paste site that does it, or write my own on top of the HLint API.

mpickering:
Here's a demo of something I put together this afternoon. It needs to be cleaned up (and it relies on a hacked version of ghc-exactprint) but I think it could be good..

http://gfycat.com/ComposedPotableEastrussiancoursinghounds

ndmitchell:
Very cool!

ndmitchell:
I like this a lot. What tech did you use? Haskell server or GHCJS?

mpickering:
I used servant. I tried to use GHCJS but it's hard to compile dependencies which depend on GHC.

ndmitchell:
Something based on GHCJS would be my preference, so we can skip server issues.

vaibhavsagar:
I have a sloppy but functional proof-of-concept up at https://lively-opaque-snipe.glitch.me/ using GHCJS, the code for which is [here](https://github.com/vaibhavsagar/experiments/tree/0ee2dbdfc20b1acf6e5dd1dca351ebab5a5bc0fb/hlint-live). I mostly followed the steps outlined in [this blog post I wrote](https://vaibhavsagar.com/blog/2019/10/29/getting-along-with-javascript/), and I had to make a couple of small changes to HLint:

1. The `yaml` library doesn't support GHCJS at all, so I switched [to using `HsYAML` and `HsYAML-aeson`](https://github.com/vaibhavsagar/experiments/commit/6a9e1cacb80395335dc1d3ec720b6675a3f09c80#diff-42e52dc3cc53ea1eb0c98de2601dc280).
2. GHCJS doesn't seem to allow me to read files from the disk, so [I used `file-embed` to copy the contents of `hlint.yaml` into my app and used that instead](https://github.com/vaibhavsagar/experiments/commit/824545895324b30e03122954f750adab146cd79b#diff-42e52dc3cc53ea1eb0c98de2601dc280).

The resulting app seems pretty laggy, because my Reflex-fu is weak and I'm probably using the HLint API in the simplest and slowest possible way, so it might not appear to work at first. I suggest trying with a warning-laden snippet of code such as:

```haskell
module Main where

main = do
  print $ "Hello world"
  print $ "Does this work?"
```

and then editing the text until hints show up:

[...]

PR #1050, opened 2020-06-14
Allow using HsYAML instead of yaml

vaibhavsagar:
As discussed in https://github.com/ndmitchell/hlint/issues/144#issuecomment-630048977, this PR adds the option of using `HsYAML` instead of `yaml` behind a flag.

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -48,6 +48,11 @@ flag ghc-lib
   manual: True
   description: Force dependency on ghc-lib-parser even if GHC API in the ghc package is supported
 
+flag hsyaml
+  default: False
+  manual: True
+  description: Use HsYAML instead of yaml
+
 library
     default-language:   Haskell2010
     build-depends:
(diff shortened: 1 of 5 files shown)


hlint.cabal
```diff
@@ -83,6 +87,14 @@ library
     else
         cpp-options: -DGPL_SCARES_ME
 
+    if flag(hsyaml) && flag(gpl)
```
ndmitchell, line 90: If you set hsyaml but not gpl, then i think it's reasonable that it uses hsyaml anyway, so I'd simplify this conditional to just hsyaml.

vaibhavsagar: Sure, I will update this.

vaibhavsagar:
I've raised it on the `HsYAML` issue tracker. I agree that it would simplify our implementation.

ndmitchell:
Thanks!

vaibhavsagar:
My pleasure!

PR #1053, opened 2020-06-17
Add missing IncreasesLaziness

josephcsible:
`zip (take 1 undefined) (take 0 [])` bottoms, but `take (min 1 0) (zip undefined [])` is `[]`.


ndmitchell:
Super subtle! Thanks for spotting.

PR #1054, opened 2020-06-18
make 'isLitString' work again

owestphal:
After the transition to the GHC parser the ```isLit*``` pattern is only supported for ```Int```. We were using ```isLitString``` for some local rules (see https://github.com/ndmitchell/hlint/pull/686#issuecomment-505625644) that we would like to see working again.


shayne-fletcher:
I can only assumed I messed this up 😞  Sorry @owestphal and thanks for taking the time to fix this!

ndmitchell:
Thanks for spotting and fixing! I'll make a release later today with this fix in it.

PR #1056, opened 2020-06-23
Generate a summary of built-in hints

zliu41:
Based on suggestions from @shayne-fletcher and @ndmitchell in #1037.

- Generated from `cabal run hlint test -- --generate-summary`
- One entry per (hint name, severity, does it support refactoring)
- No change to CI yet (i.e., it currently does not check that the summary is up to date). The check can be done either in `cabal run hlint test` or in the CI script. Not sure which one is preferred. Either way it can be done separately.

### DIFF
--- a/src/Hint/Pragma.hs
+++ b/src/Hint/Pragma.hs
@@ -15,7 +15,7 @@
 {-# OPTIONS_YHC -cpp #-}
 {-# OPTIONS_GHC -XFoo #-} -- {-# LANGUAGE Foo #-}
 {-# OPTIONS_GHC -fglasgow-exts #-} -- ??? @NoRefactor
-{-# LANGUAGE RebindableSyntax, EmptyCase, DuplicateRecordFields, RebindableSyntax #-} -- {-# LANGUAGE RebindableSyntax, EmptyCase, DuplicateRecordFields #-}
+{-# LANGUAGE RebindableSyntax, EmptyCase, RebindableSyntax #-} -- {-# LANGUAGE RebindableSyntax, EmptyCase #-}
 {-# LANGUAGE RebindableSyntax #-}
 {-# OPTIONS_GHC -cpp -foo #-} -- {-# LANGUAGE CPP #-} {-# OPTIONS_GHC -foo #-} @NoRefactor -foo is not a valid flag
 {-# OPTIONS_GHC -cpp -w #-} -- {-# LANGUAGE CPP #-} {-# OPTIONS_GHC -w #-}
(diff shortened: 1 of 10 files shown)


src/Test/Util.hs
```diff
@@ -1,22 +1,41 @@
-{-# LANGUAGE RecordWildCards, GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE RecordWildCards, GeneralizedNewtypeDeriving, StrictData #-}
```
ndmitchell, line 1: Can we just put ! on individual fields. I've been burnt by StrictData in the past.

src/Hint/Pragma.hs
```diff
@@ -15,7 +15,7 @@
 {-# OPTIONS_YHC -cpp #-}
 {-# OPTIONS_GHC -XFoo #-} -- {-# LANGUAGE Foo #-}
 {-# OPTIONS_GHC -fglasgow-exts #-} -- ??? @NoRefactor
-{-# LANGUAGE RebindableSyntax, EmptyCase, DuplicateRecordFields, RebindableSyntax #-} -- {-# LANGUAGE RebindableSyntax, EmptyCase, DuplicateRecordFields #-}
```
ndmitchell, line 18: Why this change?

ndmitchell:
Cool! One little remark, but lets merge and iterate - this is a great start!

shayne-fletcher:
I agree, this is a great start! Couple of thoughts for the next iterations:
  - should `--generate-summary` take a path argument rather than hardcoding "builtin.md"?
  - can the summary be generalized someway, somehow to consider *all* hints in scope (including those that are not builtin)?

shayne-fletcher:
Guess Appveyor needs a retry.
Screen Shot 2020-06-23 at 10 23 08 AM

src/Hint/Pragma.hs
zliu41: The shorter version fits in a single line in the markdown table. I added the long version back below it.

src/Test/Util.hs
zliu41: Done!

zliu41:
> can the summary be generalized someway, somehow to consider all hints in scope (including those that are not builtin)?

I think this is easy to do; I didn't include lhs => rhs hints because `hlint.yaml` already lists those hints.

shayne-fletcher:
> > can the summary be generalized someway, somehow to consider all hints in scope (including those that are not builtin)?
> 
> I think this is easy to do; I didn't include lhs => rhs hints because `hlint.yaml` already lists those hints.

Personally, I think "in for a penny, in for a pound". A complete report including those that aren't builtin on one handy web-page with a listing of their status (and perhaps indicators of whether they are builtin or custom) would be happy days for me 😁 . @ndmitchell thoughts?

ndmitchell:
I'd like to do all hints, builtin and otherwise, with examples for all. The lhs/rhs ones are pretty easy to do builtin ones, because the LHS is a good example. But that can come later.

Unfortunately Windows and Haskell is now a bit of a dead zone. I happily merge with Appveyor failures, but feel bad about it.

zliu41:
Sounds good, let's add the lhs/rhs hints in the near future.

ndmitchell:
Thanks!

PR #1063, opened 2020-06-25
Remove @NoRefactor for hints that don't support refactoring

zliu41:
Hints that don't support refactoring are now documented, so no need to require `@NoRefactor` for these hints - leaving `@NoRefactor` for real bugs.

This eliminates the vast majority of `@NoRefactor`s, and if a hint starts to support refactoring in the future, it will be automatically tested.


ndmitchell:
Thanks!

Issue #1064, opened 2020-06-27
Incorrect hint on unused overloaded lists pragma


LeventErkok:
This is happening with hlint 3.1.6; I'm guessing it got introduced really recently.

For this program:

```haskell
{-# LANGUAGE OverloadedLists #-}
{-# LANGUAGE TypeFamilies    #-}

import GHC.Exts (IsList(..))

data X a

instance IsList (X a) where
  type Item (X a) = a
  fromList = undefined
  toList   = undefined

g :: X a
g = []
```

I get:

```
$ ghci try.hs
GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( try.hs, interpreted )
Ok, one module loaded.
```

But hlint says:

```
$ hlint try.hs
try.hs:1:1-32: Warning: Unused LANGUAGE pragma
Found:
  {-# LANGUAGE OverloadedLists #-}
Perhaps you should remove it.

1 hint
```

Alas, after removing the hint, ghc no longer likes the program:

```
$ ghci try.hs
GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( try.hs, interpreted )

try.hs:13:5: error:
    • Couldn't match expected type ‘X a’ with actual type ‘[a0]’
    • In the expression: []
      In an equation for ‘g’: g = []
    • Relevant bindings include g :: X a (bound at try.hs:13:1)
   |
13 | g = []
   |     ^^
Failed, no modules loaded.
```

shayne-fletcher:
I confirm I can reproduce the behavior. My impression is that the `used OverloadedLists = ...` detection mechanism requires strengthening to include the usage of the extension exhibited by this case, presumably by considering instance declarations of `IsList`.

shayne-fletcher:
Note that the use of the `[]` expression in `g` is neither an `ExplicitList` or `ArithSeq`, rather in the parse tree it is an `HsVar` of the exact name `[]`. It would be sufficient to extend `isListExpr` to consider this constant  to recover the expected behavior.

shayne-fletcher:
Indeed, when instead we write
```haskell
g :: X Int
g = [1]
```
the `OverloadedLists` is classified "used" and we instead get the hint,
```
Found:
  {-# LANGUAGE OverloadedLists #-}
Note: may break the code
```

shayne-fletcher:
[...]

PR #1065, opened 2020-06-27
Add missing case to isListExpr

shayne-fletcher:
Add a missing case for the empty list in `isListExpr`. Fixes https://github.com/ndmitchell/hlint/issues/1064.


ndmitchell:
Thanks!

PR #1070, opened 2020-07-03
Warn on unused ImportQualifiedPost

zliu41:


shayne-fletcher:
The predicate should move down to ghclib-parser-ex at some point. Happy to take care of it - any chance I can ask you to raise an issue for it in that repo (I'm without a computer for a little while)?

src/Hint/Extensions.hs
```diff
@@ -456,6 +461,11 @@ used MagicHash = hasS f ||^ hasS isPrimLiteral
     f :: RdrName -> Bool
     f s = "#" `isSuffixOf` occNameStr s
 used PatternSynonyms = hasS isPatSynBind ||^ hasS isPatSynIE
+used ImportQualifiedPost = hasS isImportQualifiedPost
```
ndmitchell, line 464:
```suggestion
used ImportQualifiedPost = hasS (== QualifiedPost)
```

Is that sufficient?

googleson78: Well it does hypothetically contain breakage to `ghc-lib-parser-ex`, potentially saving users the hassle.

There are a bunch of other predicates like this  in `ghc-lib-parser-ex` (e.g. [isPViewPat](https://hackage.haskell.org/package/ghc-lib-parser-ex-8.10.0.14/docs/src/Language.Haskell.GhclibParserEx.GHC.Hs.Pat.html#isPViewPat)), I'm guessing with the same intention (?)

zliu41: Yeah the suggestion looks good.
@googleson78 I'm not sure what you mean 🙂 Why is it a breakage?

zliu41:
@shayne-fletcher Sure thing. I raised https://github.com/shayne-fletcher/ghc-lib-parser-ex/issues/62.

src/Hint/Extensions.hs
googleson78: If the `QualifiedPost` constructor ever changes, if `ghc-lib-parser-ex` exposes `isImportQualifiedPost`, only it would have to change, whereas by using `== QualifiedPost` you will also have to change your code.

zliu41: I see. Yeah, `isImportQualifiedPost` will be added to `ghc-lib-parser-ex`.

ndmitchell: The GHC API is a huge surface area that certainly will change - but not in ways we can predict. All we can do is have tests and write simple code - premature dealing with potential breakages is not that helpful. They always surprise us!

ndmitchell:
Thanks! Great to see such a wondrous extension supported throughout the ecosystem!

PR #1071, opened 2020-07-03
Don't test refactor if there's no hint

zliu41:
Apply-refact no longer attempts to parse and re-print the input since 0.8.1.0 if there's no hint, so we don't need to test refactor if there's no hint. This makes the test suite run 17% faster on my machine.


ndmitchell:
Thanks!

PR #1072, opened 2020-07-04
Update ghc-lib-parser-ex

shayne-fletcher:
## 8.10.0.15 released 2020-07-04
- New function `isImportQualifiedPost`

Just the update (one of those sanity preserving things); stops short of using `isImportQualifiedPost`. There is the option now, I'll leave whether to actually do that up to @zliu41 (I'm easy)!

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -80,7 +80,7 @@ library
       build-depends:
           ghc-lib-parser == 8.10.*
     build-depends:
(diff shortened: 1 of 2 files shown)


shayne-fletcher:
Bit of Appveyor stochasticity would seem.

Screen Shot 2020-07-04 at 12 39 57 PM

ndmitchell:
Alas, the Windows CI is pretty seriously flakey at the moment.

PR #1073, opened 2020-07-05
Add LHS/RHS hints to summary

zliu41:
As discussed in #1056, adding LHS/RHS hints to the summary. Usage: `hlint summary output.md`.

The `hints.md` included in the commit is generated by
```
cabal run hlint summary -- hints.md --no-implicit
```
where `--no-implicit` says "don't use `.hlint.yaml` implicitly".

Some room for improvement:
- Organize LHS/RHS hints by group
- Include disabled groups
- Include where a hint comes from (`hlint.yaml`, `.hlint.yaml` etc.)
- Include "Avoid restricted ..." hints

cc @shayne-fletcher


hlint.cabal
```diff
@@ -13,17 +13,17 @@ description:
...
+    data/hlint.yaml
+    data/default.yaml
+    data/Test.hs
+    data/report_template.html
+    data/hs-lint.el
+    data/hlint.1
+    data/hlint.ghci
+    data/HLint_QuickCheck.hs
+    data/HLint_TypeCheck.hs
+    src/Hint/*.hs
```
shayne-fletcher, line 26: Can you comment on why this change?

src/CmdLine.hs
```diff
@@ -140,6 +141,14 @@ data Cmd
         ,cmdCppSimple :: Bool
         ,cmdCppAnsi :: Bool
         }
+    | CmdSummary
+        {cmdGivenHints :: [FilePath]
+        ,cmdWithGroups :: [String]
+        ,cmdIgnore :: [String]
+        ,cmdFindHints :: [FilePath]
+        ,cmdNoImplicit :: Bool           -- ^ Do not use .hlint.yaml implicitly
```
shayne-fletcher, line 149: What precisely does "use .hlint.yaml implicitly" mean?

hlint.cabal
zliu41: The summary of the built-in hints is generated by running the test cases in `src/Hint/*.hs`. Adding these files to `data-files` ensures they are installed together with HLint.

src/CmdLine.hs
zliu41: If no explicit hint file is specified via `--hint`, HLint looks for `.hlint.yaml` in the current directory and uses it implicitly. This flags disables that, which is useful when running `cabal run hlint summary` from HLint dir, because HLint's `.hlint.yaml` is only relevant to the HLint codebase.

hlint.cabal
shayne-fletcher: I'm no expert on these matters (!) but perhaps these `.hs` files should be in the `extra-source-files` stanza rather than the `data-files` stanza? If so then the `data-dir` property could maybe be restored?

src/CmdLine.hs
shayne-fletcher: Ok. So, if not disabled with `--no-implicit`, the `.hlint.yaml` file found in the project will be consulted and any hints it contains will be reported on. We generally don't want this because such rules are specific to the hlint project itself  (not for general users of hlint). Have I got that right?

hlint.cabal
zliu41: I'm not expert either, but `data-files` are available at runtime via `getDataFileName`, while `extra-source-files` don't seem to be. These files are needed when running `hlint summary`.

shayne-fletcher: See https://cabal.readthedocs.io/en/stable/developing-packages.html#pkg-field-extra-source-files. Not sure if it's got the properties we need of it. Just to be clear, I have no objection to the files being data, just trying to make sure that we're doing it the "right" way 😄

src/CmdLine.hs
zliu41: That's exactly right. It's main purpose is to avoid using HLint's `.hlint.yaml` when generating `hints.md`. When users run it on their own repos they most likely do want to include their `.hlint.yaml` files.

ndmitchell:
Appologies for not revisiting this PR sooner. I got a bit ill. I hope to get back to it within the next few days.

zliu41:
@ndmitchell Get whale soon!

shayne-fletcher:
> @ndmitchell Get whale soon!

That's very funny considering @ndmitchell is well known to be a shark! 🤣

shayne-fletcher:
@zliu41 
Ec4YjGlXgAElO_X

hlint.cabal
googleson78: It's great how documented cabal (and stack too..) is and how to do stuff using it. -_-

ndmitchell:
29 days ago I said "soon". Unfortunately I've been rather busy with personal stuff since then, and since this patch requires a few hours of focus, I just haven't had time for it. Still definitely on the pile though.

ndmitchell:
Looking at it, most of this patch is great, but some of it doesn't quite match the UI I want to provide (see #1124) and I'm still not convinced about making it a user function to generate the hints.md file including builtin hints. As such, I'm going to copy/paste the vast majority of this code in, commit and tweak to how I think it should be done, and then we can discuss the remaining pieces, how they differ, and what to do about it.

zliu41:
Sounds good - closing the PR.

shayne-fletcher:
> Sounds good - closing the PR.

It's nice to see your good work landing

zliu41:
@shayne-fletcher Thanks!

ndmitchell:
Thanks for your hard work on this @zliu41, and sorry it took so disastrously long to get merged. I copied 90% of the code from this PR, integrated it slightly different, and then did some refactors on top. I'll outline my changes here, and happy to discuss whether the original formulation by @zliu41 is better in any way and adjust.

1. This feature seems to be designed to work for our central hints.md _and_ users generating their own hint database. I'm not convinced the latter is worth it since I just can't imagine people doing so.  As such, I simplified the code to not package our source code in the data directory - you just get a slightly smaller hint list if you run it without having source available.
2. I turned this from being a mode, to a flag `--generate-summary` on the `lint` command. The reason is I think we want to move away from modes as per #1124. Doesn't change much apart from where the code gets invoked from.

I also regularised the format between user hints and builtin hints, and broke the builtin hints up into their sections. I adjusted the code to use more lists and less folds. I filed #1126 and #1127 to do some follow ups, but they are way less priority. But the essence is very much the same. Thanks for your hard code @zliu41!

zliu41:
@ndmitchell Awesome, thanks for following up on this! The `hints.md` is looking pretty good now.

Issue #1075, opened 2020-07-05
Error thrown on CPP in the `base` codebase


Kleidukos:
* Version: `HLint v3.1.6, (C) Neil Mitchell 2006-2020`
* OS: Fedora 32
* Command line: `$ hlint -j --cpp-include="./libraries/base/include/" --cpp-include=includes --cpp-include=./_build/stage1/lib/ -h libraries/base/.hlint.yaml libraries/base`
* Error message:
```
hlint: # error Unknown mingw32 arch
in libraries/base/GHC/Windows.hs  at line 81 col 1
CallStack (from HasCallStack):
  error, called at ./Language/Preprocessor/Cpphs/CppIfdef.hs:113:21 in
  cpphs-1.20.8-DevT7HkEKU3DDQgUoHVnJY:Language.Preprocessor.Cpphs.CppIfdef
```

The CPP code block is :

```c
#if defined(i386_HOST_ARCH)     
# define WINDOWS_CCONV stdcall  
#elif defined(x86_64_HOST_ARCH) 
# define WINDOWS_CCONV ccall    
#else                           
# error Unknown mingw32 arch    
#endif                          
```

shayne-fletcher:
Perhaps adding `--cpp-define x86_64_HOST_ARCH` to the hlint invocation will help here?

Kleidukos:
Perfect! I still get a rather peculiar message from hlint, but other than that, no problem whatsoever!

```
Warning: unknown directive #-}
in libraries/base/Unsafe/Coerce.hs  at line 311 col 1
No hints
```
This is the file in question: https://gitlab.haskell.org/ghc/ghc/-/blob/master/libraries/base/Unsafe/Coerce.hs#L303-311

shayne-fletcher:
> Perfect! I still get a rather peculiar message from hlint, but other than that, no problem whatsoever!

Awesome.

> ```
> Warning: unknown directive #-}
> in libraries/base/Unsafe/Coerce.hs  at line 311 col 1
> No hints
> ```
> 
> This is the file in question: https://gitlab.haskell.org/ghc/ghc/-/blob/master/libraries/base/Unsafe/Coerce.hs#L303-311

Ha. That's a genuine problem. The C-preprocessor is being run over the file and it doesn't know what to make of a line that reads,
```
#-}
```
[...]

PR #1076, opened 2020-07-07
Conditionally run cpp

shayne-fletcher:
This PR relates to discussion in issue https://github.com/ndmitchell/hlint/issues/1075.

We have been running cpp without regard for whether the `CPP` extension is enabled or not. This change avoids running cpp if `CPP` is not in effect. 

Given,
```haskell
{-# Rules
-- Foo, bar, baz
#-}
main = undefined
```
then with `hlint ~/Foo.hs` the warning
```
Warning: unknown directive #-}
in Foo.hs  at line 3 col 1
```
exhibits but with `hlint -XNoCPP ~/Foo.hs` the warning is vanquished.

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -69,8 +69,7 @@ library
         extra >= 1.7.3,
         refact >= 0.3,
         aeson >= 1.1.2.0,
-        filepattern >= 0.1.1,
-        mtl >= 2.2.2
+        filepattern >= 0.1.1
 
     if !flag(ghc-lib) && impl(ghc >= 8.10.0) && impl(ghc < 8.11.0)
       build-depends:
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -69,7 +69,8 @@ library
         extra >= 1.7.3,
         refact >= 0.3,
         aeson >= 1.1.2.0,
-        filepattern >= 0.1.1
+        filepattern >= 0.1.1,
+        mtl >= 2.2.2
 
     if !flag(ghc-lib) && impl(ghc >= 8.10.0) && impl(ghc < 8.11.0)
       build-depends:
--- a/src/GHC/All.hs
+++ b/src/GHC/All.hs
@@ -9,6 +9,7 @@ module GHC.All(
     parseExpGhcWithMode, parseImportDeclGhcWithMode, parseDeclGhcWithMode,
     ) where
 
+import Control.Monad.Except
 import Util
 import Data.Char
 import Data.List.Extra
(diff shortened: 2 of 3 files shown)


src/GHC/All.hs
```diff
@@ -161,17 +161,25 @@ parseModuleEx flags file str = timedIO "Parse" file $ do
...
+        dynFlags <- parsePragmasIntoDynFlags baseDynFlags enableDisableExts file str
         case dynFlags of
           Right ghcFlags -> do
             ghcFlags <- pure $ lang_set ghcFlags $ baseLanguage flags
-            case fileToModule file ppstr ghcFlags of
+            -- Avoid running cpp unless Cpp is enabled (see
+            -- https://github.com/ndmitchell/hlint/issues/1075 as to
+            -- why this is a good idea).
+            srcText <-
+              if Cpp `xopt` ghcFlags
```
ndmitchell, line 173: There is quite a lot of vertical space here. No need for separate lines for then and else to live on their own.

src/GHC/All.hs
```diff
@@ -161,17 +161,25 @@ parseModuleEx flags file str = timedIO "Parse" file $ do
             Nothing | file == "-" -> getContentsUTF8
                     | otherwise -> readFileUTF8' file
         str <- pure $ dropPrefix "\65279" str -- remove the BOM if it exists, see #130
-        ppstr <- runCpp (cppFlags flags) file str
         let enableDisableExts = ghcExtensionsFromParseFlags flags
-        dynFlags <- parsePragmasIntoDynFlags baseDynFlags enableDisableExts file ppstr
+        dynFlags <- parsePragmasIntoDynFlags baseDynFlags enableDisableExts file str
```
ndmitchell, line 165: I think its important to get the pragmas after CPP. Imagine someone does:

```
{-# LANGUAGE CPP #-}
#if condition
{-# LANGUAGE QuasiQuotes #-}
#endif
```

GHC handles that.

shayne-fletcher: Ok, good point. What this means is (1) parse pragmas, (2) if CPP is enabled, do cpp and (3) parse pragmas again. I just looked into `GHC/Driver/Pipeline.hs` and indeed that's what GHC does.

shayne-fletcher:
In https://github.com/ndmitchell/hlint/pull/1076/commits/3f40a26769d8655b7702127d4db317284e22cc97 `parseModuleEx`re-reads pragmas after preprocessing. To help keep the resulting code tractable I've made use of an `ExcepT` monad which induces a dependency on `mtl`. Let me know if this meets with your approval or you'd rather the more verbose code that avoids the dependency? Still thinking about how to write a test for this. Haven't had any good ideas just yet.

shayne-fletcher:
@ndmitchell Happy days - https://github.com/ndmitchell/hlint/pull/1076/commits/7c1707e9ab6cbc2eec9ab14d0a0fd4804802e389 adds a test that is confirmed to fail if `-XNoCPP` is not respected and the C-preprocessor is run.

shayne-fletcher:
https://github.com/ndmitchell/hlint/pull/1076/commits/70a67ba36181f1cbbddb9748dc7cfbfef22b5869 adds tests that for the behaviors specified in comment https://github.com/ndmitchell/hlint/pull/1076#discussion_r452999392 (conditional extension enabling).

hlint.cabal
```diff
@@ -69,7 +69,8 @@ library
         extra >= 1.7.3,
         refact >= 0.3,
         aeson >= 1.1.2.0,
-        filepattern >= 0.1.1
+        filepattern >= 0.1.1,
+        mtl >= 2.2.2
```
ndmitchell, line 73: We already depend on transformers, so no need to introduce an mtl dependency.

src/GHC/All.hs
```diff
@@ -9,6 +9,7 @@ module GHC.All(
     parseExpGhcWithMode, parseImportDeclGhcWithMode, parseDeclGhcWithMode,
     ) where
 
+import Control.Monad.Except
```
ndmitchell, line 12: Use Control.Monad.Trans.Except instead which is in transformers.

ndmitchell:
Thanks! Much better

PR #1080, opened 2020-07-18
Add missing DecreasesLaziness

josephcsible:
The `zip/take` hint can decrease laziness too: `zip (take 0 []) (take undefined [])` is `[]`, but `take (min 0 undefined) (zip [] [])` bottoms.


josephcsible:
I wonder if it would be worth adding a string field to IncreasesLaziness and DecreasesLaziness, like RemovesError has, so that we could say "increases laziness of x, decreases laziness of m" instead of just "increases laziness, decreases laziness"

ndmitchell:
The problem is that there is no `x` or `m` in the users version, since these are the variables we replace. We could substitute them in the note, but that starts to get fiddly. I appreciate it would be nice to know why in this case, but I think it's a pretty rare case, so I'm OK with it being a bit of a special flower. If we did want to do something, we could say that a custom note, plus other notes, just displays the custom note - the custom note is for users, the others are for machines.

PR #1083, opened 2020-07-21
Add hints for mempty as a function

josephcsible:


ndmitchell:
CI gives the error:

```
No name for the hint \ x -> mempty ==> mempty
```

Can you give it an explicit name? HLint gets confused because you replace mempty with mempty, so there is no good default.

josephcsible:
Added.

googleson78:
```haskell
    - hint: {lhs: const mempty, rhs: mempty}
    - hint: {lhs: \x -> mempty, rhs: mempty, name: Redundant lambda}
```
I think these two hints are actively harmful for code clarity. Nobody who I know thinks "oh yeah, of course, this is mempty for `(-> b)`" - their thoughts are more along the lines of "wtf?" followed by opening code to check what the types are.

In any case, even if everyone is familiar with that instance and keeps it in mind at all times, this change makes code harder to mentally type-check for not much syntactic savings (<10 characters?) (in my opinion ofc.)

shayne-fletcher:
@googleson78 

> INobody who I know thinks "oh yeah, of course, this is mempty for `(-> b)`" - their thoughts are more along the lines of "wtf?" followed by opening code to check what the types are.

On seeing this ticket, I for one definitely fell into that camp 🤣  My reaction to it though I think differed from yours - I was excited to find out about the instance and don't these make the best hints? The ones that teach us something?

googleson78:
I have nothing against the hint as a learning experience - in fact I think this instance for `Semigroup` and `Monoid` is "pretty cool", but it's definitely not something I want to have to think about when doing a review/trying to understand a piece of code I haven't worked on.

Not because I'm 100% always against abstractions/complications, but for me a single lambda/const save isn't worth using `mempty` with a different amount of arguments (>0) than most common uses of `mempty` (0).

josephcsible:
FWIW, the real-world use case for this hint that inspired me to add it was from [my answer to this Stack Overflow question](https://stackoverflow.com/a/62865204/7509065), which actually uses that instance twice (saving 2 `const`s):

```haskell
lpref :: Int -> [Int] -> Maybe [Int]
lpref w xs
  | 0 >= w = Just []
  | otherwise = foldr go mempty xs 0 id
  where
    go :: Int -> (Int -> ([Int] -> [Int]) -> Maybe [Int]) -> Int -> ([Int] -> [Int]) -> Maybe [Int]
    go x acc s f
      | s' >= w = Just $ f [x]
      | otherwise = acc s' (f . (x:))
      where s' = x + s
```
Here, I'm using `mempty` in place of `(const (const Nothing))`, which IMO is quite a big simplification since this is already complicated enough.

ndmitchell:
I left this patch a few days to see if I felt more strongly. I'm pretty much on the fence between @googleson78 and @josephcsible. Given that, I threw it open to the mercy of Twitter: https://twitter.com/ndm_haskell/status/1290387392420085760 (not that the poll is a final decision, but good to get a sense of what the rest of the world thinks)

ndmitchell:
71% of Twitter users don't want these hints, so I suggest we put them in generalise-for-conciseness group. The one exception would be `either mempty` since that's already using mempty in the function position, so not complicating anything.

josephcsible:
Okay, I moved the hints that add usage of the function instance to generalise-for-conciseness, but I left the one that removed usage of the instance. I also added one more hint to catch the typo of ``x `mempty` y`` instead of ``x `mappend` y`` that caused a real-life bug at https://dorchard.blog/2020/06/03/a-short-exploration-of-ghcs-instance-resolution-hiding-mistakes-from-the-type-checker/.

data/hlint.yaml
```diff
@@ -240,6 +240,8 @@
     - warn: {lhs: x `mappend` mempty, rhs: x, name: "Monoid law, right identity"}
     - warn: {lhs: foldr (<>) mempty, rhs: mconcat}
     - warn: {lhs: foldr mappend mempty, rhs: mconcat}
+    - warn: {lhs: mempty x, rhs: mempty, name: Evaluate}
+    - warn: {lhs: mempty x y, rhs: mempty, name: Evaluate}
```
ndmitchell, line 244: Doesn't repeated application of `mempty x` catch this, by evaluating twice?

ndmitchell: If you mean to only catch ``x `mempty` y`` then I'd write that explicitly - and perhaps also give it the name "possible typo of mappend"

ndmitchell: I might even suggest ``x `mappend` y`` and give the Note "changes semantics likely because of a bug"

josephcsible: > Doesn't repeated application of `mempty x` catch this, by evaluating twice?

Not unless/until #1096 is fixed.

> If you mean to only catch `` x `mempty` y `` then I'd write that explicitly - and perhaps also give it the name "possible typo of mappend"

Done.

> I might even suggest `` x `mappend` y `` and give the Note "changes semantics likely because of a bug"

I don't really want to do that for the same reason I mentioned in #1092.

ndmitchell: Could always do a `Note: Perhaps you meant to use mappend` even if we make the suggestion to mempty?

ndmitchell: Ah, I see you beat me to it!

PR #1087, opened 2020-07-27
update test instruction

ChefYeum:
Do you need `stack init` when `stack.yaml` is already part of the repo?

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,7 +323,7 @@ You can customize the `Note:` for restricted modules, functions and extensions, 
 
 ## Hacking HLint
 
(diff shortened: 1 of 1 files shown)


ndmitchell:
You don't - good catch. There had been a hope that we could remove the stack.yaml, but that's probably not realistic in the near term.

ndmitchell:
And thanks for the PR!

shayne-fletcher:
Ha. Was just going to comment that @ndmitchell in the past hasn't really been in favor of there being one and I keep on forcing it on him 🤣

PR #1089, opened 2020-07-29
Clarify readme on fixities declared in the same module

zliu41:
A coworker asked whether hlint knows the fixities of operators defined in the same module, and the readme isn't explicit about that. I think it's worth clarifying.


ndmitchell:
Thanks! CI failure seems unrelated, so I'll take a look at that.

ndmitchell:
Hmm, I edited CHANGES.txt on GitHub.com as I always do. But this time it shoved in a Windows newline. No idea why. Now fixed.

PR #1091, opened 2020-07-30
Updated README.md to emphasize .hlint.yaml more.

Martinsos:
I am using hlint via Emacs extension and I had problems with hlint throwing parse error on QuasiQuotes.
Although I read the docs, it took me long time to figure out I had to create .hlint.yaml and add the line to it -> I had to find the solution in one of the issues on this repo.

Therefore, I tried here to explain better in "parse error" part of the faq that solution could be modifying .hlint.yaml. I also tried to emphasize more in the "install" part of the docs that .hlint.yaml exists, and I linked to it.

I know it is mentioned in the usage part, but there it is mentioned in the context of ignoring current warnings, which I didn't want and therefore I ended up not thinking .hlint.yaml is important. While more details are in the faq, I don't normally read faq until I have some problems.


shayne-fletcher:
LGTM

ndmitchell:
Thanks, looks good - the initial version of the README was written before `.hlint.yaml` was a thing, so definitely getting some more references in is great. Appreciated!

Issue #1093, opened 2020-08-03
Warn on unknown language extensions


ndmitchell:
See https://github.com/ndmitchell/hlint/issues/223#issuecomment-665951432 - if you type:

```
hlint -XQuasiQuoe Sample.hs
```

It doesn't warn or anything. It probably should at least warn. Not necessarily as a hint warn, but just `putStrLn $ "Warning: Unknown extension, QuasiQuoe"`.

shayne-fletcher:
Ouch. For when we get to it, I think I see where we need to intervene: function `getExtensions`, `CmdLine.hs:333`.

shayne-fletcher:
> Ouch. For when we get to it, I think I see where we need to intervene: function `getExtensions`, `CmdLine.hs:333`.

If that is indeed the right place, maybe `error` here?
  - It'll be easier to implement;
  - Consistent with the expected behavior of an unrecognized argument.

ndmitchell:
My first inclination was not to error, since if GHC adds a new extension we don't know about then we'd fail. But honestly, now we track GHC so closely, I'm not sure it's a real concern.

shayne-fletcher:
> My first inclination was not to error, since if GHC adds a new extension we don't know about then we'd fail. But honestly, now we track GHC so closely, I'm not sure it's a real concern.

I agree with this; don't think that's a concern.

shayne-fletcher:
> My first inclination was not to error, since if GHC adds a new extension we don't know about then we'd fail. 

On reflection, I think that's the behavior we want if we find ourselves in this case too since it's a very much a pragma we can't satisfy!

shayne-fletcher:
> > My first inclination was not to error, since if GHC adds a new extension we don't know about then we'd fail.
> 
> On reflection, I think that's the behavior we want if we find ourselves in this case too since it's a very much a pragma we can't satisfy!

Ah but I see your point that warning would be preferable in this case - you'd want to say "I can't do that" and proceed to analyze what you can. Ok, scratch that and in any case as noted above, we don't intend to ever be in this path anyway 😉

PR #1094, opened 2020-08-03
Call error on `-XFoo` if `Foo` is not a known extension

shayne-fletcher:
Error out if given `-XFoo` or `-XNoFoo` if `Foo` is not a known extension.

Screen Shot 2020-08-03 at 6 08 27 PM

The simplest possible strategy that maybe fixes https://github.com/ndmitchell/hlint/issues/1093.


ndmitchell:
Slightly worried this will lead to errors at runtime in new and exciting ways for use cases we aren't imagining. But let's find out!

shayne-fletcher:
> Slightly worried this will lead to errors at runtime in new and exciting ways for use cases we aren't imagining. But let's find out!

Indeed, let's see. In the event it does, at least it shouldn't be that much of a job to relax this to a warning.

PR #1097, opened 2020-08-04
Enable refactoring for `f = foo (\x y -> fun x y)` (avoid lambda)

zliu41:
"Avoid lambda" has several cases, this is one of the cases that didn't support refactoring.


ndmitchell:
Thanks!

Issue #1057, opened 2020-06-23
Hints about foldr vs. foldl (and related) when combined with reverse


jvoigtlaender:
I'd like to add the following hints to either the `default` group or the `teaching` group. I'm not sure in which of these two groups they should go, hence opening this as an issue rather than pull request for now:
```yaml
- warn: {lhs: foldr f c (reverse x), rhs: foldl (flip f) c x, name: Use foldl instead of foldr}
- warn: {lhs: foldr1 f (reverse x), rhs: foldl1 (flip f) x, name: Use foldl1 instead of foldr1}
- warn: {lhs: foldl f c (reverse x), rhs: foldr (flip f) c x, note: IncreasesLaziness, name: Use foldr instead of foldl}
- warn: {lhs: foldl1 f (reverse x), rhs: foldr1 (flip f) x, note: IncreasesLaziness, name: Use foldr1 instead of foldl1}
```

ndmitchell:
As a first approximation, no one should be using foldl, but instead using the `foldl'` variant which eliminates the space leak. Does replacing foldl of reverse potentially introduce a space leak? Replacing foldr/reverse with foldl is definitely a good thing to do always. Do we have a rule that suggests foldl to foldl'? Should we?

I think these hints belong in the normal collection - they're useful.

jvoigtlaender:
First off, I hadn't thought about `foldl'` at all, but yes, I see that it is relevant.

As far as I can see, there is no existing rule suggesting `foldl -> foldl`.

There is at least one existing rule that introduces `foldl`: https://github.com/ndmitchell/hlint/blob/fe9c3ad66968840fcd61a9b967ffd9a89ff2970d/data/hlint.yaml#L226

I wonder whether that one, and some of my new proposed rules, should indeed suggest `fold'`. That could be problematic, since for some choices of `f` the use of `foldl'` would be detrimental, in the sense of turning a terminating program into a nonterminating one.

So I guess one question to be answered up front is whether consideration of `foldl` vs. `foldl'` belongs in individual rules that (disregarding space leaks) happen to involve `foldl` or whether it should be handled separately.

ndmitchell:
It seems like `foldl` to `foldl'` should be a separate rule. Therefore, I have two concerns with these rules:

* If we suggest foldl/reverse => foldr are we making it harder for the user to then get the right strictness? A more judgement question, but if technically we potentially eliminate the space leak, that would be good enough to move forward.
[...]

PR #1098, opened 2020-08-06
in teaching-group: replacing fold after reverse by fold in opposite direction

jvoigtlaender:
See https://github.com/ndmitchell/hlint/issues/1057.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -1002,6 +1002,10 @@
     - hint: {lhs: "not (x || y)", rhs: "not x && not y", name: Apply De Morgan law}
     - hint: {lhs: "not (x && y)", rhs: "not x || not y", name: Apply De Morgan law}
     - hint: {lhs: "[ f x | x <- l ]", rhs: map f l}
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks!

jvoigtlaender:
Another suggestion:

The rules
```yaml
    - warn: {lhs: foldr' f c (reverse x), rhs: foldl' (flip f) c x, name: Use left fold instead of right fold}
    - warn: {lhs: foldl' f c (reverse x), rhs: foldr' (flip f) c x, name: Use right fold instead of left fold}
```
could also be worthwhile.

They could be added to the `teaching` group as well. They would only fire if a strict fold is already present, so only students who already know about `foldl'` or `foldr'` would be confronted with them.

But actually these two rules might also usefully go into the `general` group. The concerns in the other discussion should not apply to them, given that we are replacing a strict fold by a strict fold. So no worry that replacing `foldl'/reverse` by non-strict `foldr` could destroy some good space behavior.

ndmitchell:
Definitely happy to add those. My inclination is to start with the teaching group, since I just don't have enough intuition to know if foldl' => foldr' makes a space leak worse, or something else I'm just not familiar enough with them.

PR #1099, opened 2020-08-06
fold/reverse rules also for strict folds

jvoigtlaender:
Follow-up from https://github.com/ndmitchell/hlint/pull/1098#issuecomment-669886700.


ndmitchell:
Thanks!

PR #1101, opened 2020-08-09
Upgrade to ghc-8.10.2

shayne-fletcher:
Upgrade to `ghc-lib-parser-8.10.2.20200808` (ghc-8.10.2) & `ghc-lib-parser-ex-8.10.0.16`.


ndmitchell:
Thanks!

PR #1102, opened 2020-08-12
avoid foldr' on lists

jvoigtlaender:
That's the one thing I perceive as still open from https://github.com/ndmitchell/hlint/issues/1057 at the moment: that using `foldr'` on a right-associative data structure like standard lists is not a good idea, and shouldn't be proposed, not even in a teaching setting.


ndmitchell:
Thanks for following up on it!

PR #1109, opened 2020-08-27
Use an infinite supply of substitution variables

zliu41:
Currently `['a'..'z']` is used in a few places, which is usually sufficient, but using an infinite supply would make it more robust.

### DIFF
--- a/src/GHC/Util/HsExpr.hs
+++ b/src/GHC/Util/HsExpr.hs
@@ -34,7 +34,7 @@ import Data.Generics.Uniplate.DataOnly
 import Data.List.Extra
 import Data.Tuple.Extra
 
-import Refact (toSS)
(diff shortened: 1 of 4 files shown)


ndmitchell:
Thanks! I have a slight concern that as a CAF (top-level binding) the `substVars` will be cached forever. However, in practice we're unlikely to expand that far, and thus its not going to be a real problem.

shayne-fletcher:
> Currently `['a'..'z']` is used in a few places, which is usually sufficient, but using an infinite supply would make it more robust.

Very cool.

zliu41:
> I have a slight concern that as a CAF (top-level binding) the substVars will be cached forever. However, in practice we're unlikely to expand that far, and thus its not going to be a real problem.

Good point; I totally agree!

PR #1110, opened 2020-08-27
Add rules to simplify zipWith and zipWith3

zliu41:


ndmitchell:
Thanks, good catch!

Hmm - I see that we don't suggest `(\x y -> (x, y))` becomes `(,)`. Usually when a rule can be achieved in a few independent steps, and each of those steps is "obviously good", I won't put the composition in as a rule. But here we're not suggesting the reduction of a tuple. That seems like a mistake, so I'd be inclined to add that rule instead. Seem reasonable?

zliu41:
Yeah, that makes total sense.

PR #1113, opened 2020-08-30
Suggest: 'x >>= pure . f' -> 'x <&> f'

### LINKED ISSUE #1112: Ugly refactor for `x >>= pure . f` when x is large
The original isn't great, but I think the suggested refactoring is a little worse. Would it be sensible to restrict this rule from firing when `x` contains newlines or would rely on the parse error rule?

I don't think I've noticed such a content sensitive rule in the lints before, so perhaps it's not in the spirit of the library.

```haskell
runTravWithTravState state traversal =
    unTravT traversal state >>= pure . \case
        Left trav_err                                 -> Left [trav_err]
        Right (v, ts) | hadHardErrors (travErrors ts) -> Left (travErrors ts)
                      | otherwise                     -> Right (v,ts)
```

goes to

```haskell
runTravWithTravState state traversal =
    (\case
        Left trav_err                                 -> Left [trav_err]
        Right (v, ts) | hadHardErrors (travErrors ts) -> Left (travErrors ts)
                      | otherwise                     -> Right (v,ts)) <$> unTravT traversal state
```

expipiplus1:
Closes #1112

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -423,7 +423,7 @@
     - warn: {lhs: return =<< m, rhs: m, name: "Monad law, right identity"}
     - warn: {lhs: liftM, rhs: fmap}
     - warn: {lhs: liftA, rhs: fmap}
-    - hint: {lhs: m >>= return . f, rhs: f <$> m}
+    - hint: {lhs: m >>= return . f, rhs: m Data.Functor.<&> f}
     - hint: {lhs: return . f =<< m, rhs: f <$> m}
     - warn: {lhs: fmap f x >>= g, rhs: x >>= g . f}
     - warn: {lhs: f <$> x >>= g, rhs: x >>= g . f}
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks!

expipiplus1:
Thanks for fixing the tests, Neil!

ndmitchell:
No worries - you happened to get unlucky that the suggestion you are changing also happens to be one that found its way into a bunch of end-to-end tests!

josephcsible:
I'm wondering whether this was really a good change. `<$>` is much more well-known than `<&>` and also doesn't require an import like it does.

ndmitchell:
As a general rule, if a user writes one expression before another in a file, its often a good idea to keep them in that order. Another general rule is to prefer well-known operators. But I often favour the first over the second, because it has less opportunity for screwing up the "feel" of a users code - see #1112 for an example where it did exactly that.

PR #1114, opened 2020-09-02
Enable refactoring for Redundant as-pattern

zliu41:


ndmitchell:
Thanks!

shayne-fletcher:
The force is strong with @zliu41 ! 😉

PR #1120, opened 2020-09-07
Enable refactoring for use section

zliu41:
Added a new `View` instance to extract `RdrName` from `LHsExpr`, so that all cases are handled correctly.

Using the `LHsExpr` directly would lead to extra parentheses (`(+)` vs `+`), while using the string obtained from `rdrNameStr` would lose the module name (`Prelude.+` vs `+`).


ndmitchell:
Thanks, looks awesome!

PR #1121, opened 2020-09-07
Check operator /= "x" in use section

zliu41:
Overlooked this in the previous PR. In the `SectionR` case, if the operator is `x` then we must not use `x` as the substitution var.


ndmitchell:
Thanks!

PR #1129, opened 2020-09-14
Add hints to use const and const id

josephcsible:


data/hlint.yaml
```diff
@@ -296,6 +296,10 @@
 
     - warn: {lhs: \x -> x, rhs: id}
     - warn: {lhs: \x y -> x, rhs: const}
+    - warn: {lhs: curry fst, rhs: const}
+    - warn: {lhs: \x y -> y, rhs: const id}
+    - warn: {lhs: curry snd, rhs: const id}
+    - warn: {lhs: flip const, rhs: const id}
```
googleson78, line 302: Why do you prefer `const id` in these cases? Purely from a cosmetical standpoint it's the same - two identifiers in both cases.

Apart from that, (imo) `const id` is harder to understand (vs `flip const`), because it involves returning a function from `const id x` and applying it again, vs the usual use of `const x y` where you're already "done applying".

josephcsible: Two reasons:
1. They're only equivalent in normal form, but Haskell only reduces to weak head normal form. A concrete consequence of this is that `flip const someBigThing` will require that `someBigThing` remain alive until it has one more argument applied (since its WHNF is `\x -> const x someBigThing`), but `const id someBigThing` will allow it to be collected immediately.
2. In explanations involving lambda calculus, the equivalent of `const id` is almost universally used over `flip const`. For example, Raymond Smullyan's works refer to it as the *kite*, `KI`, not as `CK`.

data/hlint.yaml
```diff
@@ -296,6 +296,10 @@
 
     - warn: {lhs: \x -> x, rhs: id}
     - warn: {lhs: \x y -> x, rhs: const}
+    - warn: {lhs: curry fst, rhs: const}
+    - warn: {lhs: \x y -> y, rhs: const id}
```
ndmitchell, line 300: I find the lambda form more explicit and clearer. I think I'd probably even substitute the other way in my code...

data/hlint.yaml
```diff
@@ -296,6 +296,10 @@
 
     - warn: {lhs: \x -> x, rhs: id}
     - warn: {lhs: \x y -> x, rhs: const}
+    - warn: {lhs: curry fst, rhs: const}
```
ndmitchell, line 299: Definitely a good idea

data/hlint.yaml
ndmitchell: I find the `const id` construction a bit weird. I appreciate its got solid lambda calculus roots, but my brain just fails to grok it that easily. In contrast, `flip const` seems a lot clearer - but I appreciate that it is less efficient.

data/hlint.yaml
josephcsible: I removed this one, since unlike `curry snd` and `flip const`, it is just as efficient as `const id`.

data/hlint.yaml
josephcsible: I think we could theoretically do `rhs: \x y -> y`, which would solve that problem and always be correct and efficient, but since we're making up new bound variables, we might cause variable shadowing warnings to appear in the code. (Although we do have precedent for it: https://github.com/ndmitchell/hlint/blob/a836a7f46700b4d1e511d9bba6d98098240ca538/data/hlint.yaml#L230 makes up an `x`.)

josephcsible: Wait a minute. Although here, we can safely make up a variable, there we can't. I'll make a new PR to fix that.

ndmitchell: We can safely make up variables there, since there is a final check in the code that there are no new free variables. In the LHS `x` is a free variable. If `x` isn't still free in the RHS we don't suggest the hint. Maybe this one might cause issues if `x` was in both `y` and `f`, since it would still be free... ?

My inclination is to always make up the lambda, as `(\_ x -> x)`. People can always rename the `x` if they so wish after.

josephcsible: > My inclination is to always make up the lambda, as `(\_ x -> x)`

Okay, I've made that change.

src/Summary.hs
```diff
@@ -30,7 +30,7 @@ data BuiltinValue = BuiltinValue
 
 
 dedupeBuiltin :: [(BuiltinKey, BuiltinValue)] -> [(BuiltinKey, BuiltinValue)]
-dedupeBuiltin = Map.toAscList . Map.fromListWith (curry snd)
+dedupeBuiltin = Map.toAscList . Map.fromListWith (const id)
```
ndmitchell, line 33: I think I'd prefer the lambda variant here. With `\_ old -> old` as that's more meaningful

ndmitchell:
Thanks! Looks good.

Issue #1128, opened 2020-09-14
Write some pointers for contributors about how to print things


shayne-fletcher:
There's some tribe knowledge about printing parse trees and such among hlint contributors that would be useful to have in the README (as suggested in https://github.com/ndmitchell/hlint/issues/1119#issuecomment-692030793).

ndmitchell:
#1130 covered this pretty well - thanks Shayne!

PR #1130, opened 2020-09-15
Write some notes on how to print syntax

shayne-fletcher:
This goes toward issue https://github.com/ndmitchell/hlint/issues/1128. Add some notes to the "Hacking HLint" section on ways a developer can print parse trees.


ndmitchell:
Thanks!

PR #1138, opened 2020-09-18
YAML error message: show allowed keys

Profpatsch:
If you use `note` instead of `message` in function elements, the error
message looks like:

```
hlint: user error (Failed to read YAML configuration file /foo
  Aeson exception:
Error in $: Error when decoding YAML, Not allowed keys: note
Along path: root 9 functions 9
When at: Object
note: use an explicit match with `error` and reason
name: fromJust
within: []
)
```

leaving you guessing if it’s at all possible to add a note.

If we attach the list of allowed keys, the user at least sees that
there is a `message` field.

Thanks for the pull request!


ndmitchell:
Thanks!

PR #1143, opened 2020-09-23
Also look at data family instances in NewType.hs

### LINKED ISSUE #1142: Detect "Use newtype instead of data" for data family instances
e.g.
```haskell
data family F a

data instance F Int = FInt Int
```
we should suggest `newtype instance` instead here.

googleson78:
Closes #1142

### DIFF
--- a/src/Hint/NewType.hs
+++ b/src/Hint/NewType.hs
@@ -100,7 +100,7 @@ singleSimpleField (L loc (TyClD ext decl@(DataDecl _ _ _ _ dataDef)))
         Just WarnNewtype
               { newDecl = L loc $ TyClD ext decl {tcdDataDefn = dataDef
                   { dd_ND = NewType
-                  , dd_cons = map (\(L consloc x) -> L consloc $ dropConsBang x) $ dd_cons dataDef
+                  , dd_cons = dropBangs dataDef
                   }}
               , insideType = inType
               }
@@ -109,11 +109,14 @@ singleSimpleField (L loc (InstD ext inst@(DataFamInstD instExt (DataFamInstDecl 
         Just WarnNewtype
           { newDecl = L loc $ InstD ext $ DataFamInstD instExt $ DataFamInstDecl $ HsIB hsibExt famEqn {feqn_rhs = dataDef
                   { dd_ND = NewType
-                  , dd_cons = map (\(L consloc x) -> L consloc $ dropConsBang x) $ dd_cons dataDef
+                  , dd_cons = dropBangs dataDef
                   }}
               , insideType = inType
               }
 singleSimpleField _ = Nothing
+
+dropBangs :: HsDataDefn GhcPs -> [LConDecl GhcPs]
+dropBangs = map (\(L consloc x) -> L consloc $ dropConsBang x) . dd_cons
 
 -- | Checks whether its argument is a \"simple\" data definition (see 'singleSimpleField')
 -- returning the type inside its constructor if it is.
--- a/src/Hint/NewType.hs
+++ b/src/Hint/NewType.hs
@@ -115,13 +115,13 @@ singleSimpleField (L loc (InstD ext inst@(DataFamInstD instExt (DataFamInstDecl 
               }
 singleSimpleField _ = Nothing
 
--- | Checks whether its argument is a \"simple\" data definition (see 'singleSimpleFieldNew')
+-- | Checks whether its argument is a \"simple\" data definition (see 'singleSimpleField')
 -- returning the type inside its constructor if it is.
 simpleHsDataDefn :: HsDataDefn GhcPs -> Maybe (HsType GhcPs)
 simpleHsDataDefn dataDef@(HsDataDefn _ DataType _ _ _ [L _ constructor] _) = simpleCons constructor
 simpleHsDataDefn _ = Nothing
 
--- | Checks whether its argument is a \"simple\" constructor (see criteria in 'singleSimpleFieldNew')
+-- | Checks whether its argument is a \"simple\" constructor (see criteria in 'singleSimpleField')
 -- returning the type inside the constructor if it is. This is needed for strictness analysis.
 simpleCons :: ConDecl GhcPs -> Maybe (HsType GhcPs)
 simpleCons (ConDeclH98 _ _ _ [] context (PrefixCon [L _ inType]) _)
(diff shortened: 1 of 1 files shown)


src/Hint/NewType.hs
```diff
@@ -27,6 +27,15 @@ data A = A () -- newtype A = A ()
 newtype Foo = Foo Int deriving (Show, Eq) --
 newtype Foo = Foo { getFoo :: Int } deriving (Show, Eq) --
 newtype Foo = Foo Int deriving stock Show
+data instance Foo Int = Bar Bool -- newtype instance Foo Int = Bar Bool
+data instance Foo Int = Bar {field :: Bool} -- newtype instance Foo Int = Bar {field :: Bool}
+data instance Foo Int = Bar {field :: Int#}
+data instance Foo Int = Bar
+data instance Foo Int = Bar {field1 :: Bool, field2 :: ()}
+newtype instance Foo Int = Bar Bool deriving (Show, Eq) --
```
ndmitchell, line 35: Why the trailing comment? Does that mean anything? Does removing them make it still work? (I note some earlier in the file, can those be deleted too?)

src/Hint/NewType.hs
```diff
@@ -51,9 +60,15 @@ newtypeHintDecl _ = []
...
 newTypeDerivingStrategiesHintDecl decl@(L _ (TyClD _ (DataDecl _ _ _ _ dataDef))) =
-    [ignoreNoSuggestion "Use DerivingStrategies" decl | not $ isData dataDef, not $ hasAllStrategies dataDef]
+    [ignoreNoSuggestion "Use DerivingStrategies" decl | shouldSuggestStrategies dataDef]
+newTypeDerivingStrategiesHintDecl decl@(L _ (InstD _ (DataFamInstD _ (DataFamInstDecl (HsIB _ (FamEqn _ _ _ _ _ dataDef)))))) =
+    [ignoreNoSuggestion "Use DerivingStrategies" decl | shouldSuggestStrategies dataDef]
 newTypeDerivingStrategiesHintDecl _ = []
 
+-- | Determine if the given data definition should use deriving strategies.
+shouldSuggestStrategies :: HsDataDefn GhcPs -> Bool
+shouldSuggestStrategies dataDef = all not [isData dataDef, hasAllStrategies dataDef]
```
ndmitchell, line 70: Can we expand this - `not (isData dataDef) && not (hasAllStrategies dataDef)` - I find `all` on such a short list a bit odd.

src/Hint/NewType.hs
```diff
@@ -80,18 +95,36 @@ data WarnNewtype = WarnNewtype
...
+                  { dd_ND = NewType
+                  , dd_cons = dropBangs dataDef
                   }}
               , insideType = inType
               }
 singleSimpleField _ = Nothing
 
--- | Checks whether its argument is a \"simple constructor\" (see criteria in 'singleSimpleFieldNew')
+dropBangs :: HsDataDefn GhcPs -> [LConDecl GhcPs]
+dropBangs = map (\(L consloc x) -> L consloc $ dropConsBang x) . dd_cons
```
ndmitchell, line 119: Is this equivalent to `(fmap dropConsBang)` in the middle?

ndmitchell (commented):
Nice work! A few very minor details, but should be easy to sort out - let me know if not.

src/Hint/NewType.hs
googleson78: I have no clue, I thought this was some special syntax for the tests that meant something like "a hint should trigger, but it doesn't matter what the output of the hint is"

googleson78: From the `README`:

The general syntax is lhs -- rhs with lhs being the expression you expect to be rewritten as rhs. The absence of rhs means you expect no hints to fire. 

.. but we do want hints to fire here, telling the user to use `DerivingStrategies`, right?

googleson78: The commit from which I copied the test structure, if it's any help:
```diff
tree 4c23545e0e5a82f1a5b4a30563a1330ce268aa7a
parent 98298fd75be1d279bc65eac48e87fd6f3e70b1c0
author Neil Mitchell <ndmitchell@gmail.com> Mon Jul 1 21:05:21 2019 +0100
committer Neil Mitchell <ndmitchell@gmail.com> Mon Jul 1 21:05:21 2019 +0100

Fix up the newtype tests


diff --git a/src/Hint/NewType.hs b/src/Hint/NewType.hs
index 64ad3ffd..ea94419d 100644
--- a/src/Hint/NewType.hs
+++ b/src/Hint/NewType.hs
@@ -19,8 +19,8 @@ data A = A {b :: !C} -- newtype A = A {b :: C}
 data A = A Int#
 {-# LANGUAGE UnboxedTuples #-}; data WithAnn x = WithAnn (# Ann, x #)
 data A = A () -- newtype A = A ()
-newtype Foo = Foo Int deriving (Show, Eq) -- newtype Foo = Foo Int deriving newtype (Show, Eq)
-newtype Foo = Foo { getFoo :: Int } deriving (Show, Eq) -- newtype Foo = Foo { getFoo :: Int } deriving newtype (Show, Eq)
+newtype Foo = Foo Int deriving (Show, Eq) --
+newtype Foo = Foo { getFoo :: Int } deriving (Show, Eq) --
 newtype Foo = Foo Int deriving stock Show
 </TEST>
 -}
```

src/Hint/NewType.hs
googleson78: Yeah, thanks! I didn't know `GenLocated` is a `Functor`

src/Hint/NewType.hs
googleson78: Done

src/Hint/NewType.hs
ndmitchell: That's a pretty useless commit message 😆 - happy enough to let sleeping dogs lie here. I think it may well be saying "there's a hint, i don't care what". Which is a bit poor form, but happy to go with it.

ndmitchell (approved):

src/Hint/NewType.hs
ndmitchell: (Poor form = from me, originally, not complaining at you for copying me!)

ndmitchell:
Thanks, and thanks for the revisions!

PR #1146, opened 2020-09-26
Change redundant return to only highlight return line

### LINKED ISSUE #1141: Redundant return highlights too much
Given:

```
main = do
    print 1
    print 2
    print 3
    return ()
    print 4
```

We get:

```
Sample.hs:(2,8)-(7,11): Warning: Redundant return
Found:
  do print 1
     print 2
     print 3
     return ()
     print 4
Perhaps:
  do print 1
     print 2
     print 3
     print 4

1 hint
```

Given our direction of travel, it would be much simpler just to highlight the return you should remove, rather than the whole do block. Much nicer in an IDE. Much easier to see what is wrong. Harder to see why it's redundant, but in an IDE, that's not a big deal.

Based on feedback from https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4089/diffs#note_300625

rmehri01:
Closes #1141

### DIFF
--- a/src/Hint/Monad.hs
+++ b/src/Hint/Monad.hs
@@ -43,7 +43,7 @@ main = "wait" ~> do f a $ sleep 10
 {-# LANGUAGE BlockArguments #-}; main = print do 17 --
 main = f $ do g a $ sleep 10 --
 main = do f a $ sleep 10 -- @Ignore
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks very much! Looks great.

PR #1149, opened 2020-10-06
Match module wildcards

### LINKED ISSUE #402: Allow module wildcards

```
- modules:
```

tfausak:
Partly addresses some of #402.

I had never looked at the HLint codebase before this morning, so I can't say if these are good changes or not. That being said this turned out to be a relatively easy change. And I added a test case for it.

### DIFF
--- a/src/Apply.hs
+++ b/src/Apply.hs
@@ -20,6 +20,7 @@ import GHC.Hs
 import Language.Haskell.GhclibParserEx.GHC.Hs
 import qualified Data.HashSet as Set
 import Prelude
+import System.FilePattern (FilePattern, (?==))
 
 
 -- | Apply hints to a single file, you may have the contents of the file.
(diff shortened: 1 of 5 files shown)


src/Apply.hs
```diff
@@ -113,5 +114,23 @@ classify xs i = let s = foldl' (f i) (ideaSeverity i) xs in s `seq` i{ideaSeveri
...
+-- True
+-- >>> isSpec "Namespaced.ExampleSpec"
+-- True
+-- >>> isSpec "Deeply.Nested.ExampleSpec"
+-- True
+--
+-- See this issue for details: <https://github.com/ndmitchell/hlint/issues/402>.
+wildcardMatch :: FilePattern -> String -> Bool
+wildcardMatch p m =
+    let f x = if x == '.' then '/' else x
```
ndmitchell, line 135: `let f = replace "." "/"` would be simpler here and let you avoid the `fmap` below.

src/Apply.hs
```diff
@@ -113,5 +114,23 @@ classify xs i = let s = foldl' (f i) (ideaSeverity i) xs in s `seq` i{ideaSeveri
...
+    in '*' `elem` p && fmap f p ?== fmap f m
```
ndmitchell, line 136: Is the `*` test is a "cheap test" to check if it works? If yes, but has no functional difference, a comment would be good. If no, and it has a real semantic difference, what is it?

src/Config/Yaml.hs
```diff
@@ -304,14 +304,18 @@ parseRestrict restrictType v = do
...
-        L _ (HsVar _ (L _ (Qual mod x))) -> pure $ f (moduleNameString mod) (occNameString x)
-        _ -> parseFail v "Bad classification rule"
-    where
-        f mod name@(c:_) | isUpper c = [(mod,name),(mod ++ ['.' | mod /= ""] ++ name, "")]
-        f mod name = [(mod, name)]
+    s <- parseString v
+    if '*' `elem` s
+        then pure [(s, "")]
+        else do
+            x <- parseGHC parseExpGhcWithMode v
```
ndmitchell, line 311: Urk - this whole code is pretty grim. But it was before you got to it, so no complaint about this diff at all!

tests/wildcard.test
```diff
@@ -0,0 +1,7 @@
+---------------------------------------------------------------------
+RUN tests/wildcard.hs --hint=data/wildcard.yaml
+FILE tests/wildcard.hs
+module Spec where
+main = print $ (\ _ -> 'a') ()
+OUTPUT
+No hints
```
ndmitchell, line 7: I'd write this as two tests in the same file - one with `module SpecNo` that still triggers const, That makes sure you didn't just write `False` anywhere but are actually matching the wildcards.

ndmitchell (approved):
Amazing work. A few minor minor comments, but this looks good and a lot simpler than I expected.

src/Apply.hs
tfausak: Yes, it was meant as a cheap test to see if we should even bother trying a wildcard match. However now that I think about it we should be able to combine the regular and wildcard matchers. A regular matcher is just a wildcard that doesn't have any `*`s in it.

tfausak:
I said that this fixes #402, but really it only addresses the `within` part of it. I forgot there was another part about, for example, module aliases:

``` yaml
- modules:
    - { name: 'Biz.Model.*', as: Model }
```

Perhaps another PR could tackle that part of the problem.

ndmitchell:
Thanks, that's great!

PR #1152, opened 2020-10-21
Remove lpaste from readme

Tarmean:
It links to porn now.


ndmitchell:
Thanks! The great march of the internet...

PR #1154, opened 2020-10-24
Document running hlint on changed files

### LINKED ISSUE #1151: Document only running hlint on changed files
Hi, our team at work uses hlint, but it can get pretty slow on larger projects with tens of thousands of lines. We came up with this command to only run hlint on files that differ from the master branch on git. Would you want a PR that adds this tip to the README? Feels potentially out of scope, but would have helped us!

```
{ git diff --diff-filter=d --name-only $(git merge-base HEAD origin/master) -- "***.hs" && git ls-files -o --exclude-standard -- "***.hs"; } | xargs hlint -j=4
```

It's also helpful for refactoring:

```
{ git diff --diff-filter=d --name-only $(git merge-base HEAD origin/master) -- "***.hs" && git ls-files -o --exclude-standard -- "***.hs"; } | xargs -I file hlint -j=4 file --refactor --refactor-options="--inplace --step"
```

MaxGabriel:
Closes #1151

Thanks to my coworker @drostron and @ldubinets for actually coming up with these commands

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,20 @@ HLint enables/disables a set of extensions designed to allow as many files to pa
 
 You can enable extensions by specifying additional command line arguments in [.hlint.yaml](./README.md#customizing-the-hints), e.g.: `- arguments: [-XQuasiQuotes]`.
 
+#### How do I only run hlint on changed files?
+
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks!

PR #1156, opened 2020-10-27
Try to build with a later lts

### LINKED ISSUE #1155: Unicode literals gets flagged
This is not exactly a bug but more of a versioning issue.
GHC extended it's unicode support in `v8.8.3`.

But, hlint with `v8.6.5` keeps flagging a lot of unicode literals. Eg:
The following will be flagged by hlint but gets compiled with ghc`v8.8.3`
```haskell
main = putStrLn "Thinking about what to eat for lunch? 🥦🥬🥗 "
```
The error produced by hlint:
```
Error: Parse error: lexical error in string/character literal at character '\129382'
```

Is there any update when the version bump might take place?

googleson78:
Check if we build with a later lts on windows. If we do, this would be a fix for #1155 by just making a release with the newer lts (tested it locally)

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-14.20 # ghc-8.6.5
+resolver: lts-16.20 # ghc-8.8.4
 packages:
   - .
 extra-deps:


ndmitchell:
Unfortunately, Windows is a disaster zone. The Appveyor will regularly segfault on anything past 8.6.5.

But I think you've also found an issue that other libraries need bumping if you are to upgrade, not just the LTS.

shayne-fletcher:
Is this just an experiment to test the outcome in CI? In case this is on the table, I'm reluctant to land a change that affects stack.yaml in this way. As I understand it, stack.yaml is (1) there (by Neil's sufferance) for HLint developer convenience, doesn't participate in HLint's release and deployment process and (2) sets to that 8.6.5 resolver as a reminder of the lowest version number of GHC that supports building HLint (constrained by the ghc-lib-parser version in effect).

googleson78:
I was mostly testing because I don't have a windows machine on hand, sorry for the confusion!

shayne-fletcher:
> I was mostly testing because I don't have a windows machine on hand, sorry for the confusion!

Not at all. Thank-you for your hard work!

ndmitchell:
It's not really about a reminder of the lowest version - we have Travis jobs for that - most the last version of GHC that works on Windows :(

ndmitchell:
@shayne-fletcher - we are using GHC 8.10.3 successfully for the Windows CI. I suspect upgrading to a GHC 8.10.3 stackage version would let us roll forward with a new LTS.

shayne-fletcher:
> @shayne-fletcher - we are using GHC 8.10.3 successfully for the Windows CI. I suspect upgrading to a GHC 8.10.3 stackage version would let us roll forward with a new LTS.

That shouldn't be a bother. I'll prepare a PR. Stand by.

shayne-fletcher:
> > @shayne-fletcher - we are using GHC 8.10.3 successfully for the Windows CI. I suspect upgrading to a GHC 8.10.3 stackage version would let us roll forward with a new LTS.
> 
> That shouldn't be a bother. I'll prepare a PR. Stand by.

Here it is https://github.com/ndmitchell/hlint/pull/1210.

ndmitchell:
Closed by #1210.

PR #1161, opened 2020-11-09
Expose showANSI for Idea

### LINKED ISSUE #1160: Expose Idea pretty-printing functions (showEX/showANSI)
If you use hlint as a library the main entry point is `hlint` function. However, you're in trouble if you want to react to the resulting `Idea`s somehow. In our use-case, we want to only print the hints for lines touched in the current working git diff.

So far we've been parsing the `--json` output and filtering that against some ranges. Trying to use the library, means that either:

* hlint prints everything (we don't want this)
* hlint prints nothing with --quiet but you have to pretty out the resulting hints

If you just use `show`, it uses the `show = showEx id` instance which means all the colouring is lost. I suppose `showANSI` or even `showEX` functions would be nice to have so that not much of code has to be copy-pasted.

Fuuzetsu:
Fixes #1160

### DIFF
--- a/src/Language/Haskell/HLint.hs
+++ b/src/Language/Haskell/HLint.hs
@@ -15,7 +15,7 @@ module Language.Haskell.HLint(
     -- * Generate hints
     hlint, applyHints,
     -- * Idea data type
-    Idea(..), Severity(..), Note(..), unpackSrcSpan,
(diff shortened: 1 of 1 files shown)


Fuuzetsu:
Seems showEx is Idea-internal: I switched the PR to just `showANSI`.

PR #1163, opened 2020-11-10
Do not suggest "Use lambda" when there are guards

zliu41:
The suggestion for

```haskell
foo = bar (\x -> case x of Y z | z > 0 -> z)
```
is currently

```haskell
\ (Y z) | z > 0 -> z
```

This PR fixes it and it now suggests
```haskell
\case Y z | z > 0 -> z
```


ndmitchell:
Thanks!

PR #1167, opened 2020-11-16
Enable refactoring for "Use lambda"

zliu41:


zliu41:
CI appears to be down

ndmitchell:
The Travis seems to be failing legitimately, with:

```
src/Hint/Lambda.hs:1:1-74: Warning: Avoid restricted extensions
Found:
  {-# LANGUAGE LambdaCase, PatternGuards, TypeApplications, ViewPatterns #-}
```

Is the type application there actually useful? I'm surprised none of the existing subexpressions constrain the type already.

zliu41:
Oops, don't know how I missed that. I changed type application to type annotation.

`noLoc :: HasSrcSpan a => SrcSpanLess a -> a` where `SrcSpanLess` is a non-injective type family, so GHC needs to be told what `a` is.

I also re-generated the summary, which seems to have some unrelated changes.

ndmitchell:
Thanks! We need to build up some CI tooling around generating the summary, but I don't quite no what yet, and its likely we'll have to move CI entirely, so no point doing it now.

Issue #769, opened 2019-09-19
Is there a hint for unqualified imports


phadej:
I'd like to prohibit unqualified imports without explicit import list:

```haskell
import Data.List                 -- bad
import qualified Data.List as L  -- ok
import Data.List (foldl')        -- ok
```

This should be configurable, e.g. to allow some modules to be imported unqualified:

```haskell
import Prelude hiding (null)
-- or
import Prelude.Compat -- should be configurable to be ok
```

ndmitchell:
See https://github.com/ndmitchell/hlint#restricting-items for whats there now. I don't quite know the boundaries of its power, but I'm fine with patches making it do the things you describe.

phadej:
@ndmitchell thanks for prompt reply. I actually use all functionality from that section, and it doesn't mention *unqualified imports*. I'll see if I can hack that together.

phadej:
Bummer. Implicitly listing `Prelude` is fine, but implicitly listing all package own modules is cumbersome. I need to think, how to not make this a configuration misery.

ndmitchell:
There is a ticket to make those patterns take wildcards. If that happened then you might expect a users package to mostly share a common prefix, so you could whitelist `Development.Shake.**` and that would do it.

phadej:
I'd like to share the same `.hlint,yaml` for all my projects. Essentially, I'd like HLint to be able to get information from `.cabal` files about what modules are local and which aren't. But that seems to be too big a change.

danbroooks:
I'd really like this feature, even if you have to define a namespace in hlint, so something like:

`MyApp.QualifiedModules.*` would be great

ndmitchell:
Feel free to go for it. I imagine you need to add something to https://github.com/ndmitchell/hlint/blob/master/src/Hint/Restrict.hs and https://github.com/ndmitchell/hlint/blob/cf972fedcc6e31e94404cbcda6dfbb052b6b7659/src/Config/Type.hs#L28. You parse the config somehow, turn it into a RestrictedItem structure, then apply that. I don't have any particularly good idea how this might work, as personally I wouldn't be interested in using it. I'd first prototype possible .yaml file syntaxes, and their semantics, and then go from there.

danbroooks:
I had a bit of a play around and I have something working, but like you say the key to this now is how to get it so the configuration is nice...

What I have so far uses a new field on `modules` called `qualifiedStyle`, so for example the config:

```
[...]

PR #1168, opened 2020-11-16
Refactor findFunction

danbroooks:
While doing some initial experimenting for https://github.com/ndmitchell/hlint/issues/769 I ended up refactoring this function, I thought the end result was worth submitting, as it makes use of `do` notation for handling the `Maybe`, and gets rid of a duplicate name that was a bit confusing when initially reading the function.


ndmitchell:
Thanks!

PR #1169, opened 2020-11-18
replace 0 cmdThreads with getNumProcessors in earlier so it's used everywhere

dten:
fixes `-j` not actually doing any work in parallel (except gc)

noticed this because `--threads=8` performed very different than `-j` but `getNumProcessors` returns 8 for me

on my other machine with more cores
before: 
`hlint . -j` 3m54s
`hlint --threads=24` 29s

after this -j is also ~29s

### DIFF
--- a/src/HLint.hs
+++ b/src/HLint.hs
@@ -134,6 +134,8 @@ readAllSettings args1 cmd@CmdMain{..} = do
     cmd@CmdMain{..} <- if null args2 then pure cmd else getCmd $ args2 ++ args1 -- command line arguments are passed last
     settings2 <- concatMapM (fmap snd . computeSettings (cmdParseFlags cmd)) cmdFindHints
     let settings3 = [SettingClassify $ Classify Ignore x "" "" | x <- cmdIgnore]
+    cmdThreads <- if cmdThreads == 0 then getNumProcessors else pure cmdThreads
+    cmd <- pure CmdMain {..}
     pure (cmd, settings1 ++ settings2 ++ settings3)
     where
         enableGroup groupName =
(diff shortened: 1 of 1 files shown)


ndmitchell:
Oops! Great spot!

I think the real fix is to change https://github.com/ndmitchell/hlint/blob/466ed41c491eff616281adc46b104b6045e6c8c8/src/HLint.hs#L148 to add the line  to become:

```
    cmdThreads <- if cmdThreads == 0 then getNumProcessors else pure cmdThreads
    cmd <- pure CmdMain{..}
```

Or alternatively, to push the defaulting of cmdThreads up to https://github.com/ndmitchell/hlint/blob/466ed41c491eff616281adc46b104b6045e6c8c8/src/CmdLine.hs.

dten:
how's this :) not quite either of your suggestions 😅

dten:
mmm also a high thread count actually ends up performing worse than `--threads=2` due to the parallel gc being too eager. I'm using `+RTS -A64m` which makes parallel gc behave much better

ndmitchell:
My experience is that parallel GC is garbage - its always slower than serial GC. I'm a fan of the RTS flags "-with-rtsopts=-I0 -qg" - does that help (probably wants adding https://github.com/ndmitchell/hlint/blob/master/hlint.cabal#L170)

ndmitchell:
Thanks, great work! Keen to see the effect of `-qg`, and if that helps further, we can reduce times even more

ndmitchell:
Oh, and as for the code, it looks like where you put it is a bit better than either of my suggestions!

dten:
I think `-I0` should make no difference here cause hlint doesn't idle?

fresh results run back to back on 4core ht
`-A64m` 44s with 587% cpu utilisation (according to gnu time)
`-I0 -qg`  1m52s with 239% cpu util
`-I0 -qg -A64m`  54s with 443% cpu util
`-A64m` 44s with 584% cpu util (ran again to check fluke)

ndmitchell:
So sounds like adding -A32 is a good idea? And parallel GC is now not as harmful as it used to be.

ndmitchell:
I pushed a change to move us to -A32m.

PR #1171, opened 2020-11-22
Do not refactor "Redundant lambda" if the body is HsCase

zliu41:
There are some difficulties, described in https://github.com/alanz/ghc-exactprint/issues/97, in refactoring redundant lambda hints where the body of the lambda is an `HsCase` and the "case" is in the same line as the lambda, for example
```haskell
f = \x -> case scrut of
  pat -> body
```
This may be fixable from ghc-exactprint, but I'm not familiar enough with ghc-exactprint to say how easy it is. So the best thing to do for now seems to be disabling the refactoring if the body is `HsCase`.

### DIFF
--- a/src/Hint/Lambda.hs
+++ b/src/Hint/Lambda.hs
@@ -144,6 +144,7 @@ lambdaDecl
           gen :: [LPat GhcPs] -> LHsExpr GhcPs -> LHsDecl GhcPs
           gen ps = uncurry reform . fromLambda . lambda ps
           refacts = case newBody of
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks, looks reasonable. Did you run the test suite locally? CI at this point is basically both garbage (I'm actively working on moving to GitHub Actions as we speak!) so I'll trust you.

ndmitchell:
One CI branch went green, so I'm assuming the others will.

zliu41:
Yes I ran the tests locally :slightly_smiling_face:

Issue #1159, opened 2020-11-05
Warn on redundant parentheses in (foo . bar) %~ (hello . world)


yairchu:
I was surprised to see that `hlint` doesn't suggest to turn `(foo . bar) %~ (hello . world)` into `foo . bar %~ hello . world`.

I don't know why it doesn't, from reading the code it seems that it should be aware of the fixites of both `%~` and `.`

yairchu:
Checked if this is a regression, but it doesn't seem so. When lens fixities were added in d3d03781b89a5467c7e68560c7649924e5c212e7 it also didn't warn on this

ndmitchell:
HLint deliberately never suggests removing brackets where only fixities mean they aren't required. Taking an easier example, given `a + b * c` I consider whether to put in brackets as `a +(b * c)` as a personal choice, and not something that HLint should either require or remove. It's possible to write a rule to do it if you believe the are specific cases where they shouldn't be used - matching on brackets works, albeit can be a bit tricky to get right because often they are dealt with implicitly.

yairchu:
Oh. What do you think about an option to suggest removing those brackets? I have a feeling that many folks would like such an option (me being one of them).

ndmitchell:
Sure, go for it. Although be aware that HLint sometimes (regularly?) approximates the fixities, and has a few heuristics in a few places to avoid applying "correct" rules because it has low confidence in the parse tree. That might mean the hint is fairly unreliable.

yairchu:
I have an incomplete implementation at https://github.com/ndmitchell/hlint/compare/master...yairchu:brackets which I now use :)

In my branch it's currently always enabled, I guess that to make it optional
ideally it would act as if it's a different type of warning, maybe "Redundant bracket - pedantic"?
Then you could enable/disable the specific hint?
That would probably require `needBracket` to return which type of bracket is necessary (ie `NoBracket | OptionalBracket | NeedBracket`)

ndmitchell:
Looks super cool. Yep, a different name is essential - "Redundant brackets due to fixity"? "Redundant brackets in infix expression"?

I'd suggest creating a separate needBracketFixities - the functions don't overlap that much, so I'd call needBracket to emit a redundant bracket warning, and if that doesn't fire, call needBracketFixities and see if you omit one.

yairchu:
Made it a separate, ignored by default, hint now! Will make a PR after https://github.com/ndmitchell/hlint/pull/1173 which prepares for it

[...]

PR #1174, opened 2020-11-24
Remove redundant brackets #1159

yairchu:
A new "Fixities" hint, yielding ignored by default "Redundant infix bracket due to operator fixities" ideas.

### DIFF
--- a/src/Hint/Fixities.hs
+++ b/src/Hint/Fixities.hs
@@ -1,0 +1,79 @@
+{-
+
+Raise a warning if you have redundant brackets in nested infix expressions.
+
+<TEST>
+yes = 1 + (2 * 3) -- @Ignore 1 + 2 * 3
+yes = (2 * 3) + 1 -- @Ignore 2 * 3 + 1
+no = (1 + 2) * 3
+no = 3 * (1 + 2)
+no = 1 + 2 * 3
+no = 2 * 3 + 1
+yes = (a >>= f) >>= g -- @Ignore a >>= f >>= g
+no = (a >>= \x -> b) >>= g
+</TEST>
+-}
+
+module Hint.Fixities(fixitiesHint) where
+
+import Hint.Type(DeclHint,Idea(..),rawIdea,toSS)
+import Config.Type
+import Control.Monad
+import Data.List.Extra
+import Data.Map
+import Data.Generics.Uniplate.DataOnly
+import Refact.Types
+
+import BasicTypes (compareFixity)
+import Fixity
+import GHC.Hs
+import GHC.Util
+import Language.Haskell.GhclibParserEx.GHC.Utils.Outputable
+import OccName
+import SrcLoc
+import RdrName
+
+fixitiesHint :: [Setting] -> DeclHint
+fixitiesHint settings _ _ x =
+  concatMap (infixBracket fixities) (childrenBi x :: [LHsExpr GhcPs])
+   where
+     fixities = foldMap getFixity settings `mappend` fromList (toFixity <$> defaultFixities)
+     getFixity (Infix x) = uncurry singleton (toFixity x)
+     getFixity _ = mempty
+
+infixBracket :: Map String Fixity -> LHsExpr GhcPs -> [Idea]
+infixBracket fixities = f Nothing
+  where
+    msg = "Redundant bracket due to operator fixities"
+    f p o = cur p o <> concat [f (Just (i, o, gen)) x | (i, (x, gen)) <- zipFrom 0 $ holes o]
+    cur p v = do
+      Just (i, o, gen) <- [p]
+      Just x <- [remParen v]
+      guard $ redundantInfixBracket fixities i o x
+      pure $
+        rawIdea Ignore msg (getLoc v) (unsafePrettyPrint o)
+        (Just (unsafePrettyPrint (gen x))) [] [Replace (findType v) (toSS v) [("x", toSS x)] "x"]
+
+redundantInfixBracket :: Map String Fixity -> Int -> LHsExpr GhcPs -> LHsExpr GhcPs -> Bool
+redundantInfixBracket fixities i parent child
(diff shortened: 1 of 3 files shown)


src/GHC/Util/Brackets.hs
```diff
@@ -18,6 +21,9 @@ class Brackets a where
   -- | Is the child safe free from brackets in the parent
   -- position. Err on the side of caution, True = don't know.
   needBracket :: Int -> a -> a -> Bool
+  -- | Is a bracket for the child redundant due to fixities
+  redundantInfixBracket :: Map String Fixity -> Int -> a -> a -> Bool
```
ndmitchell, line 25: Why make this a part of the Brackets class given its only used for expressions. I'd suggest moving it to Hint.Fixity, specialising to Expr, and leaving the Brackets module alone.

src/Hint/Fixities.hs
```diff
@@ -0,0 +1,56 @@
...
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE TypeFamilies #-}
+{-
+
+Raise a warning if you have redundant brackets in nested infix expressions.
+
+<TEST>
+yes = 1 + (2 * 3) -- @Ignore 1 + 2 * 3
+no = (1 + 2) * 3
+no = 1 + 2 * 3
```
ndmitchell, line 11: Could we have another test of brackets in the other direction, e.g. `1 * 2 + 3`. A lot of the fixity stuff is left vs right, so an easy potential mistake.

src/Hint/Fixities.hs
```diff
@@ -0,0 +1,56 @@
...
+import GHC.Hs
+import Outputable
+import SrcLoc
+import GHC.Util
+import Language.Haskell.GhclibParserEx.GHC.Utils.Outputable
+
+fixitiesHint :: [Setting] -> DeclHint
+fixitiesHint settings _ _ x =
+  concatMap (infixBracket fixities) (childrenBi x :: [LHsExpr GhcPs]) ++
+  concatMap (infixBracket fixities) (childrenBi x :: [LHsType GhcPs]) ++
```
ndmitchell, line 35: Why bother doing this given we know fixity hints only apply to expressions?

src/Hint/Fixities.hs
```diff
@@ -0,0 +1,56 @@
...
+  concatMap (infixBracket fixities) (childrenBi x :: [LPat GhcPs])
+   where
+     fixities = foldMap getFixity settings `mappend` fromList (map toFixity defaultFixities)
+     getFixity (Infix x) = uncurry singleton (toFixity x)
+     getFixity _ = mempty
+
+infixBracket :: forall a . (Data a, HasSrcSpan a, Outputable a, Brackets a) => Map String Fixity -> a -> [Idea]
```
ndmitchell, line 42: I think this could be just `HsExpr` rather than `a`, which simplifies everything.

ndmitchell (commented):
Looks good! Some CI warnings you should fix, and I'd simplify this to being expression only rather than generic over patterns/types, but with that I'll be happy.

src/GHC/Util/Brackets.hs
yairchu: Good question. What I had in mind is that perhaps type operators have fixities too?

ndmitchell: So do patterns. But I'm not convinced that suggesting more/less fixity based brackets in either of those is a good idea ever. With expressions, everyone pretty much has a good sense that multiplication binds tighter than addition, so its plausible. With types and patterns, there aren't any universally understood fixities, so I think the case is a lot weaker.

yairchu: Alright, removed the polymorphism and it's indeed much simpler now :)

src/Hint/Fixities.hs
yairchu: Done

src/Hint/Fixities.hs
yairchu: Done

src/Hint/Fixities.hs
yairchu: Done

src/Hint/Fixities.hs
```diff
@@ -0,0 +1,79 @@
...
+no = 1 + 2 * 3
+no = 2 * 3 + 1
+yes = (a >>= f) >>= g -- @Ignore a >>= f >>= g
+no = (a >>= \x -> b) >>= g
+</TEST>
+-}
+
+module Hint.Fixities(fixitiesHint) where
+
+import Hint.Type(DeclHint,Idea(..),rawIdea,Severity(..),toSS)
```
ndmitchell, line 19: GHC says the import of Severity here is unused. Other than that, looks good to go.

yairchu: Apparently it's also being imported from `Config.Type` and I'm not getting the warning because I'm on an older GHC. Done

src/Hint/Fixities.hs
```diff
@@ -0,0 +1,79 @@
...
+  concatMap (infixBracket fixities) (childrenBi x :: [LHsExpr GhcPs])
+   where
+     fixities = foldMap getFixity settings `mappend` fromList (toFixity <$> defaultFixities)
+     getFixity (Infix x) = uncurry singleton (toFixity x)
+     getFixity _ = mempty
+
+infixBracket :: Map String Fixity -> LHsExpr GhcPs -> [Idea]
+infixBracket fixities = f Nothing
+  where
+    msg = "Redundant infix bracket due to operator fixities"
```
ndmitchell, line 47: Final change, can we shorten this a  bit - `Redundant bracket due to operator fixities` seems sufficient.

yairchu: Done :)

ndmitchell:
Thanks!

PR #1176, opened 2020-12-02
Suggest use of unzip

zliu41:


ndmitchell:
I assumed that a function in the NonEmpty module had something to do with NonEmpty. But turns out not! That's super surprising, to the extend I'm a little worried that suggesting it would really confuse people.

zliu41:
Yeah, I agree. One would expect that `unzip` to live elsewhere. I updated the hint to be the list version.

ndmitchell:
Thanks!

PR #1180, opened 2020-12-13
compare `on` a -> comparing a

siraben:
``compare `on` a ``-> `comparing a `

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -99,6 +99,7 @@
     - warn: {lhs: compare x y == GT, rhs: x > y}
     - warn: {lhs: compare x y == EQ, rhs: x == y}
     - warn: {lhs: compare x y /= EQ, rhs: x /= y}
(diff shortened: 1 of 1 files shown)


siraben:
Hm, can't seem to make the CI happy.

siraben:
Can't seem to make the CI happy with a test case involving two different rewrites: ` last (sortBy (compare `on` fst) xs)`.

ndmitchell:
Thanks for the patch. This used to be a hint (introduced in f0b076ba0f4f4f8c979416e9e8dca924b21dd279), but got removed in d9440e29b961f607168c09206bec3cd449873cb6 because it turned out to be a hint that a reasonable number of people disabled. The problem is that `comparing` is very much a special case, whereas learning the `on` function has a lot more value to it.

siraben:
Aw, I see. Thanks!

PR #1188, opened 2020-12-20
Upgrade to ghc-lib-parser-8.10.3, ghc-lib-parser-ex-8.10.0.17

shayne-fletcher:
Book-keeping basically. This leaves us safe in the knowledge that new installs of hlint are linking `ghc-lib-parser-20201220` and the verified compatible `ghc-lib-parser-ex-8.10.0.17`.

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -80,7 +80,7 @@ library
       build-depends:
           ghc-lib-parser == 8.10.*
     build-depends:
-        ghc-lib-parser-ex >= 8.10.0.16 && < 8.10.1
+        ghc-lib-parser-ex >= 8.10.0.17 && < 8.10.1
 
     if flag(gpl)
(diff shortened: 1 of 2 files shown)


shayne-fletcher:
Ooh! Look at this fancy new CI! 😀

ndmitchell:
I think our fancy new CI ignores stack.yaml entirely now. I wonder if we should delete it, or does it still bring value to you?

shayne-fletcher:
> I think our fancy new CI ignores stack.yaml entirely now. I wonder if we should delete it, or does it still bring value to you?

Please, if not deleting it is an option, let us keep it. It's integral to my workflow!

ndmitchell:
Not deleting it is absolutely fine. It can stay as long as you find it useful.

PR #1193, opened 2020-12-28
Add warnings for redundant flip

josephcsible:


josephcsible:
I was on the fence about suggesting `g >>= flip f` instead be changed to `g <**> f` to keep things in order, but `<**>` requires `import Control.Applicative`. Do you think it's worth doing anyway?

ndmitchell:
I think keeping the terms in order would be better if we can, so I think suggesting `<**>` with a Control.Applicative import is the right thing to do. Everything else looks very nice, so will merge with that update.

josephcsible:
Done.

ndmitchell:
Thanks!

PR #1202, opened 2021-01-15
Add missing parentheses for Avoid Lambda

zliu41:
Reported originally by @poscat0x04 in https://github.com/mpickering/apply-refact/issues/102.

Minimum example: `x = f (\a b -> c d b a)`, should suggest `flip (c d)`, not `flip c d`.

### DIFF
--- a/src/GHC/Util/HsExpr.hs
+++ b/src/GHC/Util/HsExpr.hs
@@ -228,6 +228,7 @@ niceLambdaR [x, y] (view -> App2 op (view -> Var_ y1) (view -> Var_ x1))
       )
   where
     gen = noLoc . HsApp noExtField (strToVar "flip")
+        . case op of L _ HsVar{} -> id; _ -> noLoc . HsPar noExtField
 
 -- We're done factoring, but have no variables left, so we shouldn't make a lambda.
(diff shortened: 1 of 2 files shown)


src/GHC/Util/HsExpr.hs
```diff
@@ -228,6 +228,7 @@ niceLambdaR [x, y] (view -> App2 op (view -> Var_ y1) (view -> Var_ x1))
       )
   where
     gen = noLoc . HsApp noExtField (strToVar "flip")
+        . case op of L _ HsVar{} -> id; _ -> noLoc . HsPar noExtField
```
ndmitchell, line 231:
```suggestion
        . if isAtom op then id else addParen
```

Is this the essence of what you were trying to do with the HsVar test?

ndmitchell (commented):
Good catch! Curious if it should be generalised first?

zliu41:
Updated! By generalise, if you mean whether this is applicable to other hints, that doesn't seem to be the case.

ndmitchell:
By generalised, I meant insert brackets in fewer places with isAtom, exactly what you did - thanks!. Sorry, that was massively unclear!

Issue #1203, opened 2021-01-15
alias rules


yairchu:
Alias rules don't trigger hints on their own, when applying them enables other rules a hint should trigger:

    - alias: {lhs: a =<< b, rhs: b >>= a}
    - alias: {lhs: a <=< b, rhs: b >=> a}
    - alias: {lhs: a <$ b, rhs: b $> a}
    - alias: {lhs: a & b, rhs: b $ a}
    - alias: {lhs: a <&> b, rhs: b <$> a}

Bonus feature: `style` rules which only apply on the result of hints to use a certain style (i.e use `<&>` rather than `<$>` in a certain project)

ndmitchell:
The problem is that matching is pretty expensive, and at the moment successful matches (which are the most expensive) always result in hints, which are valuable. If we have to apply some rewrites, that could get very expensive quickly. However, most of your rules are pretty simple, some operator reorders the parameters, so it might be possible to define something simpler that can only do that. It should also do pure/return etc to be more applicable.

yairchu:
Expanding on this, I guess such a mechanism could "normalise" code, and rules could be written for "normalised" code (or alternatively get normalised too) so that a rule for `pure` would also match for `return`.

With this in mind, perhaps the technique of duplicating rules to have dotted form could be replaced normalising converting code to dotted form, ie `concat (map f x)` would be normalised to `(concat . map f) x` before checking the rule rather than checking multiple rules?

ndmitchell:
I'm sceptical that you could normalise so dot didn't matter. Would you normalise `a . b . c` to `(a . b) . c` or `a . (b . c)`? Both are valid. Both might match.

Issue #1201, opened 2021-01-13
Support for "&" operator (similar to "$")


yairchu:
hlint knows how to fire rules also when function applications are done indirectly via "$" to reduce parentheses. Since `lens` came along the `&` operator gain popularity, and similar behavior for it will be valuable. I'll gladly implement this given some pointers on where to look :)

ndmitchell:
https://github.com/ndmitchell/hlint/blob/master/src/Hint/Match.hs is the place to look - that has a bunch of comments on $ in various places. But perhaps first you could give some examples you see that you expect to come up? E.g. `concat $ map f x` comes up all the time. Does `map f x & concat` come up similarly?

yairchu:
Someone may write `thing ^? somePrism <&> someFunc & fromMaybe someDefault`, and it would be nice to suggest `maybe someDefault someFunc (thing ^? somePrism)` just as would had happened with `fromMaybe someDefault $ someFunc <$> thing ^? somePrism`

ndmitchell:
Hints which change the order of terms tend to be less useful - if you write `a b c`, sometimes the fact they are in that order is also a meaningful choice. I wonder if `&` by necessity involves changing the order when making suggestions?

yairchu:
As in https://github.com/lamdu/lamdu/commit/621d48a20a65a8b50ad558682991291e4441380a, one would probably end up not changing the order. But even if the hint does change the order it is still useful info to guide towards the end goal. With extra work I suppose the hints could even preserve the order but that's probably much more difficult and also less important.

ndmitchell:
OK, I think I see what the reason to do it would be. Give it a whirl, and let's see how it performs in practice.

yairchu:
> We also see through (.) by translating rules that have (.) equivalents
to separate rules. For example:

Ok, iiuc, this is done by `readRule` invoking `dotVersion` to generate these equivalent rules, right? But as for

> We see through ($)/(.) by expanding it if nothing else matches.

Where is this part performed? It that what `extendInstances` is for?

ndmitchell:
Oh, I think the `$` might actually be performed at https://github.com/ndmitchell/hlint/blob/master/src/GHC/Util/Unify.hs#L174 - back when that comment was written they were probably in the same file :)

ndmitchell:
This is done now with #1204.

PR #1204, opened 2021-01-15
Support for '&' operator (similar to '$') #1201

yairchu:
This appears to work, triggers 3 hints for Lamdu and they appear to be correct :)
I suppose we may also want to add tests for these, but I also suggested #1203 which might be a different solution to consider.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -1047,6 +1047,7 @@
 # yes = f x where f x = concat . map head -- concatMap head
 # yes = concat . map f . g -- concatMap f . g
 # yes = concat $ map f x -- concatMap f x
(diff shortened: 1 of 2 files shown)


ndmitchell:
Looks good. Could we get some tests? Probably in data/hlint.yaml, something like `map f x & concat` suggests `concatMap f x`. Doesn't need to be much, just something to suggest it works.

yairchu:
> something like `map f x & concat` suggests `concatMap f x`

Does, added this one and another test

ndmitchell:
Thanks! That second test you wrote of zipWithM is something I'd have never figured out by eye, and the change is very simple, so any doubts I may have originally had about supporting `&` are certainly alleviated. Nice work!

PR #1205, opened 2021-01-17
Generalize hints which were with '&' form

yairchu:
Without the '&' the hints apply in more cases,
and since 3.2.7 hlint can see through '&' so it isn't needed in the hints


ndmitchell:
Great idea! Could we get a test for one of these hints (don't mind which) that it still applies with the `&` variant? Keen that we don't inadvertently regress these great hints you've added previously.

yairchu:
Good idea :) Done

ndmitchell:
Thanks!

PR #1206, opened 2021-01-18
Require apply-refact 0.9.0.0

### LINKED ISSUE #1179: Removing brackets may need to insert spaces
Consider:

```haskell
x = do(y)
```

Hlint currently removes the brackets leaving 

```haskell
x doy
```

Seems to happen for all keywords

zliu41:
Also mention that apply-refact can be installed from Cabal or Stack, which hopefully makes the message clearer.

Fixes #1179.


ndmitchell:
Thanks!

PR #1209, opened 2021-02-05
Update to ghc-9.0.1

shayne-fletcher:
Upgrade HLint to use the ghc-9.0.1 parse tree.


shayne-fletcher:
The Ubuntu 8.6 failure is expected. The minimum GHC now required to build HLint is 8.8.1.

src/GHC/Util/SrcLoc.hs
```diff
@@ -5,19 +5,48 @@ module GHC.Util.SrcLoc (
...
+  deriving (Outputable, Eq)
 instance Default SrcSpanD where def = SrcSpanD noSrcSpan
+-- SrcSpan no longer provides 'Ord' so we are forced to roll our own.
+compareSrcSpans (SrcSpanD a) (SrcSpanD b) =
+  case a of
+    RealSrcSpan a1 _ ->
+      case b of
+        RealSrcSpan b1 _ ->
+          a1 `compareRealSrcSpans` b1
+        _ -> error "'Duplicate.hs' invariant error: can't compare unhelpful spans"
```
ndmitchell, line 34: This seems super dangerous. Could we say all unhelpful ones are equal? It screws up stuff, but in a less dangerous way.

src/GHC/Util/SrcLoc.hs
```diff
@@ -5,19 +5,48 @@ module GHC.Util.SrcLoc (
...
+    _ -> error "'Duplicate.hs' invariant error: can't compare unhelpful spans"
+compareRealSrcSpans a b =
+  let (a1, a2, a3, a4, a5) = (srcSpanFile a, srcSpanStartLine a, srcSpanStartCol a, srcSpanEndLine a, srcSpanEndCol a)
+      (b1, b2, b3, b4, b5) = (srcSpanFile b, srcSpanStartLine b, srcSpanStartCol b, srcSpanEndLine b, srcSpanEndCol b)
+  in case compare a1 b1 of
```
ndmitchell, line 39: Just do `compare (a1, a2, a3, a4, a5) (b1, b2, b3, b4, b5)` - or don't break them into tuples above.

ndmitchell (commented):
Very nice! A few little little things to tidy up, but amazing work.

Would you like me to sneak a patch in removing 8.6 support now, before this? Or wait til after?

src/GHC/Util/SrcLoc.hs
shayne-fletcher: Yes we can.

src/GHC/Util/SrcLoc.hs
shayne-fletcher: Yep.

shayne-fletcher:
> Very nice! A few little little things to tidy up, but amazing work.

Thanks!
 
> Would you like me to sneak a patch in removing 8.6 support now, before this?

Sure! If you do that, I can rebase on that and we'll land nice and square!

ndmitchell:
Just pushed something removing 8.6, so rebase on that and you should be good.

I note that the Stack.yaml is now out of date.

shayne-fletcher:
Thank-you very much, I think we are good to go.

ndmitchell:
Thanks! I think the next release deserves to be 0.1 higher, to reflect the fact we have changed the GHC version and ghc-lib flags. I wonder if we should hold off until GHC 9.0 is released? Or move now?

shayne-fletcher:
> Thanks! I think the next release deserves to be 0.1 higher, to reflect the fact we have changed the GHC version and ghc-lib flags. I wonder if we should hold off until GHC 9.0 is released? Or move now?

9.0.1 was released Thursday (https://www.haskell.org/ghc/blog/20210204-ghc-9.0.1-released.html), we should move now.

ndmitchell:
It's not yet on the CI, so my inclination is to wait a few days until it is, then test, and then release: https://github.com/haskell/actions/pull/33

shayne-fletcher:
> It's not yet on the CI, so my inclination is to wait a few days until it is, then test, and then release: [haskell/actions#33](https://github.com/haskell/actions/pull/33)

Sounds good!

PR #1213, opened 2021-02-07
Bump to ghc-lib-parser-ex-9.0.0.3

shayne-fletcher:
This updates `hlint.cabal` and `stack.yaml` to require `ghc-lib-parser-ex-9.0.0.3`. This version of `ghc-lib-parser-ex` fixes some Cabal config mistakes present in older versions. Thank-you for your patience and sorry about this.


ndmitchell:
Thanks!

Issue #1194, opened 2020-12-30
Wildcards for TemplateHaskellQuotes


yairchu:
Suppose one would like a rule that will suggest replacing `varE 'mempty` with `[|mempty|]`, or in a more general fashion:

    - hint: {lhs: "varE 'a", rhs: "[|a|]"}

This doesn't work currently because `a` in the rule above isn't treated as a wildcard.

ndmitchell:
I think its possible - I'm just not convinced how many rules of that form people are likely to want? If it's only one or two, they could be hardcoded as built in rules (if that's a general substitution property? I have no idea).

yairchu:
It would probably be a bunch of rules, at the very least:

```
- hint: {lhs: "varE 'a", rhs: "[|a|]"}
- hint: {lhs: "pure (VarE 'a)", rhs: "[|a|]"}
- hint: {lhs: "conE ''a", rhs: "[|a|]"}
- hint: {lhs: "pure (ConE ''a)", rhs: "[|a|]"}
- hint: {lhs: "conT ''a", rhs: "[t|a|]"}
- hint: {lhs: "pure (ConT ''a)", rhs: "[t|a|]"}
- hint: {lhs: "conP ''a", rhs: "[p|a|]"}
- hint: {lhs: "pure (ConP ''a)", rhs: "[p|a|]"}

- hint: {lhs: "varE ''a `appE` b": res: "[|a b|]"}
- hint: {lhs: "conE ''a `appE` b": res: "[|a b|]"}
- hint: {lhs: "conT ''a `appT` b": res: "[t|a b|]"}
- hint: {lhs: "conP ''a `appP` b": res: "[p|a b|]"}
```

This could serve as a useful teaching tool pointing people to write things more easily using the `[||]` Template Haskell syntax.

ndmitchell:
I'm not averse to a PR - I expect its just a few additional rules in the matching engine and wouldn't add much complexity.

yairchu:
Do you have a recommended method to quickly see which `LHsExpr GhcPs` a given expression becomes? (from just reviewing [the constructors of `HsExpr`](https://hackage.haskell.org/package/ghc-lib-parser-8.10.3.20201220/docs/GHC-Hs-Expr.html#t:HsExpr) I can't come with a good guess)

ndmitchell:
[...]

PR #1215, opened 2021-02-14
Allow matching for TemplateHaskellQuotes #1194 and add some TH rules

yairchu:
The `'var` and `''TypeCon` syntaxes are now supported for rule left-hand-sides.

This covers some basic needs of TH rules.

More complete support (more rules) would also need:

* Also support matching on `[|code|]` syntax on LHS
* Support parsing `[|$var|]` syntax on RHS
* Bonus: support `'var` syntax also on RHS

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -36,6 +36,7 @@
     - import Maybe as Data.Maybe
     - import Monad as Control.Monad
     - import Char as Data.Char
+    - import Language.Haskell.TH as TH
 
 - package:
     name: lens
@@ -811,6 +812,10 @@
     - warn: {lhs: "Data.Map.Lazy.fromList []", rhs: Data.Map.Lazy.empty}
     - warn: {lhs: "Data.Map.Strict.fromList []", rhs: Data.Map.Strict.empty}
 
+    # TEMPLATE HASKELL
+
+    - hint: {lhs: "TH.varE 'a", rhs: "[|a|]", name: Use TH quotation brackets}
+
 - group:
     name: lens
     enabled: true
@@ -1217,6 +1222,8 @@
 # issue1183 = (a >= 'a') && a <= 'z' -- isAsciiLower a
 # issue1183 = (a >= 'a') && (a <= 'z') -- isAsciiLower a
 
+# import Language.Haskell.TH\
+# yes = varE 'foo -- [|foo|]
 # import Prelude \
 # yes = flip mapM -- Control.Monad.forM
 # import Control.Monad \
--- a/src/GHC/Util/Unify.hs
+++ b/src/GHC/Util/Unify.hs
@@ -224,6 +224,10 @@ unifyExp' nm root x@(L _ (HsApp _ x1 x2)) y@(L _ (HsApp _ y1 y2)) =
 unifyExp' nm root x y@(L _ (OpApp _ lhs2 op2@(L _ (HsVar _ op2')) rhs2)) =
   noExtra $ unifyExp nm root x y
 
+unifyExp' nm root (L _ (HsBracket _ (VarBr _ b0 (occNameStr -> v1))))
+                  (L _ (HsBracket _ (VarBr _ b1 (occNameStr -> v2))))
+    | b0 == b1 && isUnifyVar v1 = Just (Subst [(v1, strToVar v2)])
+
 unifyExp' nm root x y | isOther x, isOther y = unifyDef' nm x y
     where
         -- Types that are not already handled in unify.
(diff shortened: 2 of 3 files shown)


src/GHC/Util/Unify.hs
```diff
@@ -224,6 +224,10 @@ unifyExp' nm root x@(L _ (HsApp _ x1 x2)) y@(L _ (HsApp _ y1 y2)) =
 unifyExp' nm root x y@(L _ (OpApp _ lhs2 op2@(L _ (HsVar _ op2')) rhs2)) =
   noExtra $ unifyExp nm root x y
 
+unifyExp' nm root (L _ (HsBracket _ (VarBr _ b0 (occNameStr -> v1))))
+                  (L _ (HsBracket _ (VarBr _ b1 (occNameStr -> v2))))
+    | b0 == b1 && isUnifyVar v1 = Just (Subst [(v1, strToVar v2)])
```
ndmitchell, line 229: Why does the RHS have to be a HsBracket/VarBr thing? Why can't it be any arbitrary expression? Or anything inside an HsBracket?

ndmitchell (commented):
Looks good, although curious about the mentioned in the comments.

src/GHC/Util/Unify.hs
yairchu: Currently it means that things such as `''Maybe` match with `''a` on the LHS with `a` substituted into `Maybe`.

How should, say `1 + 2` match for `'a`? The rule `{lhs: "varE 'a", rhs: "[|a|]"}` will turn `varE (1 + 2)`, which is invalid code (unless `Name` had a `Num` instance), into the different and valid code `[|1 + 2|]`.

yairchu:
The tests originally worked for me because they were "without refactoring". I installed `refactor` and now get the same failure.

I tried adding `, note: RequiresExtension TemplateHaskell` in the hints or `{-# LANGUAGE TemplateHaskell #-}; ` in the test but neither solved it

src/GHC/Util/Unify.hs
ndmitchell: Fair enough - I don't really know enough details about this, but I'll trust you.

ndmitchell:
CC @zliu41 as the refactor master. @yairchu - can you produce a minimal refactor test case that triggers this? E.g. just capture the input, and we can make it a standalone test case.

yairchu:
> CC @zliu41 as the refactor master. @yairchu - can you produce a minimal refactor test case that triggers this? E.g. just capture the input, and we can make it a standalone test case.

Added a `refactor` test here: https://github.com/yairchu/apply-refact/compare/master...yairchu:th-test
Note that I did it manually and the test fails so am not 100% sure I made it right

zliu41:
This is caused by apply-refact using the wrong `DynFlags` for parsing the template. I'll fix it later today.

zliu41:
The refactoring error should go away using apply-refact 0.9.1.0 (https://github.com/ndmitchell/hlint/pull/1216).

@yairchu your new refactor test case should pass now, although your refact file is not quite right. The refact file can be obtained by `cabal run hlint -- /path/to/TemplateHaskell.hs --serialise`.

data/hlint.yaml
```diff
@@ -835,6 +835,14 @@
     - error: {lhs: "Control.Lens.nullOf (Control.Lens.at a)", rhs: "False"}
     - error: {lhs: "Control.Lens.nullOf (a . Control.Lens.at b)", rhs: "Control.Lens.nullOf a"}
 
+- group:
+    name: template-haskell
```
ndmitchell, line 839: Do we need a separate group? Why not just fold them into the main set?

ndmitchell (commented):
Cool, one last comment, and I'm happy to merge (happy to merge whether you want to keep the separate group or not)

data/hlint.yaml
yairchu: Good question :) Used a hint name instead!

data/hlint.yaml
```diff
@@ -1108,6 +1113,7 @@
 # no  = x `elem` y
 # no  = elem 1 [] : []
 # yes = a & (mapped . b) %~ c -- a <&> b %~ c
+# yes = Language.Haskell.TH.varE 'foo -- [|foo|]
```
ndmitchell, line 1116: Final nit - can we get a non-qualified example too, since otherwise we run the risk of missing real users code

yairchu: Sure thing, made the test use it non-qualified

ndmitchell:
Thanks!

PR #1216, opened 2021-02-24
Require apply-refact 0.9.1.0

zliu41:
0.9.1.0 includes a fix that uses the correct DynFlags to parse refactoring templates (https://github.com/mpickering/apply-refact/commit/7ec5a82096f1b59467c990a0cb105e7fffea056c), which should fix the refactoring error in #1215.

### DIFF
--- a/src/Refact.hs
+++ b/src/Refact.hs
@@ -74,4 +74,4 @@ runRefactoring rpath fin hints enabled disabled opts =  do
     waitForProcess phand
 
 minRefactorVersion :: Version
-minRefactorVersion = makeVersion [0,9,0,0]
+minRefactorVersion = makeVersion [0,9,1,0]


ndmitchell:
Thanks!

yairchu:
It seems that the tests now run without refactoring, see https://github.com/ndmitchell/hlint/runs/1973187770#step:7:252

Also locally, after I've installed latest refactor the tests run without it here to, even though refactor is of the correct version -

```
% refactor --version
v0.9.1.0%
```

zliu41:
@yairchu The line in the link says `Testing (with refactoring)` so it should have run the refactoring tests.
It also works for me locally. If `refactor` was installed via Cabal, and `~/.cabal/bin` is in `$PATH` then it should work. The path to `refactor` can also be specified via `--with-refactor`.

yairchu:
> @yairchu The line in the link says `Testing (with refactoring)` so it should have run the refactoring tests.
> It also works for me locally. If `refactor` was installed via Cabal, and `~/.cabal/bin` is in `$PATH` then it should work. The path to `refactor` can also be specified via `--with-refactor`.

Oops sorry you're right it was a mental parser error on my part..

PR #1217, opened 2021-02-25
Avoid making invalid "Use let" suggestions in ApplicativeDo blocks

zliu41:
In an ApplicativeDo block, the second last statement cannot be a `let`, for example, this doesn't compile:
```haskell
{-# LANGUAGE ApplicativeDo, DeriveFunctor #-}

newtype T a = T a deriving Functor

instance Applicative T where
  pure = T
  T a <*> T b = T (a b)

foo :: T Int
foo = do
  x <- pure 2
  let y = 3
  pure $ x + y
```

So we avoid making such suggestions on the second last statement, if the do-block is possibly an ApplicativeDo block.

### DIFF
--- a/src/Hint/Monad.hs
+++ b/src/Hint/Monad.hs
@@ -29,6 +29,7 @@ no = do x <- bar; return (f x x)
 {-# LANGUAGE RecursiveDo #-}; no = mdo hook <- mkTrigger pat (act >> rmHook hook) ; return hook
 yes = do x <- return y; foo x -- @Suggestion let x = y
 yes = do x <- return $ y + z; foo x -- let x = y + z
+yes = do x <- return a; y <- return b; return $ foo x y -- let x = a
 no = do x <- return x; foo x
 no = do x <- return y; x <- return y; foo x
 yes = do forM files $ \x -> return (); return () -- forM_ files $ \x -> return ()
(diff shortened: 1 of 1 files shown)


ndmitchell:
Should this be reported upstream to GHC? It seems a pretty arbitrary limitation...

src/Hint/Monad.hs
```diff
@@ -238,8 +239,21 @@ monadSteps _ _ = []
...
+    go f = \case
+      [] -> []
+      -- If the do block is possibly an ApplicativeDo block, don't make a suggestion on the
+      -- second last statement.
+      [_, x] | mayBeApplicativeDo x -> go f [x]
+      x : xs -> maybe id (:) (f x) (go f xs)
+
+    -- A do block is possibly an ApplicativeDo block if the last statement is one of
+    -- 'pure ...' , 'pure $ ...', 'return ...' or 'return $ ...'.
+    mayBeApplicativeDo = \case
```
googleson78, line 253: I think this will have too many false positives. I would wager *most* do blocks end like this :/

ndmitchell: Agreed - but at the same time - the hint isn't the greatest in the world, so I'm less concerned than I would be otherwise.

googleson78: I think I've seen it a few times in student code, so I do consider it to be useful as a teaching tool at least.

zliu41:
Yeah, agreed it should be reported upstream first to see if it's a GHC bug.

PR #1219, opened 2021-03-03
Suggest uncurry f (a, b) ==> f a b.

### LINKED ISSUE #1218: New suggestion request for: `f . uncurry (zipWith g) $ (a, b)`
This looks like a good one to have in general:

```haskell
module Z where
lhs f g a b = f . uncurry (zipWith g) $ (a, b)
rhs f g a b = f $ zipWith g a b
```
even GHC think they are the same on -O2:
```sh
$ ghc -O2 Z.hs -ddump-simpl
[1 of 1] Compiling Z                ( Z.hs, Z.o )

==================== Tidy Core ====================
<omitted>
rhs [InlPrag=NOUSERINLINE[2]]
  :: forall c t a b. ([c] -> t) -> (a -> b -> c) -> [a] -> [b] -> t
<omitted>
rhs = lhs
<omitted>
```

Javran:
Closes #1218.

(template content removed, no licensing concern from my side.)

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -310,6 +310,7 @@
     - warn: {lhs: "curry (\\(x,y) -> z)", rhs: "\\x y -> z"}
     - warn: {lhs: uncurry (curry f), rhs: f}
     - warn: {lhs: curry (uncurry f), rhs: f}
+    - warn: {lhs: "uncurry f (a, b)", rhs: f a b}
     - warn: {lhs: ($) (f x), rhs: f x, name: Redundant $}
     - warn: {lhs: (f $), rhs: f, name: Redundant $}
     - warn: {lhs: (Data.Function.& f), rhs: f, name: Redundant Data.Function.&}
@@ -1221,6 +1222,7 @@
 # issue1058 n = [] ++ issue1058 (n+1) -- issue1058 (n+1)
 # issue1183 = (a >= 'a') && a <= 'z' -- isAsciiLower a
 # issue1183 = (a >= 'a') && (a <= 'z') -- isAsciiLower a
+# issue1218 = uncurry (zipWith g) $ (a, b) -- zipWith g a b
 
 # import Language.Haskell.TH\
 # yes = varE 'foo -- [|foo|]


Javran:
Side note: while writing tests for this change, I noticed that if I write:
```
# issue1218 = f . uncurry (zipWith g) $ (a, b) -- f $ zipWith g a b
```
It fails the test with following message:

```
Found:
  f . uncurry (zipWith g) $ (a, b)
Perhaps:
  zipWith g (zipWith g a b)

WANTED: f $ zipWith g a b



Tests failed (1 of 790)
```

which looks wrong, not sure if I'm writing tests correctly.

ndmitchell:
Thanks for the diff - sorry it's taken a while to look at. Generally I recommend people don't update the changelog, and then I do it after the patch lands - otherwise conflicts are somewhat inevitable (as has happened here).

So the issue is:

```
TEST FAILURE (BAD REFACTORING)
SRC: /home/runner/.cabal/store/ghc-8.10.3/hlint-3.2.7-a58d0dd27ae13e2da84006e125f822349ffa8619eae6f646ddd8301ea2941c7a/share/hlint.yaml:1225:0
INPUT: issue1218 = uncurry (zipWith g) $ (a, b)
Refactor output is expected to contain: zipWith g a b
Actual: issue1218 = (zipWith g) a b
```

Namely that given `uncurry (zipWIth g) $ (a,b)` HLint generates the output `(zipWith g) a b`. But HLint knows that outer bracket is not useful, so strips it. The refactoring code doesn't know it's not useful, so leaves it present, and that gives us a warning.

@zliu41 as the refactor expert here. Should the test be more tolerant? Are we getting the output too early? Or is this just a test to mark `-- @NoRefactoring` and live with it?

zliu41:
When HLint strips the brackets, it should be able to make the corresponding change to `ideaRefactoring`. I can look into how to do that. Meanwhile, to unblock this PR we can just `-- @NoRefactoring` for now.

ndmitchell:
Try rebasing this on master, and hopefully the issue will have disappeared.

Javran:
Thanks for the update! I've just rebased onto master and reverted changes to the changelog.

ndmitchell:
Thanks!

PR #1220, opened 2021-03-11
Upgrade to ghc-lib-parser-ex-9.0.0.4

shayne-fletcher:
- Fixes issue with`ghc-lib-parser-ex-9.0.0.3` affecting ghc-9.0.1 non-`ghc-lib-parser` builds;
- When testing this I had to qualify the use of `singleton` in `Fixity.hs` - I don't know why but anyway fixed;
- Stack continues to be broken for ghc-9.0.1:
  - Tested with:
     ```bash
      cabal new-build --constraint "hlint -ghc-lib" --constraint "ghc-lib-parser-ex -auto +no-ghc-lib"
      ```

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -85,7 +85,7 @@ library
       build-depends:
           ghc-lib-parser == 9.0.*
     build-depends:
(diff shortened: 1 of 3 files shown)


ndmitchell:
Thanks!

PR #1221, opened 2021-03-13
Match import wildcards

### LINKED ISSUE #402: Allow module wildcards

tfausak:
Continues #1149 and fixes #402. 

I'm not necessarily happy with this code. I was trying to find the smallest change that fixed the issue. I think I did that, but the performance of this approach is probably bad. To check against a module name it does a map lookup as before, then it tries matching every pattern against the module name until it finds a match. This works but it could no doubt be faster. 

That being said, I haven't tested the real world performance of this. Perhaps it's acceptable? At any rate I'd be happy to work on a faster approach.

### DIFF
--- a/src/Apply.hs
+++ b/src/Apply.hs
@@ -20,7 +20,7 @@ import GHC.Hs
 import Language.Haskell.GhclibParserEx.GHC.Hs
 import qualified Data.HashSet as Set
 import Prelude
-import Util (wildcardMatch)
+import Util
 
 
 -- | Apply hints to a single file, you may have the contents of the file.
--- a/src/Apply.hs
+++ b/src/Apply.hs
@@ -20,7 +20,7 @@ import GHC.Hs
 import Language.Haskell.GhclibParserEx.GHC.Hs
 import qualified Data.HashSet as Set
 import Prelude
-import System.FilePattern (FilePattern, (?==))
+import Util (wildcardMatch)
 
 
 -- | Apply hints to a single file, you may have the contents of the file.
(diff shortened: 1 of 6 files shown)


ndmitchell:
Thanks, this looks pretty plausible.  People don't use restrictive matching that much, and if they start to, we can find/fix it then. Given we're quite a way since the GHC 9.0 release, and it's only just got to the point where our dependencies work with GHC 9.0, my inclination is to sit on this patch for a few days, make a GHC 9.0 release, and then do a proper review.

tfausak:
Sounds good to me! Absolutely no rush at all.

src/Util.hs
```diff
@@ -13,6 +13,8 @@ import System.IO.Unsafe
 import Unsafe.Coerce
 import Data.Data
 import Data.Generics.Uniplate.DataOnly
+import System.FilePattern (FilePattern, (?==))
+import Data.List.Extra (replace)
```
ndmitchell, line 17: Generally we just import open, especially for things like Data.List.Extra

ndmitchell (approved):
It's even got tests. What's not to love. I think this should probably be reflected with a sentence in the README somewhere - can you find an appropriate spot and add it?

src/Util.hs
tfausak, line 17:
```suggestion
import System.FilePattern
import Data.List.Extra
```

src/Hint/Restrict.hs
```diff
@@ -23,6 +23,7 @@ foo = nub s
 
 import Hint.Type(ModuHint,ModuleEx(..),Idea(..),Severity(..),warn,rawIdea)
 import Config.Type
+import Util (wildcardMatch)
```
tfausak, line 26:
```suggestion
import Util
```

src/Apply.hs
```diff
@@ -20,7 +20,7 @@ import GHC.Hs
 import Language.Haskell.GhclibParserEx.GHC.Hs
 import qualified Data.HashSet as Set
 import Prelude
-import System.FilePattern (FilePattern, (?==))
+import Util (wildcardMatch)
```
tfausak, line 23:
```suggestion
import Util
```

tfausak:
Thanks for the feedback! This should be ready for a re-review.

ndmitchell:
Thanks, looks great!

PR #1223, opened 2021-03-15
Add some traverse laws as default hints

googleson78:


data/hlint.yaml
```diff
@@ -247,6 +247,8 @@
 
     # TRAVERSABLES
 
+    - warn: {lhs: traverse pure x, rhs: pure x, name: "Traversable law"}
+    - warn: {lhs: traverse (pure . f) x, rhs: pure (map f x), name: "Traversable law"}
```
ndmitchell, line 251: Is this one true? What if the type isn't a list, but a different foldable structure?

googleson78: It definitely feels like it would be sensible for this to be right, but I'm not so sure anymore.

ndmitchell: Maybe this is true for `fmap` rather than `map`?

googleson78: Certainly it would be with `fmap`, yeah, but I'm not sure even that's true. I couldn't figure out a proof last time so I just ditched it for now..

data/hlint.yaml
```diff
@@ -247,6 +247,8 @@
 
     # TRAVERSABLES
 
+    - warn: {lhs: traverse pure x, rhs: pure x, name: "Traversable law"}
```
ndmitchell, line 250: You should drop the `x` on both sides, so `traverse pure` ==> `pure`.

data/hlint.yaml
ndmitchell: I'm confident enough in `traverse (pure . f) x` ==> `pure (fmap f x)` that I'm happy to include it.

ndmitchell (commented):
With the changes listed above I'm happy to merge this. It's good to be somewhat convinced the properties are true, but we don't need to be absolutely sure in a proof theoretic sense.

googleson78:
Btw, for a file like so:
```haskell
f = traverse (pure . g) xs
```
this produces two hints:
```
bar.hs:1:5-23: Warning: Use fmap
Found:
  traverse (pure . g)
Perhaps:
  (pure . fmap g)

bar.hs:1:5-26: Warning: Use fmap
Found:
  traverse (pure . g) xs
Perhaps:
  pure (fmap g xs)
```
Is this fine?

data/hlint.yaml
googleson78: Fixed

data/hlint.yaml
googleson78: Changed it

googleson78:
I updated the hints file like so:
```sh
hlint --generate-summary -h data/hlint.yaml
```
But there seem to be way too many changes. Am I doing something wrong?

ndmitchell:
Don't know what's up with this - can you spin the diff without the hints.md updates, and I'll just merge it with a broken CI, and fix up after?

ndmitchell:
The README says how to generate the hints - see https://github.com/ndmitchell/hlint/blob/master/README.md#how-to-run-tests, which says `hlint --generate-summary`. I just tried that and it worked for me.

googleson78:
There still seems to be some unrelated change in there when I do that..

hints.md
```diff
@@ -11574,7 +11622,7 @@ a ?~ b
 <td>
 LHS:
 <code>
-a & (mapped %~ b)
+(mapped %~ b) a
```
googleson78, line 11625: Don't know why this is happening

ndmitchell: Odd. Which GHC version are you on? I'm using 9.0.1

ndmitchell:
I've merged and fixed the hints.md locally

googleson78:
Thanks, and sorry.

I was building it using stack, for which the `stack.yaml` seems to be on 8.10.3.. Maybe if I rebased on master it would've been fine..

PR #1227, opened 2021-03-19
Support removing brackets from templates/substs

zliu41:
This fixes the refactoring test in https://github.com/ndmitchell/hlint/pull/1219.


ndmitchell:
Nice work!

Issue #1230, opened 2021-03-28
HLint fails to parse file GHC-9.0 acceps:


phadej:
```haskell
module Parse where

readRational :: String -> Rational -- NB: *does* handle a leading "-"
readRational top_s
  = case top_s of
      '-' : xs -> - (read_me xs)
      xs       -> read_me xs
  where
    read_me s
      = case (do { (x,"") <- readRational__ s ; return x }) of
          [x] -> x
          []  -> error ("readRational: no parse:"        ++ top_s)
          _   -> error ("readRational: ambiguous parse:" ++ top_s)

readRational__ = undefined
```

GHC-9.0.1 compiles this fine, HLint (3.3 compiled with GHC-8.6) says:

```
% hlint Parse.hs 
Parse.hs:6:19: Error: Parse error: on input `-'
Found:
    readRational top_s
      = case top_s of
  >       '-' : xs -> - (read_me xs)
          xs       -> read_me xs
      where
  

1 hint
```

shayne-fletcher:
The issue is that HLint is default enabling `LexicalNegation`. If you compile the above program in the presence of `{-# LANGUAGE LexicalNegation #-}` with ghc-9.0  you get the error that HLint is trying to report:
```
Parse.hs:8:19: error: parse error on input ‘-’
  |
8 |       '-' : xs -> - (read_me xs)
  |                   ^
```

One possible fix appears to be writing `'-' : xs -> -(read_me)` (remove the space following the`'-'`).

phadej:
For the record, this is second issue in short time when HLint by default enabled flags without unclear reason "why this happens" (#1222 is another one).

I'd very much like "Don't enable any extensions" configuration option.

ndmitchell:
It sounds like LexicalNegation is an extension that should be disabled by default? Not enabled in normal GHC, changes the way parsing works, but doesn't allow additional programs - concur @shayne-fletcher ?

shayne-fletcher:
If judged by that criteria then yes.

ndmitchell:
I've disabled LexicalNegation - looks like there are other diffs up for the other parts of this, so closing.

PR #1231, opened 2021-03-29
Add flag: no-default-extensions

shayne-fletcher:
Issue https://github.com/ndmitchell/hlint/issues/1230 requests a runtime flag to prevent the default enabling of language extensions. This PR implements such a flag: `no-default-extensions`.


ndmitchell:
Rather than an explicit flag, why not make `-XHaskell98` and `-XHaskell2010` also turn off all extensions they disable, and then we don't need any few flags, and you can be explicit about what "default" means. So these flags would define a end state, rather than a set of things to turn on/off from the base state.

shayne-fletcher:
Sounds good. Raised issue https://github.com/ndmitchell/hlint/issues/1233. Closing this PR.

Issue #1233, opened 2021-04-04
Make -XHaskell98 and -XHaskell2010 also turn off all extensions they disable


shayne-fletcher:
See https://github.com/ndmitchell/hlint/pull/1231 for some discussion of why this would be a good idea. The code to enhance would appear to be the `getExtensions` function in `CmdLine.hs`.

ndmitchell:
Done thanks to #1234

PR #1234, opened 2021-04-04
Seed default extensions from language if provided

shayne-fletcher:
Intended to fix issue https://github.com/ndmitchell/hlint/issues/1233 which came about from discussion on PR https://github.com/ndmitchell/hlint/pull/1231.

If a language is provided e.g. `Haskell98` or `Haskell2010` (and soon `GHC2021`) then use the set of extensions it represents as the default extension list.

Did an ad-hoc test with this program:
```haskell
module Parse where

readRational :: String -> Rational -- NB: *does* handle a leading "-"
readRational top_s
  = case top_s of
      '-' : xs -> - (read_me xs)
      xs       -> read_me xs
  where
    read_me s
      = case (do { (x,"") <- readRational__ s ; return x }) of
          [x] -> x
          []  -> error ("readRational: no parse:"        ++ top_s)
          _   -> error ("readRational: ambiguous parse:" ++ top_s)

readRational__ = undefined
```

`stack run -- ~/Foo.hs`  gives a parse error due to the default of `LexicalNegation` whereas `stack run -- -XHaskell2010 ~/Foo.hs` gives no hints as you'd hope.

### DIFF
--- a/src/CmdLine.hs
+++ b/src/CmdLine.hs
@@ -288,23 +288,41 @@ getModule _ _ _ = pure Nothing
 
 
 getExtensions :: [String] -> (Maybe Language, ([Extension], [Extension]))
-getExtensions args =
-  (lang, foldl f (if null langs then (defaultExtensions, []) else ([], [])) exts)
-    where
-        lang = if null langs then Nothing else Just $ fromJust $ lookup (last langs) ls
-        (langs, exts) = partition (isJust . flip lookup ls) args
-        ls = [(show x, x) | x <- [Haskell98, Haskell2010]]
+getExtensions args = (lang, foldl f (startExts, []) exts)
+  where
+        -- If a language specifier is provided e.g. Haskell98 or
+        -- Haskell2010 (and soon GHC2021), then it represents a
+        -- specific set of extensions which we default enable.
 
-        f (a, e) "Haskell98" = ([], [])
-        f (a, e) ('N':'o':x) | Just x <- GhclibParserEx.readExtension x, let xs = expandDisable x =
-            (deletes xs a, xs ++ deletes xs e)
+        -- If no language specifier is provided we construct our own
+        -- set of extensions to default enable. The set that we
+        -- construct default enables more extensions than GHC would
+        -- default enable were it to be invoked without an explicit
+        -- language specifier given.
+        startExts :: [Extension]
+        startExts = case lang of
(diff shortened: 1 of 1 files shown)


src/CmdLine.hs
```diff
@@ -289,20 +289,34 @@ getModule _ _ _ = pure Nothing
...
+        lang =
+          if null langs
+            then Nothing
+            else Just $ fromJust $ lookup (last langs) ls
+
         (langs, exts) = partition (isJust . flip lookup ls) args
-        ls = [(show x, x) | x <- [Haskell98, Haskell2010]]
+        ls = [ (show x, x) | x <- [Haskell98, Haskell2010 {-, GHC2021-}] ]
 
         f (a, e) "Haskell98" = ([], [])
```
shayne-fletcher, line 313: This pre-existed. Should it be removed? Can this case even trigger?

src/CmdLine.hs
```diff
@@ -18,6 +18,7 @@ import GHC.All(CppFlags(..))
 import GHC.LanguageExtensions.Type
 import Language.Haskell.GhclibParserEx.GHC.Driver.Session as GhclibParserEx
 import GHC.Driver.Session hiding (verbosity)
+import GHC.Utils.Misc(lastMaybe)
```
ndmitchell, line 21: Use `unsnoc` from `Data.List.Extra`.

src/CmdLine.hs
```diff
@@ -289,22 +290,39 @@ getModule _ _ _ = pure Nothing
...
-        lang = if null langs then Nothing else Just $ fromJust $ lookup (last langs) ls
+  -- If a language specifier is provided e.g. Haskell98 or Haskell2010
+  -- (and soon GHC2021), then it represents a specific set of
+  -- extensions which we default enable.
+
+  -- If no language specifier is provided we construct our own set of
+  -- extensions to default enable. The set that we construct default
+  -- enables more extensions than GHC would default enable were it to
+  -- be invoked without an explicit language specifier given.
+  let startExts =
```
ndmitchell, line 301: Why not shove `startExts` under the `where` block with the other definitions?

ndmitchell (approved):
Looks great, couple of minor cleanups, but yes, we should do this!

ndmitchell:
Thanks!

Issue #1237, opened 2021-04-05
Add CI to generate the hints.md document


ndmitchell:
The document https://github.com/ndmitchell/hlint/blob/master/hints.md often ends up out of date. The only way to guarantee it is always fresh is to tie it into the CI. Usual approach would be generate it, make regeneration easier, and if it changes, fail CI.

autophagy:
@ndmitchell Hi Neil! I would like to try and contribute to hlint by solving this issue. I was wondering, since the CI seems to use the `ndmitchell/neil@master` action for a large amount of the CI work, whether you think it'd be better to add the `hints.md` freshness check in the CI step here or in the `ndmitchell/neil@master` action?

ndmitchell:
Great, thanks!

You shouldn't edit `ndmitchell/neil`, since that's the actions that are common to everything. The solution is to edit https://github.com/ndmitchell/hlint/blob/master/travis.hs - that's a script that is run by my CI (named from back in the day when it actually ran on Travis). If it produces a non-zero exit code it fails. Alternatively, if it leaves the repo not in a clean state, it fails. Having that script call out to regeneration is probably sufficient.

I actually don't know how to generate the hints - its something that should be written down at https://github.com/ndmitchell/hlint/blob/master/README.md#hacking-hlint, but isn't. Once you figure it out (I think its some command line flags to HLint) can you please write it there too.

autophagy:
@ndmitchell Cool! Thank you for the pointers :slightly_smiling_face:

PR #1238, opened 2021-04-12
Add CI step to generate hints.md

autophagy:
As part of the CI steps, hlint new executes `--generate-summary`, which
means the CI will fail if any changes to `hints.md` are written. This
aims to keep the hints summary list up to date.

This commit also refreshes the `hints.md` document.

This should address #1237


ndmitchell:
Thanks, this looks great!

PR #1246, opened 2021-05-10
Allow configuration to forbid all imports from a module, with exceptions

### LINKED ISSUE #1244: Forbid all functions from module with exceptions
I'd like to forbid all functions from `Control.Exception` (requiring using `UnliftIO.Exception`), but I'd like to allow certain imports like `Exception`, where someone might reasonably want to derive `Exception` without wanting to pull in all the `unliftio` dependencies.

I think you can forbid things like
```yaml
- modules:
    - name: Control.Exception
      badidents: [catch, catchJust, handle, handleJust, ...]
```
but it'd be nice to do it the other way
```yaml
- modules:
    - name: Control.Exception
      only: [Exception, SomeException]
```

brandon-leapyear:
Resolves #1244 

This PR includes the following changes:
* Previously, `badidents` was not disallowed when used on a `function:` block, but from what I can tell, it only makes sense in a `module:` block. Updated the `allowFields` call to only allow `badidents` for a `module:` block
* Change README paragraph to a bulleted list, which makes it much easier to correspond the hint in the config to the explanation for the hint
* Some refactoring in the `checkImports` function. Not strictly necessary, but IMO _much_ easier to read and understand. If reviewers disagree, happy to revert.
* Add a `RestrictIdents` data type and the `only` option to solve the ticket. This makes it a configuration error to provide both `badidents` and `only`. I'm not sure how they should interact if these options are not mutually exclusive.
* Add tests for both `badidents` and `only`

### DIFF
--- a/src/Config/Type.hs
+++ b/src/Config/Type.hs
@@ -1,7 +1,7 @@
 
 module Config.Type(
     Severity(..), Classify(..), HintRule(..), Note(..), Setting(..),
-    Restrict(..), RestrictType(..), SmellType(..),
+    Restrict(..), RestrictType(..), RestrictIdents(..), SmellType(..),
     defaultHintName, isUnifyVar, showNotes, getSeverity, getRestrictType, getSmellType
     ) where
 
@@ -106,13 +106,23 @@ data HintRule = HintRule
 
 data RestrictType = RestrictModule | RestrictExtension | RestrictFlag | RestrictFunction deriving (Show,Eq,Ord)
 
+data RestrictIdents
+    = NoRestrictIdents -- No restrictions on module imports
+    | ForbidIdents [String] -- Forbid importing the given identifiers from this module
+    deriving Show
+
+instance Semigroup RestrictIdents where
+    NoRestrictIdents <> ri = ri
+    ri <> NoRestrictIdents = ri
+    ForbidIdents x1 <> ForbidIdents y1 = ForbidIdents $ x1 <> y1
+
 data Restrict = Restrict
     {restrictType :: RestrictType
     ,restrictDefault :: Bool
     ,restrictName :: [String]
     ,restrictAs :: [String] -- for RestrictModule only, what module names you can import it as
     ,restrictWithin :: [(String, String)]
-    ,restrictBadIdents :: [String]
+    ,restrictIdents :: RestrictIdents -- for RestrictModule only, what identifiers can be imported from it
     ,restrictMessage :: Maybe String
     } deriving Show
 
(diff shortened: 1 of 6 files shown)


tests/hint.test
```diff
@@ -99,6 +99,82 @@ Note: Custom message
...
+RUN tests/restricted-onlyidents-valid.lhs --hint=data/test-restrict.yaml
+FILE tests/restricted-onlyidents-valid.lhs
+> import Restricted.Module.OnlyIdents (good)
+OUTPUT
+No hints
+
+---------------------------------------------------------------------
+RUN tests/restricted-onlyidents-universal.lhs --hint=data/test-restrict.yaml
+FILE tests/restricted-onlyidents-universal.lhs
+> import Restricted.Module.OnlyIdents
```
brandon-leapyear, line 174: This is currently not an error, and I added this test-case to make this explicit. If this should actually be an error, we should fix this test. (Same with `badidents-universal`)

brandon-leapyear:
@ndmitchell any update on this?

src/Config/Type.hs
```diff
@@ -106,13 +106,26 @@ data HintRule = HintRule
...
+    | ForbidIdents [String] -- Forbid importing the given identifiers from this module
+    | OnlyIdents [String] -- Forbid importing all identifiers from this module, except the given identifiers
+    deriving Show
+
+instance Semigroup RestrictIdents where
+    NoRestrictIdents <> ri = ri
+    ri <> NoRestrictIdents = ri
+    ForbidIdents x1 <> ForbidIdents y1 = ForbidIdents $ x1 <> y1
+    OnlyIdents x1 <> OnlyIdents x2 = OnlyIdents $ x1 <> x2
+    ri1 <> ri2 = error $ "Incompatible restrictions: " ++ show (ri1, ri2)
```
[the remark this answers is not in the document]
ndmitchell: Yeah, it's a corner case, of people who are writing restrictions, so won't hit most people. I'm OK with that, and if it turns out to be a problem, we can improve later

tests/hint.test
ndmitchell: I would have guessed that if you say `foo` is bad, then if you mention `foo`, you are doing it wrong. Whereas with `only` I think it's more reasonable to raise an error. But I could be persuaded either way.

ndmitchell (approved):

ndmitchell:
Thanks for the patch. Sorry it took so disastrously long to review it! If we ever meet in person I definitely owe you a beer to apologise. All merged now, and I'll be making a release in the next hour.

brandon-leapyear:
No worries at all! Thanks for reviewing!

PR #1252, opened 2021-06-02
two hints in codeworld-api

jvoigtlaender:
These are patterns I noticed in student code when working with exercises in CodeWorld.


ndmitchell:
Thanks!

PR #1253, opened 2021-06-02
two general hints

jvoigtlaender:
Observed both in student code, and both seem useful in general.

I'm not sure what names they get automatically.

Probably "Use tail" for
```
warn: {lhs: reverse (init x), rhs: tail (reverse x)}
```
and "Use all" for
```
warn: {lhs: null (concat x), rhs: all null x}
```
which would both seem fine names to me.


ndmitchell:
Thanks!

PR #1254, opened 2021-06-02
hint about 'null' on singleton list

jvoigtlaender:
Not sure anyone outside a teaching context would make this "mistake". That's why I put it into the `teaching` group. But in principle the hint could also go into the default group instead.


ndmitchell:
Thanks for the diff. I think this would be good to put in the normal group. Lots of things like this happen when people refactor - while only beginners might write it directly, pretty much everything crops up after long enough.

jvoigtlaender:
Makes sense. I have moved the hint to the `default` group by ammending the commit. So it should be good to go in this form now.

PR #1255, opened 2021-06-02
suggest use of filter instead of simple list comprehension

jvoigtlaender:
Analogous to the "Use map" hint right above it in hint list.


jvoigtlaender:
This is in the `teaching` group.

ndmitchell:
Thanks, looks good

PR #1258, opened 2021-06-16
two hints about putStr -> putStrLn

jvoigtlaender:
Typical occurrence (in student code) where the first hint applies is something like
```haskell
   putStr (show n ++ "\n")
```
which this will then suggest to replace by
```haskell
   putStrLn (show n)
```
after which an already existing hlint warning will say to replace it by
```haskell
   print n
```

Typical occurrence where the second hint from this commit applies is something like
```haskell
   putStr ("The result is: " ++ show n ++ "\n")
```
which this should then suggest to replace by
```haskell
   putStrLn ("The result is: " ++ show n)
```

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -66,6 +66,8 @@
     # I/O
 
     - warn: {lhs: putStrLn (show x), rhs: print x}
+    - warn: {lhs: putStr (x ++ "\n"), rhs: putStrLn x}
+    - warn: {lhs: putStr (x ++ y ++ "\n"), rhs: putStrLn (x ++ y)}
     - warn: {lhs: mapM_ putChar, rhs: putStr}
     - warn: {lhs: hGetChar stdin, rhs: getChar}
     - warn: {lhs: hGetLine stdin, rhs: getLine}


ndmitchell:
Do we need two hints because of the associativity of the `++` operator? What if the user writes `putStr (a ++ b ++ c ++ "\n")` - I guess that then doesn't match? (I think I'm OK with that, just want to confirm the reasoning)

jvoigtlaender:
First, about whether associativity allows saving one of the two hints now proposed: You‘d have to know. 😄  Does Hlint do any magic with associativity while matching against hints? If not, I‘m afraid the default associativity of ++ is the wrong way around to save a hint here.

Second, about the situation with three occurrences of ++: Yes, that wouldn‘t be covered, or would require yet another variant of the hint. It could be added. But actually, my impression is that it occurs much less frequently than the two scenarios I indicated above. At least in the student code I‘m seeing. There, something like that (more than two things, and then a newline) mostly appears with a concrete string for your c variable (e.g., the string “.“ for end-of-sentence punctuation). And then it wouldn’t be written as c ++ “\n“ anyway, but directly as “.\n“, and none of the hints would match.

jvoigtlaender:
Ah, you probably didn’t mean to propose that associativity might save one hint. Instead, were seeking confirmation that the way in which ++ is associative made me propose two hints. Yes, that‘s the case.

ndmitchell:
Thanks, that makes sense. We have special support for `.` and `$` to deal with associativity, but nothing in general.

PR #1262, opened 2021-06-28
note that you can ignore function for certain hint

unhammer:


unhammer:
(I read the readme and thought "oh that's too bad that you can't combine them", and then I tried the obvious thing just to check and it worked)

ndmitchell:
Thanks! Good point to include

Issue #1261, opened 2021-06-27
Bad suggestion in the presence of MonadComprehensions


hseg:
```[x | True] = return x``` in the presence of ```MonadComprehensions```,
but hlint incorrectly suggests replacing it with ```[x]``` instead.
Using a trivial guard here for homogeneity in a long list of clauses like
```
fromMaybe [] . sequence $
   [ [f x y | True]
   , [g x r | r <- k]
   , [h x | True]
   , [j r s | r <- l, s <- m]
   ]
```
cf #775

**Comment by ndmitchell** [comment]
Looking at that code it seems like GHC knows whether it is a monad comp or not, so just a case of checking for that case. We should also check that monad comprehensions are not enabled by default. Patch welcome!

**Comment by shayne-fletcher** [comment]
Monad comprehensions are enabled by default.

```haskell
f :: Maybe Int
f = [ x + y | x <- Just 1, y <- Just 2 ]

main = print f
```

A quick play with this program suggests parsing succeeds regardless of whether the extension is in effect or not. The extension is required to be enabled for the program to type however.

It expect therefore disabling the extension by default won't have an effect, good or bad.

ndmitchell:
I think the GHC syntax tree is different if monad comprehensions are enabled, so we should probably disable monad comps, then if they are enabled and reflected in the syntax tree it is probably because a user enabled them explicitly, and then we should not apply the `[x | True]` modification.

shayne-fletcher:
The parser `list` rule (for comprehensions)
```
        | texp '|' flattenedpquals
```
inspects the parser's `MonadComprehensionBit` in order to calculate correctly the "statement context" (list or monad) via
```haskell
checkMonadComp = do
    monadComprehensions <- getBit MonadComprehensionsBit
    return $ if monadComprehensions
                then MonadComp
                else ListComp
```
So yes, you are right, the parse tree is different when the rule is enabled from when it's not.

I will offer a PR that stops `MonadComprehensions` from being enabled by default.

hseg:
Was going through my old open tickets, this is still broken on `Hlint 3.8`:
```haskell
{-# LANGUAGE MonadComprehensions #-}

f x = [x | True]

g = fromMaybe [] . sequence $
   [ [f x y | True]
   , [g x r | r <- k]
   , [h x | True]
   , [j r s | r <- l, s <- m]
   ]
```
emits (regardless of `MonadComprehensions`):
```
[...]

PR #1263, opened 2021-07-03
Do not default enable MonadComprehensions

shayne-fletcher:
In issue https://github.com/ndmitchell/hlint/issues/1261 @ndmitchell  advises `MonadComprehensions` should not be default enabled.

### DIFF
--- a/src/Extension.hs
+++ b/src/Extension.hs
@@ -27,6 +27,7 @@ reallyBadExtensions =
   , NegativeLiterals -- Was not enabled by HSE and enabling breaks tests.
   , StarIsType -- conflicts with TypeOperators. StarIsType is currently enabled by default,
                -- so adding it here has no effect, but it may not be the case in future GHC releases.
+  , MonadComprehensions -- Discussed in https://github.com/ndmitchell/hlint/issues/1261
   ]
 
 -- | Extensions we turn on by default when parsing. Aim to parse as


shayne-fletcher:
Unfortunately, turning off `MonadComprehensions` breaks this rule:
```
    - hint: {lhs: "[x | x <- y]", rhs: "y", side: isVar x, name: Redundant list comprehension}
```
I had a bit of a scratch around but it's not immediately obvious how/why. @ndmitchell do you mind having a look?

shayne-fletcher:
> Unfortunately, turning off `MonadComprehensions` breaks this rule:
> 
> ```
>     - hint: {lhs: "[x | x <- y]", rhs: "y", side: isVar x, name: Redundant list comprehension}
> ```
> 
> I had a bit of a scratch around but it's not immediately obvious how/why. @ndmitchell do you mind having a look?

Never mind, it's all good. I worked it out.

ndmitchell:
Regarding the rule it appeared to break. Was that a false alarm and it doesn't break it?

shayne-fletcher:
The "false alarm" was that `MonadComprehensions` were removed from `defaultExtensions` while still present in `configExtensions` breaking unification. 

Since `badExtensions` "extends" `reallyBadExtensions`, `MonadComprehensions` belongs in the second set.
```haskell

defaultExtensions :: [Extension]
defaultExtensions = enumerate \\ badExtensions

-- | Extensions we turn on when reading config files, don't have to deal with the whole world
--   of variations - in particular, we might require spaces in some places.
configExtensions :: [Extension]
configExtensions = enumerate \\ reallyBadExtensions
```

ndmitchell:
Ah, very subtle! So monad and non-monad comps won't unify. Not obvious, but also probably not too important.

Issue #1265, opened 2021-07-14
Proposing to add: '() <$ fa' -> 'void fa'


Anton-Latukha:
Encountered in the [wild](https://github.com/haskell-nix/hnix/blob/84dcb2e3781da5e9b50ac528e966438b051d4540/tests/NixLanguageTests.hs#L110), a line managed to survive several code rewrites & refactors by several people.

HLint has rules "around" this code, but not about it:
```yaml
    - hint: {lhs: const x <$> y, rhs: x <$ y}
    - warn: {lhs: const () <$> x, rhs: Control.Monad.void x}
```
in the current set there is no commuting of rules that for `() <$ x` would point to `void`.

Proposing a rule:
```yaml
    - warn: {lhs: () <$ x, rhs: Control.Monad.void x}
```

Since `() <$ x` is a [void](https://hackage.haskell.org/package/base-4.15.0.0/docs/src/Data-Functor.html#void) implementation:
```haskell
void :: Functor f => f a -> f ()
void x = () <$ x
```

Anton-Latukha:
<details><summary>Minor detail</summary>

</p>

After addition of the `void` definition, the rule:
```yaml
    - warn: {lhs: const () <$> x, rhs: Control.Monad.void x}
```
can be reduced, since it is a sequential application of two rules:
```yaml
    - hint: {lhs: const x <$> y, rhs: x <$ y}
    - warn: {lhs: () <$ x, rhs: Control.Monad.void x}
```

---

</details>

PR #1266, opened 2021-07-14
Suggest: '() <$ x' -> 'void x'

Anton-Latukha:
Closes: https://github.com/ndmitchell/hlint/issues/1265


Anton-Latukha:
Due to `Control.Monad.void` being part of the [Haskell 2010](https://www.haskell.org/onlinereport/haskell2010/haskellch13.html#x21-19300013), I left `void` in Monad rules & as `Control.Monad.void`, not as `Control.Functor.void`.

ndmitchell:
Thanks!

PR #1267, opened 2021-07-18
Change lambdaDecl to lambdaBind

### LINKED ISSUE #1264: Eta reduction under `where`
Eta reduction is not suggested under `where` clause.

It *is* suggested in the following line:
```hs
someFunc x y z = someOtherFunc x y z
```

But not when you put the same line under `where` clause:
```hs
f a b = someOperations $ someFunc a b c
  where someFunc x y z = someOtherFunc x y z
```

zliu41:
Fixes #1264


ndmitchell:
Thanks!

PR #1268, opened 2021-07-29
moving 'nub' for efficiency

jvoigtlaender:
I saw
```
catMaybes (nub x)
```
in student code.

Due to the non-linear runtime complexity of `nub`, the linear runtime complexity of `catMaybes`, and the fact that `catMaybes` never makes the list longer, performing these two operations the other way around should always be better.

The cases with `lefts` and `rights` are analogous.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -615,6 +615,18 @@
     - warn: {lhs: isNothing (fmap f x), rhs: isNothing x}
     - warn: {lhs: fromJust (fmap f x), rhs: f (fromJust x), note: IncreasesLaziness}
     - warn: {lhs: mapMaybe f (fmap g x), rhs: mapMaybe (f . g) x, name: Redundant fmap}
+    - warn: {lhs: catMaybes (nub x), rhs: nub (catMaybes x)}
+    - warn: {lhs: lefts (nub x), rhs: nub (lefts x)}
+    - warn: {lhs: rights (nub x), rhs: nub (rights x)}
+    - warn: {lhs: catMaybes (reverse x), rhs: reverse (catMaybes x)}
+    - warn: {lhs: lefts (reverse x), rhs: reverse (lefts x)}
+    - warn: {lhs: rights (reverse x), rhs: reverse (rights x)}
+    - warn: {lhs: catMaybes (sort x), rhs: sort (catMaybes x)}
+    - warn: {lhs: lefts (sort x), rhs: sort (lefts x)}
+    - warn: {lhs: rights (sort x), rhs: sort (rights x)}
+    - warn: {lhs: catMaybes (nubOrd x), rhs: nubOrd (catMaybes x)}
+    - warn: {lhs: lefts (nubOrd x), rhs: nubOrd (lefts x)}
+    - warn: {lhs: rights (nubOrd x), rhs: nubOrd (rights x)}
 
     # EITHER
 


jvoigtlaender:
There are a few further, similar potential rules that could be added:

With `reverse`:
```
    - warn: {lhs: catMaybes (reverse x), rhs: reverse (catMaybes x)}
    - warn: {lhs: lefts (reverse x), rhs: reverse (lefts x)}
    - warn: {lhs: rights (reverse x), rhs: reverse (rights x)}
```
(not necessarily big performance wins here, but at least the `reverse` might get to work on a shorter list).

With `cycle`:
```
    - warn: {lhs: catMaybes (cycle x), rhs: cycle (catMaybes x)}
    - warn: {lhs: lefts (cycle x), rhs: cycle (lefts x)}
    - warn: {lhs: rights (cycle x), rhs: cycle (rights x)}
```
(where an additional benefit is that `catMaybes` etc. need to process fewer elements after the rule applied).

With `sort`:
```
    - warn: {lhs: catMaybes (sort x), rhs: sort (catMaybes x)}
    - warn: {lhs: lefts (sort x), rhs: sort (lefts x)}
    - warn: {lhs: rights (sort x), rhs: sort (rights x)}
```
(seems correct given the `Ord` instances for `Maybe` and `Either`, and could be very beneficial again due to the non-linear runtime complexity of sorting).

jvoigtlaender:
I was also thinking about whether there could be similar rules for functions like `delete :: Eq a => a -> [a] -> [a]`, but it seems less clear what to do there.

jvoigtlaender:
Another possible direction would be to also consider functions from the `containers` and/or `extra` packages, like `nubOrd :: Ord a => [a] -> [a]` and `nubSort :: Ord a => [a] -> [a]`

ndmitchell:
Moving things like reverse/sort to work on fewer elements makes a lot of sense, as does nub. Happy to merge this PR as is, or let you extend it for the sort/reverse cases too - let me know.

In most cases I think avoiding nub is the right thing to do, using nubOrd or similar. But it would be easier if there was a non-quadratic nub-like in the base library, since I'm a little wary about recommending another dependency. I think that's definitely a wart in the standard library (`nub` should probably be deleted entirely, it's such a massive foot-gun).

jvoigtlaender:
I've added the corresponding rules for `reverse` and `sort`.

I've not added the rules for `cycle` since you did not mention/comment on them.

I see what you mean about "`nub` shouldn't even exist". Also why you are not currently having Hlint propose to replace `nub` by `nubOrd` or `nubSort` (them not being in `base`). Actually, my thinking about them here wasn't even to propose such a replacement. Only to also add rules like
```yaml
    - warn: {lhs: catMaybes (nubOrd x), rhs: nubOrd (catMaybes x)}
    - warn: {lhs: lefts (nubOrd x), rhs: nubOrd (lefts x)}
    - warn: {lhs: rights (nubOrd x), rhs: nubOrd (rights x)}
```
to the section https://github.com/ndmitchell/hlint/blob/11898f6371ea6175f11e1a7049e85d2d8c1f7e96/data/hlint.yaml#L919-L923 That is, when people are already using the extra dependency.

ndmitchell:
You can add the rules about `nubOrd` to the normal section - the idea is that the `extra` rules are opt-in since you are wanting to say you are explicitly OK with adding a dependency of extra. However, if you've already used `nubOrd` in the LHS, you must be OK with that, so they can go in the normal enabled as standard section.

jvoigtlaender:
That reasoning makes sense!

However, I am now a bit confused/puzzled, probably since I don't know how exactly hlint resolves the function names it sees:
1. If I add a rule like `- warn: {lhs: catMaybes (nubOrd x), rhs: nubOrd (catMaybes x)}` to the `default` group, how will hlint know that by `nubOrd` I mean the function from the `extra` package? (I guess one might add that as a package import to the group, but the 2. item below has me puzzled on how that works.)
2. In the existing `extra` group there is no import of the `extra` package. So what is going on there? How does hlint resolve `nubOrd` in a rule like `- warn: {lhs: "nubOrdBy compare", rhs: "nubOrd"}` to the function of that name in `Data.List.Extra`?

All this has me wondering particularly since `nubOrd` exists in many places, and not all of them are the exact same function presumably (https://hoogle.haskell.org/?hoogle=nubOrd&scope=set%3Astackage). So how does this work out at all?

ndmitchell:
HLint has a bunch of guesses about which name a particular identifier corresponds to. If it knows nubOrd came from Data.List.Extra it will see if there are any imports of that and make a best guess. If it doesn't (which is the case if you don't give it an explicit import) it makes the guess that it probably is the nubOrd you are thinking about. It's not a particularly satisfactory aspect of HLint, and there are a bunch of bugs associated with that relaxed approach that we've accumulated under a tag: https://github.com/ndmitchell/hlint/labels/scope

jvoigtlaender:
Thank you for the explanation. That really makes some aspects of hlint's rule application clearer!

I've now added the three rules for `nubOrd` as well, in the default group.

PR #1272, opened 2021-08-05
Suggest `fold` for `foldr (<>) mempty`

### LINKED ISSUE #1271: use mconcat is wrong
Hlint hints that `foldr (<>) mempty` can be replaced with `mconcat`. This isn't true.

```
foldr (<>) mempty :: (Monoid a, Foldable f) => f a -> a
mconcat :: Monoid a => [a] -> a
```

zliu41:
Fixes #1271


ndmitchell:
Thanks!

PR #1273, opened 2021-08-05
Fix OverloadedLists suggestion as documented in #114

### LINKED ISSUE #114: Incorrect suggestions with OverloadedLists
``` Haskell
{-# LANGUAGE OverloadedLists #-}
module Baz where

import Data.Foldable

foo :: Foldable t => t a -> t a
foo = id

bar = foo (True:[])
```

HLint suggests: 

```
Warning: Use list literal
Found:
  True : []
Why not:
  [True]
```

This will prevent the program from typechecking due to the ambiguity introduced by OverloadedLists

kk-hainq:
Fixes #114.

@ndmitchell I wonder if I should add a test for this case?

### DIFF
--- a/src/Hint/List.hs
+++ b/src/Hint/List.hs
@@ -45,7 +45,7 @@ import Data.List.Extra
 import Data.Maybe
 import Prelude
 
-import Hint.Type(DeclHint,Idea,suggest,ignore,substVars,toRefactSrcSpan,toSS)
+import Hint.Type(DeclHint,Idea,suggest,ignore,substVars,toRefactSrcSpan,toSS,ghcAnnotations)
 
 import Refact.Types hiding (SrcSpan)
 import qualified Refact.Types as R
@@ -67,11 +67,13 @@ import Language.Haskell.GhclibParserEx.GHC.Types.Name.Reader
 
 
 listHint :: DeclHint
-listHint _ _ = listDecl
+listHint _ modu = listDecl exts
+  where
+    exts = nubOrd $ concatMap snd (languagePragmas (pragmas (ghcAnnotations modu)))
 
-listDecl :: LHsDecl GhcPs -> [Idea]
-listDecl x =
-  concatMap (listExp False) (childrenBi x) ++
+listDecl :: [String] -> LHsDecl GhcPs -> [Idea]
+listDecl exts x =
+  concatMap (listExp exts False) (childrenBi x) ++
   stringType x ++
   concatMap listPat (childrenBi x) ++
   concatMap listComp (universeBi x)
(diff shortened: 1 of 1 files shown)


src/Hint/List.hs
```diff
@@ -168,12 +170,12 @@ isAppend :: View a App2 => a -> Bool
...
 
-checks ::[(String, Bool -> LHsExpr GhcPs -> Maybe (LHsExpr GhcPs, [(String, R.SrcSpan)], String))]
-checks = let (*) = (,) in drop1 -- see #174
+checks :: [String] -> [(String, Bool -> LHsExpr GhcPs -> Maybe (LHsExpr GhcPs, [(String, R.SrcSpan)], String))]
+checks exts = let (*) = (,) in drop1 -- see #174
   [ "Use string literal" * useString
-  , "Use list literal" * useList
   , "Use :" * useCons
   ]
+  <> ["Use list literal" * useList | "OverloadedLists" `notElem` exts] -- see #114
```
ndmitchell, line 178: Can we pull out the existence or otherwise of OverloadedLists as a boolean and sent it down from the top, instead of just the full list of extensions.

ndmitchell (commented):
Thanks for the diff, looks good. Definitely wants a test though as its subtle and easy to break. I imagine a line in the `<test>` block at the top of this file will be sufficient - something like:

```
{-# LANGUAGE OverloadedLists #-} \
foo = True:[]
```

src/Hint/List.hs
kk-hainq: Thank you for the review. I've pushed 2 new commits to address both.

src/Hint/List.hs
```diff
@@ -67,11 +69,14 @@ import Language.Haskell.GhclibParserEx.GHC.Types.Name.Reader
 
 
 listHint :: DeclHint
-listHint _ _ = listDecl
+listHint _ modu = listDecl overloadedListsOn
+  where
+    exts = nubOrd $ concatMap snd (languagePragmas (pragmas (ghcAnnotations modu)))
```
ndmitchell, line 74: Why the `nubOrd` here? It is moderately expensive, and I can't see any use for it

ndmitchell (commented):
One final question about a nubOrd, but otherwise looks good

src/Hint/List.hs
kk-hainq: Nice catch! It was originally used to pass down a clean `exts` to the following functions. Don't need it anymore for just `overloadedListsOn`, I have removed it in the latest commit.

ndmitchell:
Thanks!

PR #1274, opened 2021-08-06
Suggest: '(& f)' -> 'f'

mlang:
Thanks for the pull request!

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -840,6 +840,7 @@
     - warn: {lhs: "(mapped .~ b) a", rhs: "b <$ a"}
     - warn: {lhs: "ask <&> (^. a)", rhs: "view a"}
     - warn: {lhs: "view a <&> (^. b)", rhs: "view (a . b)"}
+    - warn: {lhs: "(& f)", rhs: "f"}
 
     # `at` pitfalls:
 


ndmitchell (commented):
Thanks for the hint! I thought we already had this one included in the standard defaults with https://github.com/ndmitchell/hlint/blob/c00bb3d72e88614463983bcbf1b24e5db38e4cb0/data/hlint.yaml#L322 - is that not firing for you for some reason?

mlang:
Mind you, I am totally new to hlint internals.  But indeed, the existing
rule for Data.Function.& (which I missed) does not fire for my test
case:

```
***@***.***:~/Projects/Haskell/hlint$ cat test.hs
module Main where

import Control.Lens

main :: IO ()
main = do
  print $ (& (+ 1)) 0

***@***.***:~/Projects/Haskell/hlint$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'upstream/master'.
***@***.***:~/Projects/Haskell/hlint$ hlint --hint data/hlint.yaml test.hs
No hints
***@***.***:~/Projects/Haskell/hlint$ git checkout redunant
Switched to branch 'redunant'
Your branch is up to date with 'origin/redunant'.
***@***.***:~/Projects/Haskell/hlint$ hlint --hint data/hlint.yaml test.hs
test.hs:7:11-19: Warning: Redundant &
Found:
  (& (+ 1))
  Perhaps:
    (+ 1)

1 hint
```

-- 
CYa,
  ⡍⠁⠗⠊⠕

ndmitchell:
I've removed the `Data.Function` qualification. At the beginning `&` was a bit of a rare function, so I wanted to be careful about when it triggered. Now it's much more common, so I think relaxing it makes a lot more sense. Thanks for the suggestion!

PR #1277, opened 2021-08-11
Fix `derives`

### LINKED ISSUE #1276: Incorrect "Unused LANGUAGE pragma" for DeriveGeneric used with a data family instance.
Hlint appears not to detect uses of deriving clauses in data families (as opposed to regular data types). Test case:

```haskell
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE TypeFamilies  #-}
import GHC.Generics

data Foo

data family Bar a

data instance Bar Foo = Foo
    deriving(Generic)
```

```
$ hlint --version
HLint v3.2.7, (C) Neil Mitchell 2006-2021
$ hlint hlint-testcase.hs 
hlint-testcase.hs:1:1-30: Warning: Unused LANGUAGE pragma
Found:
  {-# LANGUAGE DeriveGeneric #-}
Perhaps you should remove it.

1 hint
```

zliu41:
Fixes #1276. A `DataFamInstDecl` under an `InstDecl` does not have location.


ndmitchell:
Thanks!

Issue #578, opened 2019-01-08
Suggest NumericUnderscores


ndmitchell:
As per https://github.com/haskell/haskell-ide-engine/issues/1029 by @alanz - requires #577 to be implemented first (or to remain on the pile for about 3 years first)

ndmitchell:
True, happy to support that already.

dustin:
I argue that this should only be used for longer numbers.  I see this was also mentioned in https://github.com/haskell/haskell-ide-engine/issues/1029

Having applied this to a codebase with a lot of dates, it's suggested underscores in all of the years in our codebase and satisfying the hint was quite invasive.

Would it be possible to limit this to a minimum number of digits?  Years (below 10,000) never use a thousands separator and style guides often suggest that (at least in American styles), comma separators aren't used for four-digit numbers.

ndmitchell:
@dustin - that seems very sensible - making 2023 turn in to 2_023 seems horrible. Are you interested in writing a patch?

dustin:
@ndmitchell do you have any reservations for my PR?

ndmitchell:
Sorry @dustin, was between computers, this looks great and is all now merged. Would a release be useful? Or shall I just roll it into the next one?

dustin:
Thanks so much.  It sounds like we are fine without a release, so nothing urgent here.

PR #1279, opened 2021-08-12
Suggest NumericUnderscores

epn09:
Addressing #578.

Here, we suggest underscore with chunk size of 3 for decimal literals and 4 for non-decimal literals. We only suggest if `NumericUnderscores` is enabled in the file.

Example:
```
Suggestion: Use underscore
Found:
  123456789.0441234e-123456
Perhaps:
  123_456_789.044_123_4e-123_456

Suggestion: Use underscore
Found:
  0x12abc.523defp+172345
Perhaps:
  0x1_2abc.523d_efp+172_345

Suggestion: Use underscore
Found:
  3.14159265359
Perhaps:
  3.141_592_653_59
```

### DIFF
--- a/src/GHC/Util/ApiAnnotation.hs
+++ b/src/GHC/Util/ApiAnnotation.hs
@@ -3,13 +3,19 @@ module GHC.Util.ApiAnnotation (
     comment, commentText, isCommentMultiline
   , pragmas, flags, languagePragmas
   , mkFlags, mkLanguagePragmas
+  , extensions
 ) where
 
+import GHC.LanguageExtensions.Type (Extension)
 import GHC.Parser.Annotation
 import GHC.Types.SrcLoc
 
+import Language.Haskell.GhclibParserEx.GHC.Driver.Session
+
 import Control.Applicative
 import Data.List.Extra
+import Data.Maybe (mapMaybe)
+import qualified Data.Set as Set
 
 trimCommentStart :: String -> String
 trimCommentStart s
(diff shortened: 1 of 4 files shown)


src/GHC/Util/ApiAnnotation.hs
```diff
@@ -3,13 +3,19 @@ module GHC.Util.ApiAnnotation (
...
 
+import GHC.LanguageExtensions.Type (Extension)
 import GHC.Parser.Annotation
 import GHC.Types.SrcLoc
 
+import Language.Haskell.GhclibParserEx.GHC.Driver.Session
+
 import Control.Applicative
 import Data.List.Extra
+import Data.Maybe (mapMaybe)
```
ndmitchell, line 17: I tend to just import the whole of Data.Maybe and not worry

src/Hint/NumLiteral.hs
```diff
@@ -0,0 +1,106 @@
...
+import Data.List (intercalate)
+import Data.Generics.Uniplate.DataOnly (universeBi)
+import Refact.Types
+
+import Hint.Type (DeclHint, toSS, ghcAnnotations)
+import Idea (Idea, suggest)
+
+numLiteralHint :: DeclHint
+numLiteralHint _ modu decl =
+  if NumericUnderscores `elem` extensions (ghcAnnotations modu) then
```
ndmitchell, line 37: This will check the NumericUnderscores extension one per decl. If you rewrite it by removing the `decl` from the definition on line 36 and making 38 point free and 40 `const []` then it will only compute the extension inclusion once.

ndmitchell (commented):
Looks really good! Just a few minor things to clean up and then will be ready to merge.

epn09:
@ndmitchell I've addressed your suggestions. But I don't really understand why the CI failed.

ndmitchell:
The errors are failing due to the refactoring not managing to cope with those construct. Add `@NoRefactor` after the tests that are failing and it should pass CI. CC @zliu41 as the refactoring expert

epn09:
Thanks for the suggestion, I've added `@NoRefactor` @ndmitchell.

ndmitchell:
Thanks, looks good and will make it to the next release :)

PR #1289, opened 2021-08-30
various proposed rules

jvoigtlaender:
See https://github.com/ndmitchell/hlint/issues/1283.


ndmitchell:
Thanks! All look good

ndmitchell:
In undid the suggestions to use replicate as they change the strictness, aren't subject to foldr/build fusion, and might end up performing worse. In particular `map (const True) [0..]` returns the infinite list of `True`, while `replicate (length [0..]) True` loops forever.

jvoigtlaender:
Oh, that makes sense. :smile:

Didn't even occur to me from the student's code that motivated suggesting the rule. Because she wasn't dealing with infinite lists, of course. Indeed, her original code had been `mapM (const ts) [1 .. n]`, where the length is very obvious and finite.

PR #1294, opened 2021-09-11
Lens hint to use (<>) to append folds

yairchu:
I've personally been using this rule for a while, having it helps discover this neat feature which is explained in [a recent post by Gabriella Gonzalez](https://www.haskellforall.com/2021/09/optics-are-monoids.html)

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -856,6 +856,7 @@
     - warn: {lhs: "(mapped .~ b) a", rhs: "b <$ a"}
     - warn: {lhs: "ask <&> (^. a)", rhs: "view a"}
     - warn: {lhs: "view a <&> (^. b)", rhs: "view (a . b)"}
+    - warn: {lhs: "a ^.. b . c <> a ^.. b . d", rhs: "a ^.. b . (c <> d)"}
 
     # `at` pitfalls:
 


ndmitchell:
If you say that is valid, I'll believe you. Looks like symbol soup to me!

yairchu:
I'll try to explain the symbol soup:

```Haskell
a = [Left 0, Right 1, Left 2, Right 3]

> a ^.. traverse
[Left 0,Right 1,Left 2,Right 3]

> a ^.. traverse . _Right
[1,3]

b = traverse
c = _Right
d = _Left

> a ^.. b . c
[1,3]

> a ^.. b . d
[0,2]

> a ^.. b . c <> a ^.. b . d
[1,3,0,2]

> a ^.. b . (c <> d)
[0,1,2,3]
```

Also in the process of explaining it, I found that my rule has a bug 😬.. Personally I'll still use it because often when I use it I don't always care on the order of enumerated result.

So either we can revert it or add a note that it may change order of results.. Sorry for a premature PR..

ndmitchell:
I've pushed a revert. I think this is one where we definitely want perfect equivalence because:

* Changing the order of a set is super subtle - it's easy to not spot until it makes it to production.
* Lens uses tend to be quite fastidious about laws and properties.
* If we want to encourage people to use these lens hints, it's best for them to have no downsides, which is the case of all the others.

yairchu:
Yeah sounds like the right thing to do.. I've learned a subtlety and will be cautious with tricky rules next time :)

PR #1295, opened 2021-09-27
Consider TemplateHaskell redundant if it can be replaced with TemplateHaskellQuotes

zliu41:
There are reasons to prefer `TemplateHaskellQuotes` to `TemplateHaskell` whenever possible:
- THQ is in Safe Haskell; TH is not.
- Compiling a module with TH enabled always produces the `dyn_o` file, even in `-fno-code` (typechecking only) mode.

Since TH implies THQ, the hint would say
```
Found:
  {-# LANGUAGE TemplateHaskell #-}
Perhaps you should remove it.
Note: may require `{-# LANGUAGE TemplateHaskellQuotes #-}` adding to the top of the file
```

which is exactly what needs to be said.

### DIFF
--- a/src/Hint/Extensions.hs
+++ b/src/Hint/Extensions.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE LambdaCase, NamedFieldPuns #-}
+{-# LANGUAGE LambdaCase, NamedFieldPuns, ScopedTypeVariables #-}
 
 {-
     Suggest removal of unnecessary extensions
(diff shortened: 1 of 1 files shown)


zliu41:
This also happens to fix two other things:
- A hint is now produced on redundant `TemplateHaskellQuotes`.
- When TH is enabled, but neither TH nor THQ is needed, now it no longer says "may require `{-# LANGUAGE TemplateHaskellQuotes #-}` adding to the top of the file".

ndmitchell:
Thanks! Good points in favour

PR #1296, opened 2021-09-27
Remove TemplateHaskellQuotes from baseDynFlags

zliu41:
HSE is no longer used, so THQ is no longer needed.


ndmitchell:
Always nice to simplify! Thanks

PR #1297, opened 2021-09-27
Add some warnings to the cabal file

zliu41:
I've forgot to check for unused imports before opening a PR multiple times. Adding the warnings that are checked in CI to the cabal file would be helpful.

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -165,6 +165,7 @@ library
         Test.Annotations
         Test.InputOutput
         Test.Util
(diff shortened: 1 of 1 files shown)


ndmitchell:
My usual approach was that I wanted me to see these warnings (which I do in CI), but not my users. However, nowadays stack and cabal don't display warnings in dependencies, so seems no reason not to.

ndmitchell:
Thanks!

PR #1301, opened 2021-10-05
Add GHC libdir to ci cache key

zliu41:
Just to see what happens


ndmitchell:
Thanks! Let's see how this does

Issue #1298, opened 2021-09-28
Regression: many hints no longer shown


Mikolaj:
Since probably around 70 days ago hlint stopped showing some hints, e.g., the following are not shown with the current version, but are shown with hlint compiled with cabal index from 70 days ago (master branch of https://github.com/LambdaHack/LambdaHack, using `.hlint.yaml` from there):

```
./engine-src/Game/LambdaHack/Server/HandleRequestM.hs:1055:13: Warning: Redundant if
Found:
  if absentPermitted then True else k == kOld
Why not:
  absentPermitted Prelude.|| (k == kOld)

./engine-src/Game/LambdaHack/Server/ItemM.hs:224:62: Warning: Redundant do
Found:
  do when (IA.checkFlag Ability.Unique arItem) $
       modifyServer $
         \ ser -> ser{suniqueSet = ES.insert itemKindId (suniqueSet ser)}
Why not:
  when (IA.checkFlag Ability.Unique arItem) $
    modifyServer $
      \ ser -> ser{suniqueSet = ES.insert itemKindId (suniqueSet ser)}

./engine-src/Game/LambdaHack/Server/ItemRev.hs:192:28: Warning: Redundant do
Found:
  do if length flvs < 6 then
       do flavour <- oneOf flvs
          return (EM.insert key flavour assocs, availableMap)
       else
       do let available = availableMap EM.! IK.isymbol ik
              proper = ES.fromList flvs `ES.intersection` available
          assert
            (not (ES.null proper) `blame` "not enough flavours for items"
               `swith` (flvs, available, ik, availableMap))
            $
            do flavour <- oneOf $ ES.elems proper
               let availableReduced = ES.delete flavour available
               return
                 (EM.insert key flavour assocs,
                  EM.insert (IK.isymbol ik) availableReduced availableMap)
Why not:
  if length flvs < 6 then
    do flavour <- oneOf flvs
       return (EM.insert key flavour assocs, availableMap)
    else
    do let available = availableMap EM.! IK.isymbol ik
           proper = ES.fromList flvs `ES.intersection` available
       assert
         (not (ES.null proper) `blame` "not enough flavours for items"
            `swith` (flvs, available, ik, availableMap))
         $
         do flavour <- oneOf $ ES.elems proper
            let availableReduced = ES.delete flavour available
            return
              (EM.insert key flavour assocs,
               EM.insert (IK.isymbol ik) availableReduced availableMap)

[...]

PR #1303, opened 2021-10-12
Assume a name is from Prelude unless we know where it is from

zliu41:
This seems a reasonable short term improvement before we decide how to re-do scoping properly. Reason for doing this:

- Sometimes people disable Prelude by `import Prelude ()`, and re-export some of it in a custom Prelude. This would make many hints stop firing because they rely on Prelude to be in scope.
  - (The problem of re-exporting names is not specific to Prelude, but I'd imagine Prelude is much more often re-exported than say `Data.Map`)
- This also makes HLint behave the same way wrt `import Prelude ()` and `{-# LANGUAGE NoImplicitPrelude #-}`.

#1298 has some discussion.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -1281,6 +1281,17 @@
 # import Prelude((==)) \
 # import qualified Prelude as P \
 # main = P.length xs == 0 -- P.null xs
+# import Prelude () \
+# main = length xs == 0 -- null xs
+# import Prelude () \
+# import Foo \
+# main = length xs == 0 -- null xs
+# import Prelude () \
+# import Data.Text (length) \
+# main = length xs == 0
+# import Prelude () \
+# import qualified Data.Text (length) \
+# main = length xs == 0 -- null xs
 # main = hello .~ Just 12 -- hello ?~ 12
 # foo = liftIO $ window `on` deleteEvent $ do a; b
 # no = sort <$> f input `shouldBe` sort <$> x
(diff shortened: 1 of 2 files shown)


ndmitchell:
This seems pretty plausible. Can you rebase over your other patch, and then I'll give it a merge?

zliu41:
@ndmitchell rebased. I modified the approach, so that now it adds Prelude to the possible modules of a name unless we are sure where the name is from (via qualified or explicit import). Otherwise, hints still won't fire with `import Prelude ()` plus any bare import.

Mikolaj:
I've built the branch and got extra 91 hints in my project, so this solves the original issue.

Mikolaj:
The only glitch, I don't know if related, is hints such as

```
engine-src/Game/LambdaHack/Client/UI/DrawM.hs:427:35-66: Suggestion: Use fmap
Found:
  \ aid -> Just <$> getCacheBfs aid
Perhaps:
  Char.fmap Just Char.. getCacheBfs

engine-src/Game/LambdaHack/Client/UI/DrawM.hs:764:36-51: Warning: Use break
Found:
  span (not . fst)
Perhaps:
  Char.break fst
```

This from a run on https://github.com/LambdaHack/LambdaHack/commit/100690c5d2eabaca985f44d220e6d660812e21d1

Mikolaj:
Could you merge this? I can file the `Char.break` problem as a separate issue, but fixing the lack of so many hints for my contributors (who don't use a dev version of hlint) would be very helpful to me.

Mikolaj:
I checked and the `Char.break` problem completely disappears when I replace the qualified import of `Data.Char` with unqualified, despite many other qualified imports remaining in the module. And there's really nothing so strange in `Data.Char`:

https://hackage.haskell.org/package/base-4.14.3.0/docs/src/Data-Char.html

except for 

```
{-# LANGUAGE Trustworthy #-}
{-# LANGUAGE NoImplicitPrelude #-}
```

zliu41:
@ndmitchell I've rebased this on top of master and ran the tests.

ndmitchell:
Thanks! I would be surprised if this doesn't make something worse, but it clearly makes a lot of things better, so let's roll the dice on it.

Mikolaj:
Thank you. :D

Issue #1308, opened 2021-10-20
Please, add rule: (either Left) == (=<<)


Anton-Latukha:
Noticed that `=<<` needs to be recommended there.

Generally, all these kinds of `switcher mzero f x` Monad functions that does not change the outside Monad - reduce into `=<<`, since:
1. Canonical Lambda argument order for binary control path `switcher`-type function is to pass case for `mzero` first.
2. In case of that being same-monad `mzero` that gets analyzed (`x`, `Left` in this case) - it just reconstructs the `mzero` of `x`.
3. `mzero` literal defining law/property is that it short-circuits `=<<`. So `(f =<< mzero) == mzero`, which inside the context of `Either` `Monad` `Left`s seems equivalent to `either Left f (Left a)`.
4. Doing typecheck:

```haskell
-- ghci>
  
  :t (either Left)
    
    (either Left)
      :: (b1 -> Either a b2) -> Either a b1 -> Either a b2
    
    
  :t ((=<<) :: (b1 -> Either a b2) -> Either a b1 -> Either a b2)
    
    ((=<<) :: (b1 -> Either a b2) -> Either a b1 -> Either a b2)
      :: (b1 -> Either a b2) -> Either a b1 -> Either a b2
        
-- the type of (either Left f x) equal to (f =<< x), in the context of Either monad, which was needed to be proven.
```

5. I am not proficient in Hask bottom to elaborate on them.
6. `f =<< x` makes code from concrete `Either` into polymorphic `Monad`, which is generally regarded as a good thing, but, just as `maybe Nothing -> =<<`, with `x` changing the type would act accordingly to literal semantics of code.
7. The separation of concerns is one of the main code paradigms, the main separation of concerns in Haskell language design is the separation of code into 3 code layers: {term, type, kind}: so types are in the type code. `either` is a term-level function that enforces concrete type, which by comparison between `(either Left) <=> (=<<)` in `either` is the violation of separation on concerns, as it enforces the concrete type inside the term level implicitly, & that approach requires a new function for every type to enforce it so it is unsustainable language design & approach (on the big scale): if to consider the memoization & use of mapping `(endless set of concrete types -> requires even faster-growing endless set of function sets)`.
[...]

PR #1309, opened 2021-10-20
Rule (either Left) is (=<<)

Anton-Latukha:
Closes: #1308

Declaring it as a `warn` level also, as the other entries in the same group & as `=<<` is a
strong simplification & a major gateway into HLint recursive optimizations from it.


ndmitchell:
Thanks for the PR! For something like this I usually eyeball it and am happy to merge it without necessarily passing the test suite, so if you publish it, I'll merge :)

Anton-Latukha:
As already have the ability to pass the CI - just rebased & passed it. I also like to look at what projects have in CIs.

Done.

ndmitchell:
Thanks, all merged

PR #1312, opened 2021-10-31
Update to ghc-8.10.7

hseg:
Latest minor version bump. Testsuite runs correctly.

Motivation is just not wanting to have to have old GHCs installed, can just ship
this patch in my packaging if not merged. Will do so for now regardless.


ndmitchell:
Thanks!

PR #1315, opened 2021-11-13
add return -> pure as default a hint, active phase of proposal 0314 

Anton-Latukha:
Enabling suggestion by default, allows at once to deduplicate `return` & `pure` rewriting rules.

Seems because of that deduplication - now additional chains of rules kick in, so now `cabal v2-run -- hlint --test` not passes & idk what to do here.

Also `cabal v2-run hlint -- --generate-summary` seems to cache the old generation for `hints.md`, so changes in `yaml` not appear in `hints.md` (I imagine) because of some caching somewhere.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -278,7 +278,7 @@
 
     # FOLDS
 
-    - warn: {lhs: foldr  (>>) (return ()), rhs: sequence_}
+    - warn: {lhs: foldr  (>>) (pure ()), rhs: sequence_}
     - warn: {lhs: foldr  (&&) True, rhs: and}
     - warn: {lhs: foldl  (&&) True, rhs: and, note: IncreasesLaziness}
     - warn: {lhs: foldr1 (&&) , rhs: and, note: "RemovesError on `[]`"}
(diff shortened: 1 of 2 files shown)


Anton-Latukha:
If I would not reach my hands & mind to debug all what is going on in time before review - would say thanks to concise information/comments on project operation & what needs to be done to merge properly.

ndmitchell:
Sorry for the delay in response, I've been a bit ill over the last bit of time.

My understanding is the proposal mentioned warns about definitions in instances, whereas the patch here moves uses. Is that right?

Anton-Latukha:
> My understanding is the proposal mentioned warns about definitions in instances, whereas the patch here moves uses. Is that right?

Yes. Correct understanding.

---

The PR has two actions:
1. The `return -> pure` rule. Currently submitting it as a `hint`.
2. The rule graph gets adjusted.

---

I think I should reopen the PR. Seing there are 3 distinct phases to this.

ndmitchell:
I think the intent is pretty clear, so happy to discuss this in terms of words, without adjusting the code.

The proposal you link to doesn't mention suggesting people use pure instead of return. Is there community agreement that is the end state?

Anton-Latukha:
Yes, that proposal is a part of [AMP/"Monad of no return"](https://gitlab.haskell.org/ghc/ghc/-/wikis/proposal/monad-of-no-return) process, which does have it agreed in the pipeline.

Logically next stage would be `return` would be made a regular top-level binding alias to `pure`.

The stage after that is for GHC to:
> Start warning about lawful `return/>>` method overrides (in order to prepare for Phase 4) *(removal of `return/>>` in distant future)*.

So it is in the execution & process goes further either way, but having/suggestion `return -> pure` in advance would allow having softer adoption & migration process.

andreasabel:
> Is there community agreement that is the end state?

Na, I won't think so.  I prefer `return` over `pure` at the end of monadic functions as it is more suggestive there, and in line with other programming languages (such as Java).

PR #1325, opened 2021-12-10
#1324 add createModuleExWithFixities

### LINKED ISSUE #1324: A way to createModuleEx with config fixities?
In HLSs hlint plugin, if the flag `HLINT_ON_GHC_LIB` is not defined then it uses `createModuleEx` to create a `ModuleEx` but this ignores fixities found in the `.hlint.yaml` ([#2418](https://github.com/haskell/haskell-language-server/issues/2418#issue-1066910078)). I figure adding a function like
```haskell
createModuleWithFixities :: [(String, Fixity)] -> ApiAnns -> Located HsModule -> ModuleEx
createModuleWithFixities fixities anns ast = 
  ModuleEx (applyFixities (fixitiesFromModule ast ++ fixities) ast) anns
```
would allow us to use `ApiAnns` from the ghc api, and also use fixities from the `.hlint.yaml` file, at the small expense of duplicating the `toFixity` function to convert `ParseFlags` to `[(String, Fixity)]`.

Alternatively we could have `createModuleWithParseFlags` take a `ParseFlags` and use the not exposed `ghcFixitiesFromParseFlags` to do the work, and have no duplication of code on the client side.

Or we could also expose `ghcFixitiesFromParseFlags` or `Fixity.toFixity` but meh.

eddiemundo:
Fixes #1324

### DIFF
--- a/src/GHC/All.hs
+++ b/src/GHC/All.hs
@@ -5,7 +5,7 @@ module GHC.All(
     CppFlags(..), ParseFlags(..), defaultParseFlags,
     parseFlagsAddFixities, parseFlagsSetLanguage,
     ParseError(..), ModuleEx(..),
-    parseModuleEx, createModuleEx, ghcComments, modComments,
+    parseModuleEx, createModuleEx, createModuleExWithFixities, ghcComments, modComments,
     parseExpGhcWithMode, parseImportDeclGhcWithMode, parseDeclGhcWithMode,
     ) where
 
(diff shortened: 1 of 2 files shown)


src/GHC/All.hs
```diff
@@ -156,6 +156,10 @@ createModuleEx :: ApiAnns -> Located HsModule -> ModuleEx
 createModuleEx anns ast =
   ModuleEx (applyFixities (fixitiesFromModule ast ++ map toFixity defaultFixities) ast) anns
```
ndmitchell, line 157: Could you redefine `createModuleEx = createModuleExWithFixities defaultFixities`

ndmitchell (commented):
Code looks good. I'll hold off on merging it until we get the GHC 9.2 transition completed, but looks good to merge then. Hopefully won't take too long (a few days)

ndmitchell:
Could you also rebase following the move to GHC 9.2.

ndmitchell:
Thanks - will be in the next release, which needs to wait for GHC 9.2 compatibility support, so might be a month or two.

Issue #1319, opened 2021-11-30
how to enable one specific ruleset?


jwaldmann:
How do I enable just the warnings/suggestions provided by the rules from one file? 
(https://github.com/NorfairKing/haskell-dangerous-functions/issues/18)

I tried
```
hlint --hint=hlint.yaml src/ 
```
but this will use hlint's built-in rules as well - I think that's the meaing of the first to lines of `.. -v ..` output
```
# Initialise of global flags... took 0.00s
# Config of data/hlint.yaml... took 0.17s
# Config of hlint.yaml... took 0.00s
# Parse of src/...
```

I tried to put `- ignore: { }` at the start, but then I get no output at all (I think that's https://github.com/ndmitchell/hlint/issues/748)

zliu41:
Currently, if you put `- ignore: { }` at the start, then you must explicitly turn on any hint you want.
In other words, this does not produce any hint:
```
- ignore: { }
- warn: {lhs: mapM, rhs: traverse}
```

You need to turn on the hint by

```
- ignore: { }
- warn: {name: Use traverse}
- warn: {lhs: mapM, rhs: traverse}
```

This is obviously redundant, and what's worse, the hint name "Use traverse" is auto-generated, and may not be obvious to whoever created the hint. I think we should make it so that whatever follows `- ignore` is turned on automatically.

ndmitchell:
That seems a very reasonable solution @zliu41  - patch welcome

PR #1330, opened 2021-12-15
Make `HintRule` imply `Classify` in the hint file

zliu41:
This change makes it so that a `Classify` is automatically added for each rule in the hint file. For example, `- error: {lhs: mapM, rhs: traverse}` now implies `- error: {name: Use traverse}`. See #1319 for motivation.

This only applies to the user hint file, and not `data/hlint.yaml`, for the following reasons:
- Efficiency. There's no need to generate a ton of `Classify`s for the rules in `data/hlint.yaml`.
- There are multiple rules with the same name, e.g., there's a builtin "Use join" which is a suggestion, and a few other "Use join" rules in `data/hlint.yaml` which are warnings. Applying the change to `data/hlint.yaml` would cause "Use join" to always be warnings.

Also, currently `Classify`s are only generated for rules. Other types of configs, such as
```
- modules:
  - {name: Control.Arrow, within: []}
```
does not generate a `- warn: {name: Avoid restricted module}`. This still needs to be added explicitly if needed, but at least the number is bounded.

### DIFF
--- a/data/hintrule-implies-classify.yaml
+++ b/data/hintrule-implies-classify.yaml
@@ -1,0 +1,2 @@
+- ignore: { }
+- warn: {lhs: mapM, rhs: traverse}
--- a/src/Config/Yaml.hs
+++ b/src/Config/Yaml.hs
@@ -4,14 +4,17 @@
 
 module Config.Yaml(
     ConfigYaml,
+    ConfigYamlBuiltin (..),
+    ConfigYamlUser (..),
     readFileConfigYaml,
-    settingsFromConfigYaml
+    settingsFromConfigYaml,
+    isBuiltinYaml,
     ) where
 
 import GHC.Driver.Ppr
 import GHC.Parser.Errors.Ppr
 import Config.Type
-import Data.Either
+import Data.Either.Extra
 import Data.Maybe
 import Data.List.Extra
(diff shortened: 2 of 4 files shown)


ndmitchell:
Thanks for the patch. This seems a reasonable way to do it. I'm slightly concerned that things like:

```
warn: {name: foo, lhs: ...}
ignore: {name: foo}
error: {name: foo, lhs: ...}
```

Turns on the first lint at error severity. But I think we've got a slightly weird situation where hints have some kind of identity by reference when they are being defined, but thereafter, only have identity by name. As a result, it seems like this is a pretty reasonable route to take, even if it has a few unexpected corners.

Could you add an end to end test (just another one under `tests/`) capturing the behaviour of #1319 so that we keep this working for the specific case we have in mind in future.

zliu41:
@ndmitchell Test added!
I think we can make it so that the `- error: {name: foo}` automatically added by `error: {name: foo, lhs: ...}` only applies to this particular rule. It probably requires a much bigger change, but am happy to open a separate PR for that.

ndmitchell:
Thanks! Happy to leave it as the slightly simpler form for now - I'm not totally sure what the right approach is, but doing less work doesn't seem unreasonable as a default.

PR #1331, opened 2021-12-20
Rules to use min and max

yairchu:
This is based on real [advent of] code of a friend, and can serve to teach new folks about `min` and `max`


ndmitchell:
Thanks!

PR #1332, opened 2021-12-20
Suggest using iterate instead of scanl/foldl where suitable

yairchu:
Based on situations in real [advent of] code of a friend


ndmitchell:
Thanks!

PR #1336, opened 2021-12-25
Hint for unused OverloadedRecordDot

shayne-fletcher:
Hint if extension `OverloadedRecordDot` is enabled but not used.


ndmitchell:
Thanks!

Issue #1232, opened 2021-04-01
Loading settings differs between library and executable


tfausak:
First, some pointers. I discovered this issue while investigating a bug against Splint, a GHC plugin that uses HLint: https://github.com/tfausak/splint/issues/9. I think this issue is related to both #143 and #98. 

The shortest way to state this problem is that I think calling [`autoSettings`](https://hackage.haskell.org/package/hlint-3.3/docs/Language-Haskell-HLint.html#v:autoSettings) (or `argsSettings`) behaves differently than calling [`hlint`](https://hackage.haskell.org/package/hlint-3.3/docs/Language-Haskell-HLint.html#v:hlint). In particular any arguments specified in the configuration file seem to be ignored. For example, consider the following Haskell module:

``` hs
-- Main.hs
import Text.RawString.QQ -- from raw-strings-qq
main = print [r||]
```

Clearly it needs the `QuasiQuotes` language extension in order to compile. No problem, you can add that to your HLint configuration. To wit:

``` yaml
# .hlint.yaml
- arguments: [ -XQuasiQuotes ]
```

And indeed everything works fine that way. You can run `hlint` through your shell, or you can run `Language.Haskell.HLint.hlint` and it will succeed. 

The problem comes when you try to split apart reading the settings from linting the file. I'm not terribly familiar with HLint's internals, but it felt right to call `autoSettings` followed by `parseModuleEx` to roughly recreate HLint's command-line interface. Unfortunately that does not work:

``` hs
>>> (parseFlags, _, _) <- autoSettings
>>> result <- parseModuleEx parseFlags "Main.hs" Nothing
>>> :sprint result
result = Left (ParseError _ _ _)
```

You can see that it fails because it tries to parse the module without quasi-quotes enabled. Explicitly enabling quasi-quotes works fine:

``` hs
>>> (parseFlags, _, _) <- argsSettings [ "-XQuasiQuotes" ]
>>> result <- parseModuleEx parseFlags "Main.hs" Nothing
>>> :sprint result
result = Right (hlint-3.1.6:GHC.All.ModuleEx _ (_,_))
```

Clearly one solution to this problem is to simply call `hlint` and avoid loading the settings separately. For my use case that's not ideal, since loading the settings takes a small but appreciable amount of time. I'd prefer to load the settings once, and then lint a bunch of files with those settings. 

[...]

PR #1340, opened 2022-01-16
Include Restriction hints in splitSettings (and argsSettings) output

tgiannak:
This is a mitigation for the issue identified in #1232.


ndmitchell:
Thanks!

PR #1342, opened 2022-02-01
Fix handling of wildcards and add extensive tests

### LINKED ISSUE #402: Allow module wildcards

tfausak:
This builds off my previous work in #1149 and #1221, both of which aimed to fix #402. Hopefully this PR can be the last in the series 😆 

Recently I was trying to use module wildcards with HLint, both for matching imported modules and for matching which modules rules should apply to. While the former worked, I was distressed to see that the latter did not. This confused me since I thought I had implemented this feature myself. Turns out I never tested it! 

This PR fixes module wildcard handling and adds tests for all the cases I can think of right now.


tfausak:
@ndmitchell can you review this? The diff may suggest that this is a big change, but really only a couple lines of code changed. Everything else is tests to make sure that the wildcard matches work properly in a variety of situations.

tests/wildcards.test
```diff
@@ -0,0 +1,755 @@
...
+RUN tests/wildcard-xr.hs --hint=data/wildcards.yaml
+FILE tests/wildcard-xr.hs
+module XR where x = read ""
+OUTPUT
+No hints
+
+----
+RUN tests/wildcard-x-r.hs --hint=data/wildcards.yaml
+FILE tests/wildcard-x-r.hs
+module X.R where x = read ""
```
zliu41, line 675: So `*R` doesn't match `X.R` - is this intentional? It seems counter intuitive.

data/wildcards.yaml
```diff
@@ -0,0 +1,30 @@
+- modules:
+  - { name: A, as: A }
+  - { name: '*B', as: B }
+  - { name: '**.C', as: C }
+  - { name: '**.*D', as: D }
+  - { name: E, within: E }
+  - { name: F, within: '*F' }
+  - { name: G, within: '**.G' }
+  - { name: H, within: '**.*H' }
```
zliu41, line 9: Perhaps add a rule where both `name` and `within` contain wildcards?

tests/wildcards.test
tfausak: Yes, that's intentional. The module wildcard matching follows the same rules as file globbing. `*R` would match `R` and `XR` but not `X.R`. Along the same lines, `**.R` would match `R` and `X.R` but not `XR`.

tfausak: Ah yeah, I can see why that's confusing. The module matching is like file globbing, but importantly `.` characters in module names are treated like `/` characters in file paths. 

```
-- file glob
**/*Spec.hs
  Spec.hs -- match
  FooSpec.hs -- match
  Foo/BarSpec.hs -- match

-- module glob
**.*Spec
  Spec -- match
  FooSpec -- match
  Foo.BarSpec -- match

ndmitchell:
Thanks for the awesome tests, and sorry it took so long to review!

tfausak:
No problem at all. Thanks!

PR #1348, opened 2022-02-13
Improve module restrict hints

lierdakil:
I'm not updating the readme for now, since I want some general feedback before doing any more work.

- Add new fields to module restrict config:
    - asRequired: Bool; if true, `as` alias is required. Ignored if `as`
      is empty.
    - ~~qualified: either 'always', 'never', 'exhaustive' or 'ignore';
      should the module be qualified? 'exhaustive' only requires
      qualification if there's no explicit import list, e.g.
      `import Foo (a, b, c)` is fine, but `import Bar hiding (d)` and
      `import Baz` will generate a hint.~~
    - importStyle: one of qualified, unqualified, explicit, explicitOrQualified, ignore.
      The preferred import style. `explicitOrQualified` accepts both `import Foo (a,b,c)`
      and `import qualified Foo`, but not `import Foo` or `import Foo hiding (x)`.
      `explicit` is basically the same, but doesn't accept `import qualified`
      (I'm not sure if it's very useful tbh). `qualified` and `unqualified` do not care
      about the import list at all.
    - qualifiedStyle: either 'pre', 'post' or 'ignore'; how should the
      module be qualified?

- Wildcard match handling is extended, such that restrictions are
  unified between wildcard and specific matches. With the new
  fields, the more specific option takes precedence. The list fields are
  merged. With multiple wildcard matches, the precedence between them is
  not guaranteed (but in practice, names are sorted in the
  reverse lexicograpicall order, and the first one wins -- which
  hopefully means the more specific one more often than not)

  This hopefully explains the reason for 'ignore' option: we can
  override the wildcard match with it.

  The primary motivation for this is to be able to specify the default
  pre/post style with something like

  ```
  - {name: ['**'], qualifiedStyle: post}
  ```

  But it does also add a fair bit of flexibility for writing rules (at the cost of some complexity).

  If there's a better way to do this, let me know.

If the same module is specified multiple times, for the new fields, only
the first definition will take effect.

If you're wondering why `asRequired` is a `Bool`, and not `always/never/ignore`, that's because `false` is `ignore` here, and there's no `never` option. One can already effectively forbid aliasing by specifying something like `name: Data.Map, as: Data.Map`.

### DIFF
--- a/src/Config/Type.hs
+++ b/src/Config/Type.hs
@@ -2,11 +2,13 @@
 module Config.Type(
     Severity(..), Classify(..), HintRule(..), Note(..), Setting(..),
     Restrict(..), RestrictType(..), RestrictIdents(..), SmellType(..),
+    RestrictImportStyle(..), QualifiedStyle(..),
     defaultHintName, isUnifyVar, showNotes, getSeverity, getRestrictType, getSmellType
     ) where
 
 import Data.Char
 import Data.List.Extra
+import Data.Monoid
 import Prelude
 
 
@@ -119,11 +121,28 @@ instance Semigroup RestrictIdents where
     OnlyIdents x1 <> OnlyIdents x2 = OnlyIdents $ x1 <> x2
     ri1 <> ri2 = error $ "Incompatible restrictions: " ++ show (ri1, ri2)
 
+data RestrictImportStyle
+  = ImportStyleQualified
+  | ImportStyleUnqualified
+  | ImportStyleExplicit
+  | ImportStyleExplicitOrQualified
+  | ImportStyleUnrestricted
+  deriving Show
+
+data QualifiedStyle
+  = QualifiedStylePre
+  | QualifiedStylePost
+  | QualifiedStyleUnrestricted
+  deriving Show
+
 data Restrict = Restrict
     {restrictType :: RestrictType
     ,restrictDefault :: Bool
     ,restrictName :: [String]
     ,restrictAs :: [String] -- for RestrictModule only, what module names you can import it as
+    ,restrictAsRequired :: Alt Maybe Bool -- for RestrictModule only
+    ,restrictImportStyle :: Alt Maybe RestrictImportStyle -- for RestrictModule only
+    ,restrictQualifiedStyle :: Alt Maybe QualifiedStyle -- for RestrictModule only
     ,restrictWithin :: [(String, String)]
     ,restrictIdents :: RestrictIdents -- for RestrictModule only, what identifiers can be imported from it
     ,restrictMessage :: Maybe String
--- a/src/Config/Yaml.hs
+++ b/src/Config/Yaml.hs
@@ -39,6 +39,7 @@ import Fixity
 import Extension
 import GHC.Unit.Module
 import Data.Functor
+import Data.Monoid
 import Data.Semigroup
 import Timing
 import Prelude
@@ -192,6 +193,12 @@ maybeParse :: (Val -> Parser a) -> Maybe Val -> Parser (Maybe a)
 maybeParse parseValue Nothing = pure Nothing
 maybeParse parseValue (Just value) = Just <$> parseValue value
 
+maybeParseEnum :: [(T.Text, a)] -> Maybe Val -> Parser (Maybe a)
+maybeParseEnum _ Nothing = pure Nothing
+maybeParseEnum dict (Just val) = case getVal val of
+  String str | Just x <- lookup str dict -> pure $ Just x
+  _ -> parseFail val . T.unpack $ "expected '" <> T.intercalate "', '" (fst <$> dict) <> "'"
+
 parseBool :: Val -> Parser Bool
 parseBool (getVal -> Bool b) = pure b
 parseBool v = parseFail v "Expected a Bool"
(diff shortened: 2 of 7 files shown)


lierdakil:
Side note: could also tackle https://github.com/ndmitchell/hlint/issues/769 here to some extent, e.g. by adding `qualified: 'exhaustive'` option, which would generate the hint only if the import doesn't have an explicit import list.

.hlint.yaml
```diff
@@ -28,6 +28,11 @@
   - {name: [Data.Set, Data.HashSet], as: Set}
   - {name: [Data.Map, Data.HashMap.Strict, Data.HashMap.Lazy], as: Map}
   - {name: Control.Arrow, within: []}
+  - {name: HypotheticalModule1, as: HM1, asRequired: true}
```
zliu41, line 31: This file should not be used for testing. You can create a separate yaml file under `data`, and use it in a `test` file under `tests`. See `data/wildcard.yaml` and `tests/wildcard.test` for an example.

.hlint.yaml
```diff
@@ -28,6 +28,11 @@
...
+  - {name: HypotheticalModule2, qualified: exhaustive}
```
zliu41, line 32: I'm not sure the meaning of `qualified: exhaustive` is immediately obvious. Perhaps `import_style: qualified / explicit / qualified_or_explicit`?

zliu41 (commented):
I like this - rules for enforcing import styles can be useful for many Haskell projects.

.hlint.yaml
lierdakil: :thinking: with the proposed `import_style` we lose the ability to just qualification without requiring an explicit import list. That said, I don't know if it's particularly useful.

lierdakil: I've made the changes. Not sure if `unqualified` and `explicit` are particularly useful, but maybe there's a use case for those I didn't think of.

zliu41 (commented):
This looks good to me. I've always wanted to use HLint to enforce qualified or explicit import, and this does the job. @ndmitchell could you review this?

.hlint.yaml
zliu41: > unqualified

I'd suggest removing this - "This module must be imported unqualified!" says no one, ever :smile:

lierdakil: > says no one, ever

I would avoid cutting obvious functionality just because you can't think of a use case from the top of your head, it ultimately only leads to user frustration. It's not particularly costly to implement, and I'd like to have the option on the off-chance this comes in handy. One example where I imagine this might be somewhat useful is, say, custom prelude imports, which you might want to always be unqualified, but not necessarily always explicit.

ndmitchell:
This conflicts with #1342, which renamed it to wildcard-imports.test and rewrote most of that file. Can you reapply there?

src/Config/Type.hs
```diff
@@ -119,11 +121,28 @@ instance Semigroup RestrictIdents where
     OnlyIdents x1 <> OnlyIdents x2 = OnlyIdents $ x1 <> x2
     ri1 <> ri2 = error $ "Incompatible restrictions: " ++ show (ri1, ri2)
 
+data RestrictImportStyle
+  = ImportStyleQualified
+  | ImportStyleUnqualified
+  | ImportStyleExplicit
+  | ImportStyleExplicitOrQualified
+  | ImportStyleIgnore
```
ndmitchell, line 129: I guess Ignore means "I don't care" - would `ImportStyleUnrestricted` be a better name?

src/Config/Type.hs
```diff
@@ -119,11 +121,28 @@ instance Semigroup RestrictIdents where
...
+  deriving Show
+
+data QualifiedStyle
+  = QualifiedStylePre
+  | QualifiedStylePost
+  | QualifiedStyleIgnore
```
ndmitchell, line 135: Ditto, QualifiedStyleUnrestricted?

Although I wonder how useful qualified style is? There is a GHC warning to require post qualified, so wouldn't people just use that? Although I'm not against it.

src/Config/Type.hs
```diff
@@ -119,11 +121,28 @@ instance Semigroup RestrictIdents where
...
+  deriving Show
+
 data Restrict = Restrict
     {restrictType :: RestrictType
     ,restrictDefault :: Bool
     ,restrictName :: [String]
     ,restrictAs :: [String] -- for RestrictModule only, what module names you can import it as
+    ,restrictAsRequired :: Alt Maybe Bool -- for RestrictModule only
+    ,restrictImportStyle :: Alt Maybe RestrictImportStyle -- for RestrictModule only
+    ,restrictQualifiedStyle :: Alt Maybe QualifiedStyle -- for RestrictModule only
```
ndmitchell, line 145: Why Alt rather than just `Maybe QualifiedStyle`?

src/Config/Type.hs
ndmitchell: That said, if we want ignore as the API for users in YAML, this seems fine. I think I might be ambivalent

src/Config/Yaml.hs
```diff
@@ -322,11 +323,27 @@ parseRestrict restrictType v = do
...
             b <- parseBool def
             allowFields v ["default"]
-            pure $ Restrict restrictType b [] [] [] NoRestrictIdents Nothing
+            pure $ Restrict restrictType b [] mempty mempty mempty mempty [] NoRestrictIdents Nothing
         Nothing -> do
             restrictName <- parseFieldOpt "name" v >>= maybe (pure []) parseArrayString
             restrictWithin <- parseFieldOpt "within" v >>= maybe (pure [("","")]) (parseArray >=> concatMapM parseWithin)
             restrictAs <- parseFieldOpt "as" v >>= maybe (pure []) parseArrayString
+            restrictAsRequired <- parseFieldOpt "asRequired" v >>= fmap Alt . maybeParse parseBool
+            restrictImportStyle <- parseFieldOpt "importStyle" v >>= fmap Alt . maybeParse (\case
```
ndmitchell, line 332: Would be nice to drag the matching of a list of explicit strings into a enum into a helper function, taking a list of string and value pairs. That way we could guarantee the parseFail suggested list was always perfectly correct

ndmitchell (approved):
Sorry it took so long to review! Looks good to me. For Alt, happy either way. For Ignore vs unrestricted, I'll let you decide, whatever you prefer. The tests need fixing after the previous diff merge, but once they are good I'll merge.

.hlint.yaml
ndmitchell: I actually quite like saying modules must be imported unqualified. Every time I see `import Data.List(sort)` I reply with just don't bother listing explicit imports. So I think it's possibly something I'd use 😄 .

src/Config/Type.hs
lierdakil: As I outlined in the OP, the primary reason for having this is not as much to have a warning (as you say, GHC already has a warning for this), as to be able to specify the default for HLint's hints. It would be pretty annoying if HLint says to use `import qualified X` and then GHC complains that it should be `import X qualified` instead, I think. And since we already have the option, adding a check is literally two lines of code.

src/Config/Type.hs
lierdakil: Because I want to use `sconcat` for `RestrictedItem`s. Granted, I could define `Semigroup RestrictedItem` in terms of `<|>`, but defining `<>` in terms of `<|>` feels a bit weird to me.

lierdakil:
Okay, I've rebased and applied some suggested changes (see https://github.com/ndmitchell/hlint/commit/e215ae33ea008854666c2dc654f197125e189b59), and also updated the README.

ndmitchell:
Thanks for all your changes, looks great

Issue #1314, opened 2021-11-01
Can't compile with GHC 9.2.1


LeventErkok:
I can't quite tell if this is a problem with `hlint` itself or maybe the dependencies went out of sync, but I can no longer compile `hlint` with GHC 9.2.1:

```
Building library for hlint-3.3.4..
[ 1 of 59] Compiling EmbedData        ( src/EmbedData.hs, dist/build/EmbedData.o, dist/build/EmbedData.dyn_o )
[ 2 of 59] Compiling Extension        ( src/Extension.hs, dist/build/Extension.o, dist/build/Extension.dyn_o )

src/Extension.hs:64:25: error:
    • Couldn't match type ‘ghc-boot-th-9.2.1:GHC.LanguageExtensions.Type.Extension’
                     with ‘Extension’
      Expected: [(Extension, ([Extension], [Extension]))]
        Actual: [(ghc-boot-th-9.2.1:GHC.LanguageExtensions.Type.Extension,
                  ([ghc-boot-th-9.2.1:GHC.LanguageExtensions.Type.Extension],
                   [ghc-boot-th-9.2.1:GHC.LanguageExtensions.Type.Extension]))]
      NB: ‘Extension’
            is defined in ‘GHC.LanguageExtensions.Type’
                in package ‘ghc-lib-parser-9.2.1.20211030’
          ‘ghc-boot-th-9.2.1:GHC.LanguageExtensions.Type.Extension’
            is defined in ‘GHC.LanguageExtensions.Type’
                in package ‘ghc-boot-th-9.2.1’
    • In the expression: GhclibParserEx.extensionImplications
      In an equation for ‘extensionImplications’:
          extensionImplications = GhclibParserEx.extensionImplications
   |
64 | extensionImplications = GhclibParserEx.extensionImplications
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

shayne-fletcher:
hlint does not currently support building with ghc-9.2.1. work is underway to fix that.

garyamorris:
might be a silly question, new to Haskell. How do I install hlint if I have 9.2.1 installed? cabal install hlint throws errors for me.

LeventErkok:
@garyamorris You can't, until they make a new release. But as I understand it, it should be out any day now.

LeventErkok:
[...]

PR #1353, opened 2022-03-09
update lower bound on ghc-lib-parser-ex

shayne-fletcher:
use ghc-lib-parser-ex-9.2.0.3 as lower bound (fixes https://github.com/ndmitchell/hlint/issues/1314#issuecomment-1063382753)

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -89,7 +89,7 @@ library
       build-depends:
           ghc-lib-parser == 9.2.*
     build-depends:
-        ghc-lib-parser-ex >= 9.2.0.2 && < 9.2.1
+        ghc-lib-parser-ex >= 9.2.0.3 && < 9.2.1
 
     if flag(gpl)
         build-depends: hscolour >= 1.21


src/Hint/List.hs
```diff
@@ -63,7 +63,7 @@ import GHC.Builtin.Types
 import GHC.Util
 import Language.Haskell.GhclibParserEx.GHC.Hs.Pat
 import Language.Haskell.GhclibParserEx.GHC.Hs.Expr
-import Language.Haskell.GhclibParserEx.GHC.Hs.Types
+import Language.Haskell.GhclibParserEx.GHC.Hs.Type
```
ndmitchell, line 66: This change is just the inverse of #1352 - is that deliberate?

shayne-fletcher: change was deliberate. i can't tell you why #1352 was made. it's not right and this fixes it.

shayne-fletcher: ah wait. i guess i get it. i failed to pull in this change. let me rebase and push again.

shayne-fletcher: right, so it wasn't the inverse, it was applying the same change because i failed to branch from head. have rebased and pushed again. that will sort this

PR #1354, opened 2022-03-13
Replace "function" with "definition" where applicable in README

georgefst:
e.g. one can disable a hint for a type definition, or a constant, not only a function.


ndmitchell:
Thanks!

Issue #1355, opened 2022-03-13
Build plan broken without --allow-newer 


hseg:
b34a9497 should've also bumped the `ghc-lib-parser-ex` version in `stack.yaml`, with the bump everything works fine

PR #1356, opened 2022-03-13
Bump ghc-lib-parser-ex version

hseg:
Forgotten in b34a949
Closes: #1355

### DIFF
--- a/.ghci
+++ b/.ghci
@@ -1,3 +1,4 @@
+:set -W
 :set -Wincomplete-patterns
 :set -Wunused-binds -Wunused-imports -Worphans
 :set -isrc


hseg:
the build failures seem to be unrelated:
`neil: The .ghci file does not enough of -W -fwarn-unused-binds -fwarn-unused-imports -fwarn-orphans`

hseg:
Hrm. Unsure of why this had this effect, but perhaps 4f02b44 isn't properly triggering the ``"-W" `elem` src`` branch of `checkGhci` in https://github.com/ndmitchell/neil?

hseg:
Presuming #1357 works, have rebased this PR on top of it to prove it fixes the build

ndmitchell:
The only change in this diff seems to be to .ghci? Or did merging things underneath it make this unnecessary?

hseg:
Hrm, odd, should've checked again after rebasing. #1358 appears to have saved the original content, so it should be fine to just close this one.

PR #1357, opened 2022-03-13
Set -Wextra (fixes CI job)

hseg:
`Cabal.hs` in [@ndmitchell/neil](https://github.com/ndmitchell/neil) checks that either `-W` is set or a couple `-fwarn-*`
flags are. Simplest fix is to just set `-W` (synonym for `-Wextra`)


ndmitchell:
The idea of -W was for projects which ban all warnings, which isn't HLint. I've fixed up the neil script to accept -W now instead of -fwarn. Sorry for the breakage!

PR #1358, opened 2022-03-13
bump ghc-lib-parser-ex to 9.2.0.3

shayne-fletcher:
fixes https://github.com/ndmitchell/hlint/issues/1355 (modifications to stack.yaml seem to have been deleted from https://github.com/ndmitchell/hlint/pull/1356; it's a good catch - we definitely want this). fwiw, afaik,  stack.yaml doesn't affect CI and i'm the only user.


ndmitchell:
Thanks!

Issue #1360, opened 2022-03-19
`hlint` doesn't recognize `CPP` language extension when `-XHaskell2010` is passed


RyanGlScott:
If I run `hlint -XHaskell2010` on this module:

```hs
{-# LANGUAGE CPP #-}
{-# LANGUAGE MagicHash #-}
module Bug where

#if 1
import GHC.Exts

add# :: Int# -> Int# -> Int#
add# = (+#)
#endif
```

It will fail:

```
$ hlint -XHaskell2010 Bug.hs
Bug.hs:5:2: Error: Parse error: lexical error in pragma at character 'i'
Found:
    module Bug where
    
  > #if 1
    import GHC.Exts
  

1 hint
```

In order to work around this, I have to explicitly pass `-XCPP` to `hlint`:

```
$ hlint -XHaskell2010 -XCPP Bug.hs
No hints
```

This step feels unnecessary, however, because `Bug.hs` _already_ enables `CPP` at the top. Moreover, I didn't have to pass `-XMagicHash` to `hlint`, even though that is also required for the code to parse, so there appears to be a discrepancy in which flags one must explicitly pass to `hlint`.

ndmitchell:
Thanks for the report, the PR from @shayne-fletcher in #1361 fixes it (and I think we can iterate a bit from there to get the right solution).

PR #1361, opened 2022-03-20
Override NoCpp if neccessary

shayne-fletcher:
If `CPP` is not a default enabled extension (because e.g. `-XHaskell2010` was provided and so hlint itself doesn't decide what the default extensions in effect are) and if further, `-XCPP` is not provided on the command line, then, the logic of `CmdLine.cmdCpp`  writes `NoCpp` into the prevailing `ParseFlags`.

In this scenario when later we determine `CPP` is in effect for a module because we read so from its pragmas, then we `runCpp` on it in`parseModuleEx` but the `NoCpp` means doing so has no effect! So, here, don't call `runCpp` with `cppFlags flags` but rather with `Cpphs defaultCpphsOptions`. Perhaps this will do?

Fixes https://github.com/ndmitchell/hlint/issues/1360.

### DIFF
--- a/src/GHC/All.hs
+++ b/src/GHC/All.hs
@@ -39,8 +39,7 @@ import GHC.Util
 
 -- | What C pre processor should be used.
 data CppFlags
-    = NoCpp -- ^ No pre processing is done.
-    | CppSimple -- ^ Lines prefixed with @#@ are stripped.
+    = CppSimple -- ^ Lines prefixed with @#@ are stripped.
     | Cpphs CpphsOptions -- ^ The @cpphs@ library is used.
 
 -- | Created with 'defaultParseFlags', used by 'parseModuleEx'.
@@ -54,7 +53,7 @@ data ParseFlags = ParseFlags
 
 -- | Default value for 'ParseFlags'.
 defaultParseFlags :: ParseFlags
-defaultParseFlags = ParseFlags NoCpp Nothing defaultExtensions [] defaultFixities
+defaultParseFlags = ParseFlags CppSimple Nothing defaultExtensions [] defaultFixities
 
 -- | Given some fixities, add them to the existing fixities in 'ParseFlags'.
 parseFlagsAddFixities :: [FixityInfo] -> ParseFlags -> ParseFlags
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks for the diff, really useful, especially the comments, to understand what is going on. After reading around a bit, I don't think this is the right solution. For example, it doesn't set `#define __HLINT__ 1`, and if the user passes `--cpp-ansi` or similar those options won't get wired through. I think the right answer might actually be to delete `NoCpp` entirely, and have `cmdCpp` not look at whether `Cpp` is in the list of extensions. In that world you have `CppFlags` saying _how_ to run the preprocessor if you run it, and separately the extensions say _if_ you run it at all.

shayne-fletcher:
Makes sense. Have pushed an update! @ndmitchell  I think this is ready for a re-review.

ndmitchell:
Awesome- thanks for the fix. It would be good to have an explicit test for this since it's a little subtle, but I'm happy to follow up with that - your choice.

src/GHC/All.hs
```diff
@@ -54,7 +53,7 @@ data ParseFlags = ParseFlags
 
 -- | Default value for 'ParseFlags'.
 defaultParseFlags :: ParseFlags
-defaultParseFlags = ParseFlags NoCpp Nothing defaultExtensions [] defaultFixities
+defaultParseFlags = ParseFlags CppSimple Nothing defaultExtensions [] defaultFixities
```
shayne-fletcher, line 56: perhaps we might prefer something like
```haskell
Cpphs defaultCpphsOptions
        {boolopts=defaultBoolOptions{hashline=False, stripC89=True, ansi=False}
        ,includes = []
        ,preInclude = []
        ,defines = [("__HLINT__","1")]
        }
```

shayne-fletcher:
> Awesome- thanks for the fix. It would be good to have an explicit test for this since it's a little subtle, but I'm happy to follow up with that - your choice.

i'll work something out. consider it put on my stack.

src/GHC/All.hs
ndmitchell: I think we always override it before using it? If so, the simpler the value the better. But I wouldn't be averse to what you suggest either.

shayne-fletcher: that was my assumption as well (that we override before use) so, "if it ain't broke don't fix it". let's leave it be for now.

PR #1362, opened 2022-03-20
Add support for language specifier GHC2021

shayne-fletcher:
`CmdLine.getExtensions` now considers the data constructor `GHC2021` (in addition to `Haskell98` and `Haskell2010`).


ndmitchell:
Thanks!

PR #1363, opened 2022-03-21
Fmap hints also for infix version (<$>)

ysangkok:
Tested using

    module Main where
    import Data.Maybe
    main = print (isJust (even <$> x))
      where x = Just 5

which yields the following warning as expected:

    $ hlint -h ./data/hlint.yaml test.hs
    test.hs:4:15-33: Warning: Redundant <$>
    Found:
      isJust (even <$> x)
    Perhaps:
      isJust x

    1 hint

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -131,6 +131,7 @@
     # LIST
 
     - warn: {lhs: concat (map f x), rhs: concatMap f x}
+    - warn: {lhs: concat (f <$> x), rhs: concatMap f x}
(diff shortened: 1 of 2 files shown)


ysangkok:
@ndmitchell Seems like the tests failed because I hadn't regenerated the markdown rendition. So I have pushed a commit that does that.

ysangkok:
I must be generating the markdown file in the wrong way. What I am trying is `-h ./data/hlint.yaml --generate-summary` after having done `cabal install --overwrite-policy=always`. Is that not correct?

PR #1367, opened 2022-03-27
Add brackets on refactoring templates when needed

### LINKED ISSUE #1364: Missing parentheses in refactoring templates
The bug reported in https://github.com/mpickering/apply-refact/issues/124 is caused by missing parentheses in the refactoring template.

The refactoring template is `f Control.Monad.<=< g`, which should be `(f Control.Monad.<=< g)`.

zliu41:
Fixes #1364, fixes https://github.com/mpickering/apply-refact/issues/124

We call `addBracket` on the RHS to add brackets when needed. It appears the same should be done to `tpl` so that the refactoring output also includes the brackets.


ndmitchell:
Thanks! Makes sense we need to do it to the template and normal RHS

PR #1368, opened 2022-03-27
Move TH quotation bracket rule to new group

yairchu:
(see discussion in #1194)

**Note**: The test for this rule was removed due to not being enabled by default.
Is there a way to test disabled-by-default rules?


ndmitchell:
Unfortunately no way to test an off-by-default hint. Before we moved to .yaml files, we had a Test.hs which let us test out hints that weren't available at all, but I think that capability has fallen by the wayside since. Would be good to restore that if we could, then we this hint on as standard in a test setting, and test it. But until then, I think it should be fine.

PR #1372, opened 2022-04-15
Directly suggest `forM_ x y` given `void (forM x y)`

zliu41:
Currently the suggestion for `void (forM x y)` is `void (forM_ x y)`. This is kind of weird if not confusing - if we are keeping the `void`, then why bother changing `forM` to `forM_`. This PR makes HLint directly suggest `(forM_ x y)`.

### DIFF
--- a/src/Hint/Monad.hs
+++ b/src/Hint/Monad.hs
@@ -1,4 +1,10 @@
-{-# LANGUAGE LambdaCase, ViewPatterns, PatternGuards, FlexibleContexts #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE ViewPatterns #-}
+
 {-
     Find and match:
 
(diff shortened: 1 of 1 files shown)


src/Hint/Monad.hs
```diff
@@ -45,7 +51,9 @@ main = f $ do g a $ sleep 10 --
 main = do f a $ sleep 10 -- @Ignore
 main = do foo x; return 3; bar z --
 main = void $ forM_ f xs -- forM_ f xs
-main = void $ forM f xs -- void $ forM_ f xs
+main = void (forM_ f xs) -- (forM_ f xs)
```
ndmitchell, line 54: These brackets seem a bit odd?

ndmitchell (commented):
Are the brackets a necessity, or could we drop them?

src/Hint/Monad.hs
zliu41: Probably, but this is orthogonal and unrelated to this change. `void (forM_ f xs)` -> `(forM_ f xs)` is the same as the previous behavior. The new behavior is `void (forM f xs)` -> `(forM_ f xs)`.

zliu41: @ndmitchell I hoped this addresses your question - let me know if you have any other concerns!

ndmitchell: Ah, didn't realise that was what it did before. Have fixed that separately in accc63710b17bf9b533288b3c0ba06ecd5c3be06

Issue #1374, opened 2022-04-18
Feature request: Add `--generate-ignore` flag


philderbeast:
When a large project has not been linted before it can be quite a bit of work to tackle and I'd prefer to do this a suggestion at a time. Towards that end, it would be helpful if we could generate an `.hlint.yaml` file for a clean pass with comments showing the number of occurrences. Something like this (from [liquid-fixpoint](https://github.com/ucsd-progsys/liquid-fixpoint)):

```yaml
# .hlint.yaml
# Lint hints found with hlint-3.3.6
# All supressed with the following ignores.
- ignore: {name: Redundant bracket} # 198 hints
- ignore: {name: Redundant if} # 1 hint
- ignore: {name: Redundant guard} # 2 hints
- ignore: {name: Redundant lambda} # 3 hints
- ignore: {name: Redundant $} # 20 hints
- ignore: {name: Redundant <$>} # 29 hints
- ignore: {name: Redundant bang pattern} # 33 hints
- ignore: {name: Redundant flip} # 1 hint
- ignore: {name: Redundant return} # 1 hint
- ignore: {name: Use camelCase} # 2 hints
- ignore: {name: Use newtype instead of data} # 3 hints
- ignore: {name: Use section} # 3 hints 
- ignore: {name: Use gets} # 9 hints
- ignore: {name: Use guards} # 3 hints
- ignore: {name: Use maybe} # 7 hints
- ignore: {name: Use fold} # 1 hint
- ignore: {name: Use map} # 2 hints
- ignore: {name: Use map once} # 1 hint
- ignore: {name: Use mapMaybe} # 3 hints
- ignore: {name: Use concatMap} # 2 hints
- ignore: {name: Use infix} # 3 hints
- ignore: {name: Use elem} # 1 hint
- ignore: {name: Use notElem} # 2 hints
- ignore: {name: Use <$>} # 10 hints
- ignore: {name: Use <$} # 2 hints
- ignore: {name: Use $>} # 7 hints
- ignore: {name: Use if} # 1 hint
- ignore: {name: Use /=} # 1 hint
- ignore: {name: Use ++} # 1 hint
- ignore: {name: Use const} # 1 hint
- ignore: {name: Use void} # 2 hints
- ignore: {name: Use first} # 1 hint
- ignore: {name: Use second} # 1 hint
- ignore: {name: Use all} # 2 hints
- ignore: {name: Use tuple-section} # 1 hint
[...]

PR #1375, opened 2022-04-19
Add a command to suppress outstanding hints.

philderbeast:
Thanks for the pull request!

> All good, I've read them.

For #1374. This adds a `--willful-ignorance` command that is the subset of `--default` but also shows the number of occurrences of each hint suppressed. Example output:

```
liquid-fixpoint on develop > hlint --willful-ignorance src/ tests/ win/ unix/
Warning: unknown directive ##
in src/Language/Fixpoint/SortCheck.hs  at line 779 col 1
Warning: unknown directive ##
in src/Language/Fixpoint/SortCheck.hs  at line 792 col 1
Warning: unknown directive ##
in src/Language/Fixpoint/SortCheck.hs  at line 830 col 1
# Warnings currently triggered by your code
- ignore: {name: "Avoid lambda"} # 4 hints
- ignore: {name: "Eta reduce"} # 30 hints
- ignore: {name: "Evaluate"} # 1 hint
- ignore: {name: "Fix pragma markup"} # 38 hints
- ignore: {name: "Functor law"} # 3 hints
- ignore: {name: "Move brackets to avoid $"} # 8 hints
- ignore: {name: "Redundant $"} # 20 hints
- ignore: {name: "Redundant <$>"} # 29 hints
- ignore: {name: "Redundant bang pattern"} # 33 hints
- ignore: {name: "Redundant bracket"} # 198 hints
- ignore: {name: "Redundant flip"} # 1 hint
- ignore: {name: "Redundant guard"} # 2 hints
- ignore: {name: "Redundant if"} # 1 hint
- ignore: {name: "Redundant lambda"} # 3 hints
- ignore: {name: "Redundant return"} # 1 hint
- ignore: {name: "Unused LANGUAGE pragma"} # 29 hints
- ignore: {name: "Use $>"} # 7 hints
- ignore: {name: "Use ++"} # 1 hint
- ignore: {name: "Use /="} # 1 hint
- ignore: {name: "Use <$"} # 2 hints
- ignore: {name: "Use <$>"} # 10 hints
- ignore: {name: "Use all"} # 2 hints
- ignore: {name: "Use camelCase"} # 2 hints
- ignore: {name: "Use concatMap"} # 3 hints
- ignore: {name: "Use const"} # 1 hint
- ignore: {name: "Use elem"} # 1 hint
- ignore: {name: "Use first"} # 1 hint
- ignore: {name: "Use fold"} # 1 hint
- ignore: {name: "Use gets"} # 9 hints
- ignore: {name: "Use guards"} # 3 hints
- ignore: {name: "Use if"} # 1 hint
- ignore: {name: "Use infix"} # 3 hints
- ignore: {name: "Use map"} # 2 hints
- ignore: {name: "Use map once"} # 1 hint
- ignore: {name: "Use mapM_"} # 1 hint
- ignore: {name: "Use mapMaybe"} # 4 hints
- ignore: {name: "Use maybe"} # 7 hints
- ignore: {name: "Use newtype instead of data"} # 3 hints
- ignore: {name: "Use notElem"} # 2 hints
- ignore: {name: "Use record patterns"} # 5 hints
- ignore: {name: "Use second"} # 1 hint
- ignore: {name: "Use section"} # 3 hints
- ignore: {name: "Use tuple-section"} # 1 hint
- ignore: {name: "Use void"} # 2 hints
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,18 +241,20 @@ See discussion in [issue #372](https://github.com/ndmitchell/hlint/issues/372).
 
 ## Customizing the hints
 
-To customize the hints given by HLint, create a file `.hlint.yaml` in the root of your project. For a suitable default run:
+To customize the hints given by HLint, create a file `.hlint.yaml` in the root of your project. For a suitable configuration use either `hlint --suppress > .hlint.yaml` or `hlint --default > .hlint.yaml`.
 
-```console
-hlint --default > .hlint.yaml
-```
+* `--suppress` prints settings for ignoring all the hints currently outstanding.
+  ```yaml
+  # Warnings currently triggered by your code
+  - ignore: {name: "Redundant $"} # 20 hints
+  - ignore: {name: "Unused LANGUAGE pragma"} # 29 hints
+  ```
+* `--default` implies `--suppress` and shows examples (as comments) of how to:
 
-This default configuration contains lots of examples, including:
-
-* Adding command line arguments to all runs, e.g. `--color` or `-XNoMagicHash`.
-* Ignoring certain hints, perhaps within certain modules/functions.
-* Restricting use of GHC flags/extensions/functions, e.g. banning `Arrows` and `unsafePerformIO`.
-* Adding additional project-specific hints.
+  * Add command line arguments to all runs, e.g. `--color` or `-XNoMagicHash`.
+  * Ignore certain hints, perhaps within certain modules/functions.
+  * Restrict the use of GHC flags/extensions/functions, e.g. banning `Arrows` and `unsafePerformIO`.
+  * Add additional project-specific hints.
 
 You can see the output of `--default` [here](https://github.com/ndmitchell/hlint/blob/master/data/default.yaml).
 
--- a/src/CmdLine.hs
+++ b/src/CmdLine.hs
@@ -108,6 +108,7 @@ data Cmd
         ,cmdFindHints :: [FilePath]      -- ^ source files to look for hints in
         ,cmdDataDir :: FilePath          -- ^ the data directory
         ,cmdDefault :: Bool              -- ^ Print a default file to stdout
+        ,cmdIgnoreFound :: Bool          -- ^ Suppress outstanding hints
         ,cmdPath :: [String]
         ,cmdCppDefine :: [String]
         ,cmdCppInclude :: [FilePath]
@@ -147,6 +148,7 @@ mode = cmdArgsMode $ modes
         ,cmdFindHints = nam "find" &= typFile &= help "Find hints in a Haskell file"
         ,cmdDataDir = nam_ "datadir" &= typDir &= help "Override the data directory"
         ,cmdDefault = nam "default" &= help "Print a default file to stdout"
+        ,cmdIgnoreFound = nam_ "willful-ignorance" &= help "Suppress outstanding hints"
         ,cmdPath = nam "path" &= help "Directory in which to search for files"
         ,cmdCppDefine = nam_ "cpp-define" &= typ "NAME[=VALUE]" &= help "CPP #define"
         ,cmdCppInclude = nam_ "cpp-include" &= typDir &= help "CPP include path"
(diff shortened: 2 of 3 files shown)


philderbeast:
The help:

```
> hlint --help
HLint v3.3.5, (C) Neil Mitchell 2006-2022

hlint [OPTIONS] [FILE/DIR]

Common flags:

  -d --default                             Print a default file to stdout
     --willful-ignorance                   Suppress outstanding hints
```

ndmitchell:
Thanks for the request! So the difference between this and `--default` is that it shows the number of hints that were ignored? Any reason that we couldn't do that for `--default` and stick with the single flag? Also, if we're showing the numbers, it seems to make sense to put the most/least at the top and sort them?

ndmitchell:
As to other options to make this more discoverable, we could try:

1. Adding another command line option as an alias for it.
2. Improving the description.
3. Suggesting that if the user doesn't have a .yaml file, and they have warnings, then print a warning they might want to generate one (and mentioning `--default`)

philderbeast:
> So the difference between this and `--default` is that it shows the number of hints that were ignored? Any reason that we couldn't do that for `--default` and stick with the single flag?

I had considered that but `--default` generates way more than just the list of ignores with about 60 lines of comments following the ignores. These are really useful as documentation for first time use of hlint but something I'd be editing away if only interested in maintaining an `.hlint.yaml` file and whittling away the number of ignores by following the hints in the code being linted.

philderbeast:
> Also, if we're showing the numbers, it seems to make sense to put the most/least at the top and sort them?

I like how the list comes sorted alphabetically. I find it easy to scan the list to see the big numbers.

philderbeast:
@ndmitchell I've had another crack at this and added updated the readme. From the options you suggest, I did 1 & 2 but not 3.

philderbeast:
Here's the rendered readme:

## Customizing the hints

To customize the hints given by HLint, create a file `.hlint.yaml` in the root of your project. For a suitable configuration use either `hlint --suppress > .hlint.yaml` or `hlint --default > .hlint.yaml`.

* `--suppress` prints settings for ignoring all the hints currently outstanding.
  ```yaml
  # Warnings currently triggered by your code
  - ignore: {name: "Redundant $"} # 20 hints
  - ignore: {name: "Unused LANGUAGE pragma"} # 29 hints
  ```
* `--default` implies `--suppress` and shows examples (as comments) of how to:

  * Add command line arguments to all runs, e.g. `--color` or `-XNoMagicHash`.
  * Ignore certain hints, perhaps within certain modules/functions.
  * Restrict the use of GHC flags/extensions/functions, e.g. banning `Arrows` and `unsafePerformIO`.
  * Add additional project-specific hints.

You can see the output of `--default` [here](https://github.com/ndmitchell/hlint/blob/master/data/default.yaml).

If you wish to use the [Dhall configuration language](https://github.com/dhall-lang/dhall-lang) to customize HLint, there [is an example](https://kowainik.github.io/posts/2018-09-09-dhall-to-hlint) and [type definition](https://github.com/kowainik/relude/blob/master/hlint/Rule.dhall).

philderbeast:
Here's the rendered help:

```
> hlint --help
HLint v3.3.5, (C) Neil Mitchell 2006-2022

hlint [OPTIONS] [FILE/DIR]

Common flags:
...
  -d --default                             Print default to stdout
     --suppress                            Suppress ideas with ignore
```

ndmitchell:
Are the extra notes added to `--default` harmful? It still seems like compressing these into one piece of functionality, but retaining both flag aliases, might be easier in the long run? I think the idea of generating a lot of documentation that usually gets edited away by experienced users is quite standard, e.g. tools like `cargo new` do it as just one example I ran into in the last day or two.

philderbeast:
This could be a good time to pause and compare to other tools. A common enough set of commands are `init` followed by `update` or `outdated` but then tools like this write directly to their config file.

One thing that surprises me is that `hlint --default > .hlint.yaml` is not the same as having no `.hlint.yaml` configuration at all and so for the unwary hides hints.

philderbeast:
I'm closing this PR in favour of #1380.

Issue #1376, opened 2022-04-29
Consider defaulting to ghc-lib


pepeiborra:
The Cabal flag `ghc-lib` currently defaults to False. Consider changing the default value to True so that projects like HLS that depend on `hlint` can have a working `cabal install ...` command across all GHC versions.

ndmitchell:
There were two reasons for doing it that way:

* For compatible GHC versions, it allows HLS to pass the AST from GHC along without modification, which means less memory, less parsing time, and great compatibility. Is that no longer desired? Or do you have ways to get those advantages without using the same one as the GHC API?
* It makes it slower to compile for users - quite significantly slower. But if it's easier for HLS users, that's probably a worthwhile tradeoff, given there are provided binaries.

pepeiborra:
> There were two reasons for doing it that way:
> 
> 
> 
> * For compatible GHC versions, it allows HLS to pass the AST from GHC along without modification, which means less memory, less parsing time, and great compatibility. Is that no longer desired? 

Less memory and parsing time, yes, but why do you say greater compatibility? 

> Or do you have ways to get those advantages without using the same one as the GHC API?

No, but we are willing to give them up in exchange for greater compatibility. Manual Cabal flags have a poor user interface, specially in combination with Hackage. 

> * It makes it slower to compile for users - quite significantly slower. But if it's easier for HLS users, that's probably a worthwhile tradeoff, given there are provided binaries.

Ah, this is a good point. 

Summarising, HLS really wants to depend on the ghc-lib version of hlint, because that one works on all ghc versions and HLS already depends on ghc-lib anyway. Or as we do now, selectively use the ghc-lib hlint on non-native ghc versions, but in a declarative way. 

I can see how defaulting the flag to true would be a net loss for standalone hlint use cases. 

It occurs to me that the best solution would be a separate package hlint-ghclib.

ndmitchell:
> but why do you say greater compatibility?

You get a AST out of GHC, that must have had all the right CPP, extensions etc set, because you are using it for compilation. The worry is that when HLS tries to create a second GHC AST, it might get some things wrong. This is a worry, but one careful coding can get around.

> No, but we are willing to give them up in exchange for greater compatibility. 

Cool, that sounds a reasonable trade-off to make.

> Manual Cabal flags have a poor user interface, specially in combination with Hackage.

Agreed!

[...]

PR #1378, opened 2022-05-04
Replace the ghc-lib flag with a subcomponent

pepeiborra:
Addresses #1376 by creating a subcomponent `hlint:ghclib` that replaces the `ghc-lib` flag. This enables other packages upstream to declaratively select which "version" of `hlint` they need, e.g. HLS: https://github.com/pepeiborra/ide/commit/637aa0ba169b40dbcd73b54afb3e2834e1608b4e

A priori, there is no downside to introducing the `ghclib` component:

1. Building the `hlint` binary doesn't pull the `ghclib` component
2. No additional Hackage upload required

But there might be some hidden costs. Most tooling struggles with Cabal subcomponents. I don't know if `stack` can handle them, for instance.

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -1,8 +1,8 @@
-cabal-version:      1.18
+cabal-version:      3.0
 build-type:         Simple
 name:               hlint
 version:            3.4
-license:            BSD3
+license:            BSD-3-Clause
 license-file:       LICENSE
 category:           Development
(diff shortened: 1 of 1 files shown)


shayne-fletcher:
@pepeiborra https://github.com/ndmitchell/hlint/pull/1410 has been accepted that flips the `ghc-lib` flag default to `True`. given past discussions i think then we might choose to abandon this PR now and close it out?

PR #1380, opened 2022-05-16
Add counts as comments for `--default`.

philderbeast:
Alternative pull request for #1374 without adding an extra command as an alternative to #1375.


ndmitchell:
Thanks, this looks great :)

PR #1385, opened 2022-06-05
Flip ghc-lib default to True

shayne-fletcher:
Summary:

- Flip cabal flag `ghc-lib` default to `True` (i.e. disable 'auto' mode; unconditionally link `ghc-lib-parser`)
- Upgrade to `ghc-lib-parser-ex-9.2.1.0` (which has the same link behavior with respect to `ghc-lib-parser`)
- Fixes https://github.com/ndmitchell/hlint/issues/1376

Test plan:

(1) Execute `stack build --resolver nightly-2022-05-27 --cabal-verbose` (i.e. ghc-9.2.2). Observe,
```
depends:
  ...
  ghc-lib-parser-9.2.3.20220527-Apb5ZlADTS72sl2ZpNOJTy
  ghc-lib-parser-ex-9.2.1.0-13IoJhA5FNUCTxsgEW3SsR
  ...
```
This shows that `ghc-lib-parser` is linked and not ghc libs despite the compiler version being >=9.2.2 && <9.3.

(2) Enable in `stack.yaml` the stanza 
```yaml
flags:
   hlint:
     ghc-lib: false
   ghc-lib-parser-ex:
     auto: true
     no-ghc-lib: false
```
and execute`stack build --resolver nightly-2022-05-27 --cabal-verbose`. Observe now,
```
depends:
  ...
  ghc-9.2.2
  ghc-boot-9.2.2
  ghc-boot-th-9.2.2
  ...
```
This shows that with the default flags overridden we can recover the "old" behavior of linking the the native ghc libs rather than `ghc-lib-parser` when the compiler version is >= 9.2.2 && < 9.3

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -5,19 +5,20 @@ packages:
 
 extra-deps:
   - ghc-lib-parser-9.2.3.20220709
-  - ghc-lib-parser-ex-9.2.0.4
+  - ghc-lib-parser-ex-9.2.1.0
 # To test hlint against experimental builds of ghc-lib-parser-ex,
 # modify extra-deps like this:
 #  - archive: /users/shayne/project/ghc-lib-parser-ex/ghc-lib-parser-ex-8.10.0.18.tar.gz
 #    sha256: "0000000000000000000000000000000000000000000000000000000000000000"
 
-ghc-options: {"$locals": -ddump-to-file -ddump-hi -Werror=unused-imports -Werror=unused-local-binds -Werror=unused-top-binds -Werror=orphans}
-
-# Enabling this stanza forces both hlint and ghc-lib-parser-ex to
-# depend on ghc-lib-parser.
+# Enabling this stanza allows hlint and ghc-lib-parser-ex to link to
+# the native ghc compiler libs rather than ghc-lib-parser when the
+# compiler version is suitable.
 # flags:
-#   hlint:
-#     ghc-lib: true
-#   ghc-lib-parser-ex:
-#     auto: false
-#     no-ghc-lib: false
+#    hlint:
+#     ghc-lib: false
+#    ghc-lib-parser-ex:
+#      auto: true
+#      no-ghc-lib: false
+
+ghc-options: {"$locals": -ddump-to-file -ddump-hi -Werror=unused-imports -Werror=unused-local-binds -Werror=unused-top-binds -Werror=orphans}
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -53,7 +53,7 @@ flag gpl
     description: Use GPL libraries, specifically hscolour
 
 flag ghc-lib
-  default: False
+  default: True
   manual: True
   description: Force dependency on ghc-lib-parser even if GHC API in the ghc package is supported
 
@@ -90,7 +90,7 @@ library
       build-depends:
           ghc-lib-parser == 9.2.*
     build-depends:
-        ghc-lib-parser-ex >= 9.2.0.3 && < 9.2.1
+        ghc-lib-parser-ex >= 9.2.1.0 && < 9.2.2
 
     if flag(gpl)
         build-depends: hscolour >= 1.21


shayne-fletcher:
it looks like macos CI is struggling with errors of this kind?
```
/Users/runner/.ghcup/ghc/9.2.2/lib/ghc-9.2.2/lib/../lib/x86_64-osx-ghc-9.2.2/rts-1.0.2/include/ffitarget.h:10:10: fatal error: 'ffitarget_x86.h' file not found
#include "ffitarget_x86.h"
         ^~~~~~~~~~~~~~~~~
1 error generated.
```
this is a known GHC issue. a workaround while it's not fixed upstream is to execute build commands in an environment in which `C_INCLUDE_PATH=`xcrun --show-sdk-path`/usr/include/ffi`. hope this helps!

stack.yaml
```diff
@@ -4,20 +4,21 @@ packages:
   - .
 
 extra-deps:
-  - ghc-lib-parser-9.2.3.20220507
-  - ghc-lib-parser-ex-9.2.0.4
+  - ghc-lib-parser-9.2.3.20220527
```
shayne-fletcher, line 7: changing this version is an incidental bugfix

stack.yaml
```diff
@@ -4,20 +4,21 @@ packages:
...
+  - ghc-lib-parser-ex-9.2.1.0
```
shayne-fletcher, line 8: changing this though is not, it goes together with the change in default of `ghc-lib`.

shayne-fletcher:
> it looks like macos CI is struggling with errors of this kind?
> 
> ```
> /Users/runner/.ghcup/ghc/9.2.2/lib/ghc-9.2.2/lib/../lib/x86_64-osx-ghc-9.2.2/rts-1.0.2/include/ffitarget.h:10:10: fatal error: 'ffitarget_x86.h' file not found
> #include "ffitarget_x86.h"
>          ^~~~~~~~~~~~~~~~~
> 1 error generated.
> ```
> 
> this is a known GHC issue (see more about this [here](https://github.com/digital-asset/ghc-lib/issues/352)). tl;dr a workaround while it's not fixed upstream is to execute build commands in an environment in which `C_INCLUDE_PATH=$(xcrun --show-sdk-path)/usr/include/ffi`. hope this helps!

this is resolved with ghc-9.2.3 however (https://gitlab.haskell.org/ghc/ghc/-/issues/20592#note_435963) so maybe the best resolution is to ghcup that one.

ndmitchell:
Stack has moved to GHC 9.2.3 for its nightly. Shayne, can you push a rebase so I can get fresh signal on the CI. If github via ghcup has moved to 9.2.3 I think we're good to land this, otherwise I think we might have to wait.

shayne-fletcher:
that's a shame. the ghc version continues to be ghc-9.2.2.

shayne-fletcher:
> Stack has moved to GHC 9.2.3 for its nightly. Shayne, can you push a rebase so I can get fresh signal on the CI. If github via ghcup has moved to 9.2.3 I think we're good to land this, otherwise I think we might have to wait.

sadly we are looking stuck. i guess there's no provision afforded us by github where we can arrange our builds be executed in an environment that sets `C_INCLUDE_PATH`?

ndmitchell:
I have no doubt we can fix the CI builds. What I worry is about having CI as a signal for end users, where HLint might no longer compile.

shayne-fletcher:
i've got a build running in https://github.com/ndmitchell/hlint/pull/1396 that 
```
         - os: macOS-latest
-          ghc: '9.2'
+          ghc: '9.2.3'
```
so far it's looking good. if it goes through would you consider this a fix?

shayne-fletcher:
> I have no doubt we can fix the CI builds. What I worry is about having CI as a signal for end users, where HLint might no longer compile.

i think i'm hearing your concern as being users encountering this error in the wild when building hlint? if so, yes, the only "fix" i know we have for that are workarounds that set `C_INCLUDE_PATH` (or equivalent) (there are plenty of these in https://gitlab.haskell.org/ghc/ghc/-/issues/20592). 

i wonder if this isn't already a problem for users on the current version of hlint (e.g. this report https://github.com/digital-asset/ghc-lib/issues/352). i simply don't know.

shayne-fletcher:
> i wonder if this isn't already a problem for users on the current version of hlint (e.g. this report [digital-asset/ghc-lib#352](https://github.com/digital-asset/ghc-lib/issues/352)). i simply don't know.

argh, of course it is not. mac-OS/ghc-9.2.2 with the current version of hlint will build depend on `ghc` not `ghc-lib-parser`. that's why it's manifesting in this diff, and hasn't before now. ok, i guess we are on the same page now.

shayne-fletcher:
> i've got a build running in #1396 that
> 
> ```
>          - os: macOS-latest
> -          ghc: '9.2'
> +          ghc: '9.2.3'
> ```
> 
> so far it's looking good. if it goes through would you consider this a fix?

fwiw, macOS/ghc-9.2.3 builds with the changes of this diff have succeeded. making this modification to `ci.yml` in this PR would fix CI. do please let me know if you want me to amend this PR with it.

shayne-fletcher:
given https://github.com/ndmitchell/hlint/pull/1410 has been accepted which contains this change, i am abandoning this PR.

PR #1387, opened 2022-06-17
Bypass camelCase hint for new tasty_... custom test prefix

brprice:
In tasty-discover 4.2.4 support was added for "custom test cases" with a
prefix of 'tasty_'. We bypass the camelCase naming hint for these,
matching the current behaviour for the other tasty-discover prefixes.

(See https://github.com/haskell-works/tasty-discover#write-tests for the list of supported identifiers.)

### DIFF
--- a/src/Hint/Naming.hs
+++ b/src/Hint/Naming.hs
@@ -122,7 +122,7 @@ isSym _ = False
 suggestName :: String -> Maybe String
 suggestName original
     | isSym original || good || not (any isLower original) || any isDigit original ||
-        any (`isPrefixOf` original) ["prop_","case_","unit_","test_","spec_","scprop_","hprop_"] = Nothing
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks!

brprice:
Would it be possible to cut a new release with this change? It is not urgent, but would let us bump our dependencies on `tasty-*`.

brprice:
For anyone who stumbles across this: note that it is possible to configure the released version of hlint to ignore all camelcasing warnings (within particular modules if desired). This can be done by adding the following to your `.hlint.yaml`
```
- ignore:
   name: "Use camelCase"
   within: Tests.**
```

ndmitchell:
I hope to release tomorrow - sorry for the delay!

ndmitchell:
I released 3.4.1 which has this change

Issue #1391, opened 2022-06-27
A hint about QuickCheck?


jvoigtlaender:
I had a student write code like this:
```haskell
do e <- elements [gen1, gen2, gen3]
   e
```
I was quite impressed HLint proposed to rewrite this into:
```haskell
do join (elements [gen1, gen2, gen3])
```
But now I would like it to go one step further, to:
```haskell
do oneOf [gen1, gen2, gen3]
```

I do think that `join . elements = oneOf` holds. But it seems it would be the first hint about functions from the QuickCheck library at all? Is it reasonable to bring QuickCheck in as another "dependency"?

ndmitchell:
Fine to bring quick check in as another dependency - see https://github.com/ndmitchell/hlint/blob/e41ba201696154254608738ca4097c4bbe82892c/data/hlint.yaml#L907 for how to do it with autoparsec.

PR #1393, opened 2022-07-06
in QuickCheck: join . elements = oneOf

jvoigtlaender:
See https://github.com/ndmitchell/hlint/issues/1391.


ndmitchell:
Thanks!

PR #1394, opened 2022-07-06
replace maximum/minimum on two-element lists by basic max/min

jvoigtlaender:
Encountered in student code.


ndmitchell:
Thanks - great spot - this should probably have been a hint for a while!

PR #1397, opened 2022-07-11
evaluation rules for minimum/maximum on singleton lists

jvoigtlaender:
Probably nobody would write these left-hand sides directly, but they could arise as a consequence of step-wise changes or refactorings.


ndmitchell:
Thanks!

PR #1403, opened 2022-07-26
last . reverse -> head

jvoigtlaender:
Similar to https://github.com/ndmitchell/hlint/blob/eccfc5146a53e75cbda5aec21062501359b352f3/data/hlint.yaml#L151 and https://github.com/ndmitchell/hlint/blob/eccfc5146a53e75cbda5aec21062501359b352f3/data/hlint.yaml#L156 (and occurring in student code for similar reasons).


ndmitchell:
Thanks!

PR #1404, opened 2022-08-01
dual versions of existing min/max hints (from if-comparison-else)

jvoigtlaender:
These would be superfluous if hlint somehow "saw through comparisons", matching
```haskell
if a >= b then b else a
```
against
```haskell
if b <= a then b else a
```
But I guess it doesn't.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -120,9 +120,13 @@
     - hint: {lhs: flip (g `on` h), rhs: flip g `on` h, name: Move flip}
     - hint: {lhs: (f `on` g) `on` h, rhs: f `on` (g . h), name: Fuse on/on}
     - warn: {lhs: if a >= b then a else b, rhs: max a b}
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Similar "dualizations" might apply to https://github.com/ndmitchell/hlint/blob/c9bfda7de8c2872d0d207212b123345756084e98/data/hlint.yaml#L360-L363 but maybe this is overkill (making four out of each individual of these lines).

What I do guess people something write is these versions:
```haskell
'a' <= a && a <= 'z'
'A' <= a && a <= 'Z'
'0' <= a && a <= '9'
'0' <= a && a <= '7'
```

ndmitchell:
Thanks - I think adding a few dualisations is fine, but adding them all just because they logically fit is probably too much. Generally I only do the ones that seem plausible - e.g. `'a' <= a` is plausible, but putting the `z` variant first or doing the `z` literal on the RHS of a comparison both seem unlikely.

PR #1406, opened 2022-08-01
make lefts/rights hints apply more often

jvoigtlaender:
It seems `hlint` isn't realizing that the `a`s in the current hints have different scopes. They seem to be matched literally, so that the hints do match less often than they could.


ndmitchell:
Good catch! This seems like just a bug, we should have always written it as you have done now

PR #1408, opened 2022-08-01
evaluate trivial calls of map

jvoigtlaender:
Shares some motivation with https://github.com/ndmitchell/hlint/pull/1407.


ndmitchell:
Thanks!

PR #1409, opened 2022-08-02
notNull hints

jvoigtlaender:
Variants of: https://github.com/ndmitchell/hlint/blob/c9bfda7de8c2872d0d207212b123345756084e98/data/hlint.yaml#L210-L211 https://github.com/ndmitchell/hlint/blob/c9bfda7de8c2872d0d207212b123345756084e98/data/hlint.yaml#L811-L812

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -1072,6 +1072,10 @@
     - warn: {lhs: "last $ x : y", rhs: "lastDef x y"}
     - warn: {lhs: "drop 1", rhs: "drop1"}
     - warn: {lhs: "dropEnd 1", rhs: "dropEnd1"}
+    - warn: {lhs: notNull (concat x), rhs: any notNull x}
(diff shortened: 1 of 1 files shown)


ndmitchell:
Sounds like it would be a good idea to add `not (null ...)` versions too, as only people who use HLint will see the notNull hints fire

jvoigtlaender:
I don't think explicit `not (null ...)` variants are needed:
```haskell
    - warn: {lhs: not (null (concat x)), rhs: any notNull x}
    - warn: {lhs: not (null (filter f x)), rhs: any f x}
    - warn: {lhs: "not (null [x])", rhs: "True", name: Evaluate}
    - warn: {lhs: "not (null [])", rhs: "False", name: Evaluate}
```
because such expressions would already be addressed by other, existing rules (e.g., an `Evaluate` rule for `null [] -> True` plus one for `not True -> False`).

ndmitchell:
Ah, very true!

PR #1410, opened 2022-08-08
upgrade to ghc-9.4 parse tree

shayne-fletcher:
- this pr migrates hlint to the ghc-9.4.1 parse tree
- the minimum ghc version for building hlint with this pr becomes ghc-9.0.1
  - accordingly i've removed 8.10 from `.github/workflws/ci.yml`
  - of course with 9.4.1 being now available, one hopes we can add `9.4` to that matrix in the near future
- regarding issue https://github.com/ndmitchell/hlint/pull/1385:
  - ghc-lib-parser-ex switched to always linking `ghc-lib-parser` by default in pr https://github.com/shayne-fletcher/ghc-lib-parser-ex/pull/123
  - the companion change to hlint in pr https://github.com/ndmitchell/hlint/pull/1385 has ~~yet to land~~ been abandoned in light of this PR being accepted
  - since the choice of defaults must be consistent between these two packages, replicate https://github.com/ndmitchell/hlint/pull/1385 in this pr accordingly

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -53,7 +53,7 @@ flag gpl
     description: Use GPL libraries, specifically hscolour
 
 flag ghc-lib
-  default: False
+  default: True
   manual: True
   description: Force dependency on ghc-lib-parser even if GHC API in the ghc package is supported
 
@@ -81,16 +81,16 @@ library
         deriving-aeson >= 0.2,
         filepattern >= 0.1.1
 
-    if !flag(ghc-lib) && impl(ghc >= 9.2.2) && impl(ghc < 9.3.0)
+    if !flag(ghc-lib) && impl(ghc >= 9.4.1) && impl(ghc < 9.5.0)
       build-depends:
-        ghc == 9.2.*,
+        ghc == 9.4.*,
         ghc-boot-th,
         ghc-boot
     else
       build-depends:
-          ghc-lib-parser == 9.2.*
+          ghc-lib-parser == 9.4.*
     build-depends:
-        ghc-lib-parser-ex >= 9.2.0.3 && < 9.2.1
+        ghc-lib-parser-ex >= 9.4.0.0 && < 9.4.1
 
     if flag(gpl)
         build-depends: hscolour >= 1.21
--- a/src/Hint/Extensions.hs
+++ b/src/Hint/Extensions.hs
@@ -268,6 +268,8 @@ import GHC.Hs
 import GHC.Types.Basic
 import GHC.Types.Name.Reader
 import GHC.Types.ForeignCall
+import qualified GHC.Data.Strict
+import GHC.Types.PkgQual
 
 import GHC.Util
 import GHC.LanguageExtensions.Type
@@ -286,7 +288,7 @@ extensionsHint :: ModuHint
 extensionsHint _ x =
     [
         rawIdea Hint.Type.Warning "Unused LANGUAGE pragma"
-        (RealSrcSpan (anchor sl) Nothing)
+        (RealSrcSpan (anchor sl) GHC.Data.Strict.Nothing)
         (comment_ (mkLanguagePragmas sl exts))
         (Just newPragma)
         ( [RequiresExtension (show gone) | (_, Just x) <- before \\ after, gone <- Map.findWithDefault [] x disappear] ++
@@ -372,21 +374,9 @@ usedExt DeriveLift = hasDerive ["Lift"]
 usedExt DeriveAnyClass = not . null . derivesAnyclass . derives
 usedExt x = used x
 
--- The ghc-lib-parser-ex functions are getting fixed to have the new
--- signatures.
-isMDo' :: HsStmtContext GhcRn -> Bool
-isMDo' = \case MDoExpr _ -> True; _ -> False
-isStrictMatch' :: HsMatchContext GhcPs -> Bool
-isStrictMatch' = \case FunRhs{mc_strictness=SrcStrict} -> True; _ -> False
-
-isGetField :: LHsExpr GhcPs -> Bool
-isGetField = \case (L _ HsGetField{}) -> True; _ -> False
-isProjection :: LHsExpr GhcPs -> Bool
-isProjection = \case (L _ HsProjection{}) -> True; _ -> False
-
 used :: Extension -> Located HsModule -> Bool
 
-used RecursiveDo = hasS isMDo' ||^ hasS isRecStmt
+used RecursiveDo = hasS isMDo ||^ hasS isRecStmt
 used ParallelListComp = hasS isParComp
 used FunctionalDependencies = hasT (un :: FunDep GhcPs)
 used ImplicitParams = hasT (un :: HsIPName)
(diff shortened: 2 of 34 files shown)


shayne-fletcher:
sadly, this pr fails with macOS 9.2.2 same as https://github.com/ndmitchell/hlint/pull/1385. fixed in 9.2.3 and note also that 9.2.2 is now two minor versions behind (ghc-9.2.4 was released early aug). perhaps it's worth reconsidering if we might patch 'ci.yml' as per this comment https://github.com/ndmitchell/hlint/pull/1385#issuecomment-1179757012?

shayne-fletcher:
ubuntu ci fails at the end with 
```
--- a/hints.md
+++ b/hints.md
@@ -1420,13 +1420,13 @@ Example:
 </pre>
 Found:
 <pre>
-{-# LANGUAGE RebindableSyntax #-}
 {-# LANGUAGE EmptyCase, RebindableSyntax #-}
+{-# LANGUAGE RebindableSyntax #-}
 
 </pre>
 Suggestion:
 <code>
-{-# LANGUAGE RebindableSyntax, EmptyCase #-}
+{-# LANGUAGE EmptyCase, RebindableSyntax #-}
 </code>
 <br>
 </td>
 ```
 the reordering is expected but i don't know what to update where to fix this. @zliu41 perhaps you know?

**Comment by shayne-fletcher** [comment]
i think we'll be needing @ndmitchell 's final say on this & its timing but nonetheless, if you have the time & patience, your review would be very welcome @zliu41 😊 🙏 !

File: hlint.cabal [hunk 1]
```diff
@@ -53,7 +53,7 @@ flag gpl
     description: Use GPL libraries, specifically hscolour
 
 flag ghc-lib
-  default: False
+  default: True
```
ndmitchell, line 56: Good call - this seems like the right time to flip this default

src/Hint/Extensions.hs
```diff
@@ -98,7 +98,7 @@ deriving instance Show Bar -- {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE DeriveGeneric, GeneralizedNewtypeDeriving #-} \
 newtype Micro = Micro Int deriving Generic -- {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DeriveGeneric, TypeFamilies #-} \
-data family Bar a; data instance Bar Foo = Foo deriving (Generic)
```
ndmitchell, line 101: Why did this need to change?

ndmitchell (approved):
Nice work!

hlint.cabal
shayne-fletcher: jolly good. i've taken the liberty of pushing the change to `ci.yml` then that bumps the macOS compiler to ghc-9.2.3 .

src/Hint/Extensions.hs
shayne-fletcher: gosh. i'm generally not in the habit of making random changes but in this instance i'm stumped - i have no idea why i did that and accordingly have reverted the change. sorry!

shayne-fletcher: good catch & thank you for the careful review!

shayne-fletcher:
fixes https://github.com/ndmitchell/hlint/issues/1376

ndmitchell:
Thanks Shayne! I did leave some notes, which it appears GitHub completely dropped, so from memory:

> of course with 9.4.1 being now available, one hopes we can add 9.4 to that matrix in the near future

What stops us from doing that now?

> perhaps it's worth reconsidering if we might patch 'ci.yml' as per this comment

Why is what we get in CI lagging so far behind?

shayne-fletcher:
> What stops us from doing that now?

i just did a quick foray to find out:

  - ghc-9.4.1 was released sun 2022-08-07 (just this week):
    - there's an install issue on macOS (at least):

      - i've patched my local install to work-around this
      - expect `ghc-9.4.2` to be released this week to fix that:

        - see https://www.mail-archive.com/ghc-devs@haskell.org/msg20589.html
    - we don't have CI testing of `ghc-lib-parser`, `ghc-lib-parser-ex` w/9.4.1 yet
      - i have however verified my local builds via cabal w/9.4.1:

        - `ghc-lib-parser`, `ghc-lib-parser-ex` & `hlint` (9.4 parse tree)
  - now regarding hlint CI:
    - seems we can't say `ghc: ['9.4', ...]` in `ci.yml` for `ubuntu-latest`:

      - see https://github.com/ndmitchell/hlint/runs/7809786617
    - seems we can't say `ghc: '9.4'` in `ci.yml` for `windows-latest`:

      - see https://github.com/ndmitchell/hlint/runs/7809894556
    - we can however say `ghc: ['9.4.1', ...]` for `ubuntu-latest` & `ghc: '9.4.1'` for `windows-latest`

      -  here i skip `macOS-latest` due to the install issue mentioned above
    - however, both of these builds fail on `ghc-exactprint` (needed for `apply-refact-0.10.0.0`) (@zliu41  any chance you help with this?)
    - see https://github.com/ndmitchell/hlint/runs/7809368346:

```
          Building library for ghc-exactprint-1.5.0..
          [ 1 of 10] Compiling Language.Haskell.GHC.ExactPrint.Orphans ( src\Language\Haskell\GHC\ExactPrint\Orphans.hs, dist\build\Language\Haskell\GHC\ExactPrint\Orphans.o )
        
          src\Language\Haskell\GHC\ExactPrint\Orphans.hs:55:18: error:
          Error:     Not in scope: type constructor or class ‘AnnsLet’
          cabal.exe: Failed to build ghc-exactprint-1.5.0 (which is required by
             |
          exe:refactor from apply-refact-0.10.0.0). See the build log above for details.
          55 | instance Default AnnsLet where
             |                  ^^^^^^^
```

> Why is what we get in CI lagging so far behind?

sorry, i have no skills in this area - i'm just cargo-culting from what i find here. perhaps you have a contact somewhere  you might ask for us?

shayne-fletcher:
> What stops us from doing that now?

we may find ourselves waiting on some package upgrades also. for example, the ghc-lib project (ghc-lib-gen's dependence on aeson mostly i think) indicates that the upper-bounds at least these packages need relaxing to be built with ghc-9.4.1:

OneTuple
splitmix
attoparsec
scientific
hashable
integer-logarithms
primitive
scientific
data-fix
aeson
indexed-traversable
semialign
indexed-traversable-instances
unordered-containers
assoc
these
text-short
time-compat
uuid-types
split

hlint.cabal
sol: @ndmitchell @shayne-fletcher what's the rationale for this change?

For `ormolu` I was proposing to use `ghc` by default when possible (https://github.com/tweag/ormolu/pull/948). Are there any potential drawbacks with that?

shayne-fletcher: @sol the motivation is discussed in https://github.com/ndmitchell/hlint/issues/1376. i don't know of any drawbacks. you may find [this blog post](https://blog.shaynefletcher.org/2021/02/configuring-cabal-build-flags.html) helpful.

PR #1411, opened 2022-08-12
hints about empty strings/lists

jvoigtlaender:
I noticed that things like `x ++ []` were already detected by hlint, but `x ++ ""` wasn't. So I went through the hints file and tried to add all meaningful variants of hints on empty strings/lists where one variant was "missing".

I actually didn't add all. In https://github.com/ndmitchell/hlint/blob/a62a3b56d7e76715633d266939f40ac9b2e94301/data/hlint.yaml#L629 there would have been altogether three variants to add.

Likewise in https://github.com/ndmitchell/hlint/blob/a62a3b56d7e76715633d266939f40ac9b2e94301/data/hlint.yaml#L834
In the latter case, I actually wondered (independently of the matter of strings vs. lists) why there aren't hints firing on `zip [] x` and `zip x []` (instead of just on `zip [] []`).

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -130,6 +130,7 @@
     # READ/SHOW
 
     - warn: {lhs: showsPrec 0 x "", rhs: show x}
+    - warn: {lhs: "showsPrec 0 x []", rhs: show x}
     - warn: {lhs: readsPrec 0, rhs: reads}
     - warn: {lhs: showsPrec 0, rhs: shows}
     - hint: {lhs: showIntAtBase 16 intToDigit, rhs: showHex}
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks for the patch. I've applied some of these (as a separate patch), but not all:

* Things like anyM on `""` seem super unlikely. I know someone could technically do it, but I doubt it crops up in reality.
* Lots of people use overloaded strings, meaning `""` isn't quite equivalent to `[]` - that means that `null x` and `x == ""` don't do the same thing in many cases.

I've pushed a subset in 8ccf7eb17cd5548e425a4830cdd498ed644a8b61. Let me know if you think any of the ones I've missed might still be useful.

PR #1417, opened 2022-08-22
dualized four comparisons

jvoigtlaender:
See https://github.com/ndmitchell/hlint/pull/1404#issuecomment-1207292967


ndmitchell:
Thanks!

PR #1418, opened 2022-08-30
quickcheck selection functions on singleton lists

jvoigtlaender:
The thing I actually encountered in code was of the form
```haskell
  x <- oneof [y]
```
(after some previous refactoring).

I guess the other hints, for `elements` etc., could also be encountered.


ndmitchell:
Thanks!

PR #1421, opened 2022-09-05
zip/repeat -> map with tuple section

jvoigtlaender:
The two new hints are mirroring existing `zipWith`/`repeat` hints (directly above in the file).


ndmitchell:
Thanks! I was a little concerned about suggesting tuple sections without the user already using it, but I see some existing hints already do that, so can't object to this one doing so too.

PR #1424, opened 2022-10-15
Don't suggest sortOn for Down, do avoid rev sort

### LINKED ISSUE #1423: `sortOn Down` is suggested in place of cheaper `sortBy (compare Down)` and `reverse (sort x)` is not corrected
#669 covers this in some detail, but it doesn't appear that the [fix](https://github.com/ndmitchell/hlint/commit/39334297702e4f98bf0c9d4eabfd914c216fc4dd) agreed entirely with the conclusions in Roman Cheplyaka's [article](https://ro-che.info/articles/2016-04-02-descending-sort-haskell). In particular:

1. `sortOn Down` is, according to the article, some 7x slower than `sortBy (compare Down)`. The former should not be suggested for the latter. This is actually a bit more nuanced: It seems that `sortOn f` is only appropriate when `f` is sufficiently expensive to overwhelm the cost of caching. When `f` is `Down`, the hint should not be fired, and the user should otherwise be warned of the nuance.
2. `reverse (sort x)` is inefficient, especially when not all items from `x` are required. It also isn't stable. It should be replaced, but with `sortBy (comparing Down)`, not `sortOn Down`.

Regarding the PR, two issues:

1. It seems that LHS matching isn't equiped to equate plain and qualified names, so the condition `notEq f Down && notEq f Data.Ord.Down` is not redundant. If I can do better, please let me know.
2. While manually testing using `stack run -- hlint my-tests-file.hs` produces the desired result, the tests that I added are not passing (the expected hints are not being fired). I'm fairly new to haskell and unfamiliar with the test generation method, so I've probably mucked up something obvious. Guidance would be appreciated.

thisisrandy:
Fixes #1423, but new tests are not passing for unknown reasons

Thanks for the pull request!

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -112,8 +112,7 @@
     - warn: {lhs: head (sortBy f x), rhs: minimumBy f x, side: isCompare f}
     - warn: {lhs: last (sortBy f x), rhs: maximumBy f x, side: isCompare f}
     - warn: {lhs: reverse (sortBy f x), rhs: sortBy (flip f) x, name: Avoid reverse, side: isCompare f, note: Stabilizes sort order}
-    - warn: {lhs: sortBy (flip (comparing f)), rhs: sortOn (Down . f)}
-    - warn: {lhs: sortBy (comparing f), rhs: sortOn f, side: notEq f fst && notEq f snd && notEq f Down && notEq f Data.Ord.Down, note: Only appropriate if f is expensive enough to justify the cost of caching its results. See the documentation for sortOn}
+    - warn: {lhs: sortBy (flip (comparing f)), rhs: sortBy (comparing (Down . f))}
     - warn: {lhs: reverse (sortOn f x), rhs: sortOn (Data.Ord.Down . f) x, name: Avoid reverse, note: Stabilizes sort order}
     - warn: {lhs: reverse (sort x), rhs: sortBy (comparing Data.Ord.Down) x, name: Avoid reverse, note: Stabilizes sort order}
     - hint: {lhs: flip (g `on` h), rhs: flip g `on` h, name: Move flip}
@@ -1419,9 +1418,8 @@
 # main = hello .~ Just 12 -- hello ?~ 12
 # foo = liftIO $ window `on` deleteEvent $ do a; b
 # no = sort <$> f input `shouldBe` sort <$> x
-# sortBy (comparing length) -- sortOn length
 # no = sortBy (comparing Down)
-# yes = sortBy (comparing f) -- sortOn f
+# yes = sortBy (flip (comparing f)) -- sortBy (comparing (Down . f))
 # yes = reverse (sort x) -- sortBy (comparing Data.Ord.Down) x
 # myJoin = on $ child ^. ChildParentId ==. parent ^. ParentId
 # foo = typeOf (undefined :: Foo Int) -- typeRep (Proxy :: Proxy (Foo Int))
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -113,10 +113,9 @@
     - warn: {lhs: last (sortBy f x), rhs: maximumBy f x, side: isCompare f}
     - warn: {lhs: reverse (sortBy f x), rhs: sortBy (flip f) x, name: Avoid reverse, side: isCompare f, note: Stabilizes sort order}
     - warn: {lhs: sortBy (flip (comparing f)), rhs: sortOn (Down . f)}
-    - warn: {lhs: sortBy (comparing f), rhs: sortOn f, side: notEq f fst && notEq f snd}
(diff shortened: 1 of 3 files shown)


thisisrandy:
I want to emphasize, as noted in #1423, that the tests I added don't seem to be working, and I'm not sure why. For that reason, please do **not** accept this PR as is. I would appreciate some insight on how to get the tests working.

tests/hint.test
```diff
@@ -239,3 +239,35 @@ Found:
 Note: Custom message
 
 4 hints
+
+---------------------------------------------------------------------
```
ndmitchell, line 243: Normally end to end tests like this are excessive for simple rules - see the `<TEST>` block in `data/hlint.yaml` which will run faster and be easier to write

data/hlint.yaml
```diff
@@ -113,10 +113,9 @@
     - warn: {lhs: last (sortBy f x), rhs: maximumBy f x, side: isCompare f}
     - warn: {lhs: reverse (sortBy f x), rhs: sortBy (flip f) x, name: Avoid reverse, side: isCompare f, note: Stabilizes sort order}
     - warn: {lhs: sortBy (flip (comparing f)), rhs: sortOn (Down . f)}
-    - warn: {lhs: sortBy (comparing f), rhs: sortOn f, side: notEq f fst && notEq f snd}
+    - warn: {lhs: sortBy (comparing f), rhs: sortOn f, side: notEq f fst && notEq f snd && notEq f Down && notEq f Data.Ord.Down, note: Only appropriate if f is expensive enough to justify the cost of caching its results. See the documentation for sortOn}
```
ndmitchell, line 116: I think rather than warning about a fairly complex condition, it might be better to just delete this hint entirely

data/hlint.yaml
```diff
@@ -113,10 +113,9 @@
     - warn: {lhs: last (sortBy f x), rhs: maximumBy f x, side: isCompare f}
     - warn: {lhs: reverse (sortBy f x), rhs: sortBy (flip f) x, name: Avoid reverse, side: isCompare f, note: Stabilizes sort order}
     - warn: {lhs: sortBy (flip (comparing f)), rhs: sortOn (Down . f)}
```
ndmitchell, line 115: Sounds like this one also needs fixing/deleting?

ndmitchell (commented):
Thanks for the diff! A few notes - I wonder if we should just never suggest introducing sortOn?

tests/hint.test
thisisrandy: I'm still in the dark as to why the e2e tests were failing, but thanks for the suggestion. This is covered in 013928c01a63060946e7a254af93c928337acc24.

data/hlint.yaml
thisisrandy: I agree. While there are obviously cases where `sortOn` is the right choice, a user should probably be specifically wanting to optimize their expensive sort and hopefully come across `sortOn` in their own research. Covered in 86363703feaa7e1abf2d41fe458512f3943beab4.

data/hlint.yaml
thisisrandy: Covered in 86363703feaa7e1abf2d41fe458512f3943beab4.

data/hlint.yaml
```diff
@@ -112,11 +112,9 @@
     - warn: {lhs: head (sortBy f x), rhs: minimumBy f x, side: isCompare f}
     - warn: {lhs: last (sortBy f x), rhs: maximumBy f x, side: isCompare f}
     - warn: {lhs: reverse (sortBy f x), rhs: sortBy (flip f) x, name: Avoid reverse, side: isCompare f, note: Stabilizes sort order}
-    - warn: {lhs: sortBy (flip (comparing f)), rhs: sortOn (Down . f)}
-    - warn: {lhs: sortBy (comparing f), rhs: sortOn f, side: notEq f fst && notEq f snd}
+    - warn: {lhs: sortBy (flip (comparing f)), rhs: sortBy (comparing (Down . f))}
```
zliu41, line 115:
```suggestion
    - warn: {lhs: sortBy (flip (comparing f)), rhs: sortBy (comparing (Data.Ord.Down . f))}
```

zliu41 (commented):
Looks good, just one suggested change.

zliu41:
Oh and also you need to regenerate hints.md for the test to pass.

data/hlint.yaml
thisisrandy: Thanks for the suggestion. The corresponding test also needs to use the fully-qualified name. Addressed in 018facf7.

thisisrandy:
Apologies, I forgot to regenerate hints after the review commits. Addressed in cb124475.

thisisrandy:
The failure was "API rate limit exceeded", so it should probably be rerun or ignored, depending on maintainers' preference.

zliu41:
Thanks for the PR.

PR #1428, opened 2022-10-27
Rename map tuple-section hints

### LINKED ISSUE #1427: Hints with the same name
The only way to ignore hints is by their name, but names are not unique; for example, according to `hints.md`, there are 8 different hints named "Use map":

```ShellSession
$ rg '<td>Use map</td>' hints.md
755:<td>Use map</td>
2843:<td>Use map</td>
4187:<td>Use map</td>
4203:<td>Use map</td>
4219:<td>Use map</td>
4235:<td>Use map</td>
9151:<td>Use map</td>
9167:<td>Use map</td>
```

Two of these hints introduce tuple sections, which I would like to hide, but I have no way to select them.

Maybe we could work on getting each hint a unique numeric identifier?

9999years:
Closes #1427


ndmitchell:
Thanks!

Issue #1434, opened 2022-12-15
HLint seems to ignore -XNumericUnderscores


friedbrice:
HLint does not seem to correctly report `"Use underscore"` hints unless a file exlicitly has a `NumericUnderscores` pragma, despite specifying `-XNumericUnderscores` in `.hlint.yaml`.

Repo: https://github.com/friedbrice/hlint-use-underscore

PR #1435, opened 2022-12-16
Add support for NumericUnderscores extensions from CLI/config

friedbrice:
Closes https://github.com/ndmitchell/hlint/issues/1434

### DIFF
--- a/src/GHC/All.hs
+++ b/src/GHC/All.hs
@@ -5,7 +5,7 @@ module GHC.All(
     CppFlags(..), ParseFlags(..), defaultParseFlags,
     parseFlagsAddFixities, parseFlagsSetLanguage,
     ParseError(..), ModuleEx(..),
-    parseModuleEx, createModuleEx, createModuleExWithFixities, ghcComments, modComments,
+    parseModuleEx, createModuleEx, createModuleExWithFixities, createModuleExWithFixitiesAndExtensions, ghcComments, modComments,
     parseExpGhcWithMode, parseImportDeclGhcWithMode, parseDeclGhcWithMode,
     ) where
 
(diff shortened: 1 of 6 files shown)


friedbrice:
@ndmitchell Is there a different way you'd like me to approach this problem? I'm happy to rework this.

src/Hint/NumLiteral.hs
```diff
@@ -34,10 +36,16 @@ import Idea (Idea, suggest)
...
+  -- TODO: there's a subtle bug when the module disables `NumericUnderscores`.
+  -- This seems pathological, though, because who would enable it for their
+  -- project but disable it in specific files?
+  if NumericUnderscores `elem` activeExtensions then
      concatMap suggestUnderscore . universeBi
   else
      const []
+  where
+    moduleExtensions = toList (extensions $ modComments modu)
+    activeExtensions = hlintExtensions modu <> toList moduleExtensions
```
googleson78, line 48: I think the `toList` here is redundant - you've already called `toList` inside `moduleExtensions`, right?

src/GHC/All.hs
```diff
@@ -84,8 +84,9 @@ data ParseError = ParseError
     }
 
 -- | Result of 'parseModuleEx', representing a parsed module.
-newtype ModuleEx = ModuleEx {
-    ghcModule :: Located HsModule
+data ModuleEx = ModuleEx {
+    ghcModule :: Located HsModule,
+    hlintExtensions :: [Extension]
```
googleson78, line 89: I think some name with `config` might be clearer - it took me a while to realise that this `hlint` means "it comes from invoking the `hlint` tool with certain CLI args or a configuration file"

```suggestion
    hlintConfigExtensions :: [Extension]
```
Perhaps?

friedbrice:
Thanks, @googleson78! I'll start working on a patch.

ndmitchell:
Sorry it took me so disastrously long to get to this diff - it feel off my radar. I updated the merge conflicts, but didn't test after (did it in the github UI).

I can see how this would apply to people using the API, but as far as I can tell, this won't have any impact on the command line users? Or are you expecting that this will be used through the Haskell IDE?

friedbrice:
@ndmitchell it works using our `.hlint.yaml` file. If it doesn't work for CLI uses, then that's an oversight on my part that I need to fix.

parsonsmatt:
@ndmitchell anything we can do to get this ready to merge?

PR #1437, opened 2022-12-21
When possible, preserve across hints whether or not a Monad constraint is required

josephcsible:


ndmitchell:
Thanks - very good idea!

PR #1438, opened 2023-01-04
Don't suggest NumericUnderscores for four-digit numbers

dustin:
This is related to #578 and excludes the common case of four digit integers.


ndmitchell:
Thanks, this looks great!

PR #1447, opened 2023-02-08
dualized <$> hints with <&>

jvoigtlaender:
Done fairly mechanically. Left out a select few further opportunities, involving `Just` applications: https://github.com/ndmitchell/hlint/blob/1591b773dabbede12912cd43dabe73a578502103/data/hlint.yaml#L593-L594 https://github.com/ndmitchell/hlint/blob/1591b773dabbede12912cd43dabe73a578502103/data/hlint.yaml#L973-L974 https://github.com/ndmitchell/hlint/blob/1591b773dabbede12912cd43dabe73a578502103/data/hlint.yaml#L1056

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -142,6 +142,7 @@
 
     - warn: {lhs: concat (map f x), rhs: concatMap f x}
     - warn: {lhs: concat (f <$> x), rhs: concatMap f x}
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks, makes sense! Agree on the ones you left out.

PR #1448, opened 2023-02-08
<$>/<&> hints for Either

jvoigtlaender:
Mirroring the `fmap` hints already existing in that context.


jvoigtlaender:
Thanks for merging. Now saw that maybe I had copy-pasted to aggressively: should the names of the first two new hints be changed, since `fmap` does not literally appear in them? 
```yaml
    - warn: {lhs: either f g (h <$> x), rhs: either f (g . h) x, name: Redundant fmap}
    - warn: {lhs: either f g (x <&> h), rhs: either f (g . h) x, name: Redundant fmap}
```

PR #1449, opened 2023-02-08
fmap hints specialized to list-map (where type-wise possible)

jvoigtlaender:
In other occurrences of `fmap` in the hint base, this specialization is not possible since, e.g., the context implies that the `fmap` is used on `Maybe` type or so.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -489,9 +489,11 @@
     - hint: {lhs: pure . f =<< m, rhs: f <$> m}
     - hint: {lhs: return . f =<< m, rhs: f <$> m}
     - warn: {lhs: fmap f x >>= g, rhs: x >>= g . f}
+    - warn: {lhs: map f x >>= g, rhs: x >>= g . f}
     - warn: {lhs: f <$> x >>= g, rhs: x >>= g . f}
     - warn: {lhs: x Data.Functor.<&> f >>= g, rhs: x >>= g . f}
     - warn: {lhs: g =<< fmap f x, rhs: g . f =<< x}
+    - warn: {lhs: g =<< map f x, rhs: g . f =<< x}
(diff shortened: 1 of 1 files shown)


ndmitchell:
Generally when we are changing the type so it is more general you can run into issues with type inference suddenly failing and it can be less clear to the reader (things like `map` help anchor them). Therefore HLint usually doesn't make such suggestions as default. There are sections of `generalize` and `generalize-for-conciseness` which some people use. Some of these are indeed more concise, so probably fit into the second category. Maybe all of them?

jvoigtlaender:
As you see fit.

The last two, about `concatForM` and `concatForM_`, are already in a by-default-disabled group, for the `extra` package. Maybe there is no point in creating a separate group `extra-package-and-generalizing`?

Also, I couldn't find `concatForM_` now anymore. Did I previously phantasize its existence, or was it dropped from `extra`?

ndmitchell:
`concatForM_` doesn't exist because if you are discarding the result, there is no point in first concatenating the elements. No need for a special `extra-package-and-generalizing`, I think given they are off by default we have a little more latitude there anyway.

jvoigtlaender:
> `concatForM_` doesn't exist ...

But the current `master` branch contains hints introducing it. :smile:
https://github.com/ndmitchell/hlint/blob/af1366e971014ac51b04ee33b5dc7b7509b506af/data/hlint.yaml#L1198-L1200

ndmitchell:
Oops! Now gone in 0cfa715762b590ef1fe97ee62e59945d6c85a968

jvoigtlaender:
Also from this pull request.

jvoigtlaender:
The remaining hints added in this PR are now all in off-by-default groups.

ndmitchell:
Thanks!

PR #1450, opened 2023-02-08
introducing nubSort

jvoigtlaender:
The first of these is a "defining property" stated in the documentation of `nubSort`.

The second one, `sort (nub x)`, is something I spotted in student code and seems to be even more deserving of a rewrite, I think, since the inner `nub` is doing excess work if we later want to sort anyway.

Is there need for some kind of warning here that the `Ord`/`Eq` instances are assumed to be defined sensibly?

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -1117,6 +1117,8 @@
     - warn: {lhs: "notNull [x]", rhs: "True", name: Evaluate}
     - warn: {lhs: "notNull []", rhs: "False", name: Evaluate}
     - hint: {lhs: "\\(x,y) -> (f x, f y)", rhs: both f}
+    - hint: {lhs: nub (sort x), rhs: nubSort x}
+    - warn: {lhs: sort (nub x), rhs: nubSort x}
 
 # hints that will be enabled in future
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Side question:

A few lines above that in the code, I see these hints: https://github.com/ndmitchell/hlint/blob/d94286960088445413ea160eada06fd7129f9cd5/data/hlint.yaml#L1111-L1112 I don't understand why those are not written like this instead:
```yaml
    - warn: {lhs: "head (x ++ [y])", rhs: "headDef y x"}
    - warn: {lhs: "last (x : y)", rhs: "lastDef x y"}
```
Aren't we writing hints with "silent function application" and rely on HLint's ability to "see throught `$`", instead of the other way around?

ndmitchell:
I think if you are sending it through sort, you have to assume that Ord is sensible, since the details of sort on a unsensible Ord is going to be highly unpredictable. As a result, moving from sort to nubSort might change things, but its basically random.

For the `headDef $` stuff you spot, I'd say this should definitely have been written without the `$`. Want to send a fix?

PR #1454, opened 2023-02-10
replace map by mapMaybe in context of filter isJust

jvoigtlaender:
Not sure you consider these hints always worthwhile, but thought I would at least propose them.

A meaningful rule name would also still be needed.

The reason I consider this rewrite worthwhile when the opportunity arises is that through the `map Just` it is then more explicit "for the compiler" that the output list contains only `Just` values, potentially enabling more fusion etc.


ndmitchell:
Seem reasonable hints. What is the inferred name? I'd imagine `Use mapMaybe` would be a good enough hint. Although I worry it defaults to `Use map`?

jvoigtlaender:
The default name for the first hint was `Use Just`, for the others `Use map`.

I have now set `Use mapMaybe` for all four hints.

ndmitchell:
Thanks!

PR #1455, opened 2023-02-12
Warn that .hlint.yml is not a substitute for .hlint.yaml.

philderbeast:
Adds a footnote that configuration in `.hlint.yml` won't be picked up by default.


ndmitchell:
Good idea - but not convinced by the use of a footnote. Browsing it on GitHub, the footnote looks enticing, but when I click, I lose my context and jump to the bottom.  If you just delete the footnote and inline the text that looks useful.

ndmitchell:
Thanks!

PR #1459, opened 2023-02-14
Remove/empty haddock empty comment

philderbeast:
Can we suggest removing empty comments and removing empty haddocks? I've seen examples of these littering code, like the following:

```haskell
-- |
foo = 1
```

I had a crack at doing this.

- If I run `cabal run hlint -- --refactor tests/empty-comment-test.hs` the comments are removed but single line comments are replaced by an empty line. I'd really like to remove the whole line. Is there a way to do that? I tried `advanceSrcLoc (realSrcSpanEnd realSpan) '\n'` for the refactor span but this didn't work as I'd hoped.
- I'm not quite sure what is the best way to convert `tests/empty-comment-test.hs` into tests.

```
$ cabal run hlint -- --refactor --refactor-options=inplace tests/empty-comment-test.hs
$ git diff
```

```diff

diff --git a/tests/empty-comment-test.hs b/tests/empty-comment-test.hs
index 33b38157..84811b24 100644
--- a/tests/empty-comment-test.hs
+++ b/tests/empty-comment-test.hs
@@ -1,11 +1,11 @@
--- |
+
 haddockAboveSingle = 1
 
 haddockBelowSingle = 1
--- ^
 
--- |
---
+
+
+
 haddockAboveSingles = 1
 
 -- |
@@ -24,51 +24,45 @@ haddockBelowIntoNonEmpty = 1
 --
 -- foo
 
-{--}
+
 commentMultiEmpty = 1
 
-{- -}
+
 commentMultiEmptySpace = 1
 
-{-
 
-    -}
 commentMultiEmptyLines = 1
 
-{- |-}
+
 haddockAboveMultiEmpty = 1
 
-{- | -}
+
 haddockAboveMultiEmptySpace = 1
 
-{- |
 
-    -}
 haddockAboveMultiEmptyLines = 1
 
 haddockBelowMultiEmpty = 1
-{- ^-}
+
 
 haddockBelowMultiEmptySpace = 1
-{- ^-}
+
 
 haddockBelowMultiEmptyLines = 1
-{- ^
 
-    -}
 
 data HaddockSingle = HaddockSingle
   { haddockFieldEmptyBelow :: Int
-  -- ^
-  -- |
+  
+  
   , haddockFieldAboveEmpty :: Int
-  , haddockFieldEmptyAfter :: Int -- ^
+  , haddockFieldEmptyAfter :: Int 
   }
 
 data HaddockMulti = HaddockMulti
   { haddockFieldEmptyBelow :: Int
-  {- ^-}
-  {- |-}
+  
+  
   , haddockFieldAboveEmpty :: Int
-  , haddockFieldEmptyAfter :: Int {- ^-}
-  }
\ No newline at end of file
+  , haddockFieldEmptyAfter :: Int 
+  }
```

### DIFF
--- a/tests/empty-comment-test.hs
+++ b/tests/empty-comment-test.hs
@@ -1,0 +1,74 @@
+-- |
+haddockAboveSingle = 1
+
+haddockBelowSingle = 1
+-- ^
+
+-- |
+--
+haddockAboveSingles = 1
+
+-- |
+-- >>>
+haddockAboveDoctest = 1
+
+-- |
+--
+--
+-- foo
+haddockAboveIntoNonEmpty = 1
+
+haddockBelowIntoNonEmpty = 1
+-- ^
+--
+--
+-- foo
+
+{--}
+commentMultiEmpty = 1
+
+{- -}
+commentMultiEmptySpace = 1
+
+{-
+
+    -}
+commentMultiEmptyLines = 1
+
+{- |-}
+haddockAboveMultiEmpty = 1
+
+{- | -}
+haddockAboveMultiEmptySpace = 1
+
+{- |
+
+    -}
+haddockAboveMultiEmptyLines = 1
+
+haddockBelowMultiEmpty = 1
+{- ^-}
+
+haddockBelowMultiEmptySpace = 1
+{- ^-}
+
+haddockBelowMultiEmptyLines = 1
+{- ^
+
+    -}
+
+data HaddockSingle = HaddockSingle
+  { haddockFieldEmptyBelow :: Int
+  -- ^
+  -- |
+  , haddockFieldAboveEmpty :: Int
+  , haddockFieldEmptyAfter :: Int -- ^
+  }
+
+data HaddockMulti = HaddockMulti
+  { haddockFieldEmptyBelow :: Int
+  {- ^-}
+  {- |-}
+  , haddockFieldAboveEmpty :: Int
+  , haddockFieldEmptyAfter :: Int {- ^-}
+  }
--- a/src/Hint/Comment.hs
+++ b/src/Hint/Comment.hs
@@ -33,40 +33,43 @@ directives = words $
 
 
 commentHint :: ModuHint
-commentHint _ m = concatMap (chk singleLines singleSomeLines) comments
+commentHint _ m = concatMap (check singleLines someLines) comments
   where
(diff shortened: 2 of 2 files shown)


tests/empty-comment-test.hs
```diff
@@ -0,0 +1,74 @@
+-- |
```
ndmitchell, line 1: This file doesn't actually seem to be used in a test suite?

ndmitchell:
No idea on the refactor thing - @zliu41 is the expert there.

Can you turn your test script into something that runs via the `.test` files in the `tests` directory?

tests/empty-comment-test.hs
philderbeast: Sorry I didn't see this earlier. I've got the hang of hlint `.test` testing now and converted that one `.hs` file into a `.test` file with multiple tests within it.

philderbeast:
> Can you turn your test script into something that runs via the `.test` files in the `tests` directory?

Thanks for the suggestion @ndmitchell, I've done that now.

ndmitchell:
I note this is still in draft - is there something else you are hoping to do here before I take a look?

philderbeast:
Thanks @ndmitchell but this is not ready for review (and apologies for reworking it on this PR branch and not on another branch).

I had an initial implementation that detected simple empty comments, like these;

* the empty multi-line haddock comment.
* the empty single line haddock comment or run of comments.

As I did more testing I found more types of empty comment that I'd like to include, doctest comments in particular;

* the empty doctest leader `>>>`
* the empty multi-line doctest `:{ .. :}`

If an otherwise empty haddock comment includes a empty doctest comment then which new HLint suggestion should take precedence or should I report both?

```
-- |
-- >>>
--
```

I'm also considering reporting leading or trailing blank lines in haddocks. For instance, in the Agda source there are many trailing blank lines in its haddocks.

```
-- | A symbolic library name.
--
type LibName = String
```

Should there be separate HLint suggestions for trimming leading or trailing blank lines? Should the same go for trimming a run of blank lines down to single blank line in otherwise non-empty comments?

I'm in the process of redoing the implementation to first partition comment types and then do the detection as before.

In the earlier implementation I processed single line comments one at a time but I think I could provide more nuanced empty comment detection if I treated a run of comments (with a haddock leader) like a multi-line haddock comment.

PR #1463, opened 2023-02-26
rename hints

jvoigtlaender:
Just in case. See https://github.com/ndmitchell/hlint/pull/1448#issuecomment-1445360273.


ndmitchell:
Yep, seems like a good idea

Issue #420, opened 2018-01-04
Warn about partial functions


mjepronk:
Would it be an idea if HLint would warn about partial functions like `head`?

Of course the list should be hardcoded (since there is no way to automatically identify them). 

I couldn't find if this has been discussed previously, so sorry if it was.

ndmitchell:
It seems reasonable, however, I've resisted for a few reasons:

* Which functions are "partial" isn't entirely clear. Is `readFile`? It can indeed throw errors.
* I'd much rather people used partial functions safely, and even devoted several years of my life to that (see https://github.com/ndmitchell/catch)
* A static checker which made sure all partial functions had a `Partial =>` annotation would be a much better approximation.
* It certainly can't be on by default, because using partial functions isn't always a bad idea.

ndmitchell:
Solution is to have a group in `.hlint.yaml` which is off by default, then people can turn on a warning of groups of partial functions. I'm happy for someone to contribute this.

ndmitchell:
Probably good to start with the list in https://wiki.haskell.org/Avoiding_partial_functions

anka-213:
I think one important distinction to make is partial pure functions and partial impure functions. If you're in `IO`, you'll probably expect operations to fail, not as much so in pure code. `readFile` is kind of a corner-case though, since it is lazyIO, which can cause crashes in pure code.

ndmitchell:
Having a warning for lazy-IO functions as a separate set would be possible. I leave it to what people want to contribute.

PR #1467, opened 2023-03-05
Warn on partial functions.

chungyc:
The warnings are for partial functions in the `base` package.  There are two groups that are disabled by default: `partial` for functions that should be relatively easy to avoid, and `partial-strict` for functions whose use may not be feasible to avoid.

Suggestions have been added for partial functions in the `partial` group where they can be reasonably added.  For those that must necessarily remain partial, `error _` is used as the value for the undefined cases.  This both serves as a signal that the case remains undefined, and forces it to be rewritten appropriately by being a syntax error.

By raising this pull request I confirm I am licensing my contribution under all licenses that apply to this project, and that I have no patents covering my contribution.

This is a pull request for https://github.com/ndmitchell/hlint/issues/420.

### DIFF
--- a/data/default.yaml
+++ b/data/default.yaml
@@ -51,6 +51,9 @@
 #
 # Generalise map to fmap, ++ to <>
 # - group: {name: generalise, enabled: true}
+#
+# Warn on use of partial functions
+# - group: {name: partial, enabled: true}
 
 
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks!

PR #1475, opened 2023-03-21
rules for (elem False) and (notElem True)

jvoigtlaender:
Among other things, these would allow more fusion. For example, I had a student code:
```haskell
False `elem` map (...) (...)
```
There is no rule about `elem` and `map`, but with one of the newly proposed rules the above becomes
```haskell
any not (map (...) (...))
```
after which https://github.com/ndmitchell/hlint/blob/23d5d2211892935167862d72905b809cc0693dad/data/hlint.yaml#L192 can fire.


ndmitchell:
Thanks!

PR #1476, opened 2023-03-21
generalise map-consuming rules from lists to Foldables/Traversables

jvoigtlaender:
All these rules are counterparts to rules that already exist for `map`, but where the surrounding functions are actually more general than just applying to lists.


ndmitchell:
Thanks!

Issue #1471, opened 2023-03-13
OverloadedRecordDot usage doesn't trigger redundant bracket warning


ocharles:
The following code triggers a redundant bracket warning:

```haskell
    referenceDesignator = ReferenceDesignator (placementReferenceDesignator)
```

However, the following doesn't:

```haskell
    referenceDesignator = ReferenceDesignator (p.placementReferenceDesignator)
```

This is a bug, as the brackets here are redundant.

benbellick:
@zliu41 I would like to start by trying to take on this issue. I have been familiarizing myself with the codebase, but if I have questions that come up, can I ask them to you? If so, is this the best place to ask?

zliu41:
Yes, feel free to post your questions here.

benbellick:
@zliu41 I have opened a PR. I fixed the problem by treating x.y as an atom. Does that make sense in context? I am honestly not sure what exactly Haskell considers an atom versus not. 

Also, is there a generally good way to familiarize yourself with a Haskell codebase? I did most of the fixing by visual inspecting, but it would have been quite convenient if I could have somehow seen every time a function was called along with its arguments, kind of akin to a debugger in an imperative-style programming language.

Anyways, thanks!

zliu41:
Fixed in https://github.com/ndmitchell/hlint/pull/1478

PR #1478, opened 2023-03-21
Fix no warn on (x.y) by treating x.y as atom

benbellick:
This is to resolve issue #1471

### DIFF
--- a/src/GHC/Util/Brackets.hs
+++ b/src/GHC/Util/Brackets.hs
@@ -38,6 +38,8 @@ instance Brackets (LocatedA (HsExpr GhcPs)) where
       HsUnboundVar{} -> True
       -- Technically atomic, but lots of people think it shouldn't be
       HsRecSel{} -> False
+      -- Only relevant for OverloadedRecordDot extension
+      HsGetField{} -> True
       HsOverLabel{} -> True
       HsIPVar{} -> True
       -- Note that sections aren't atoms (but parenthesized sections are).
--- a/src/Hint/Bracket.hs
+++ b/src/Hint/Bracket.hs
@@ -50,7 +50,7 @@ referenceDesignator = ReferenceDesignator (p.placementReferenceDesignator)
 
 -- record dot syntax
 {-# LANGUAGE OverloadedRecordDot #-} \
-referenceDesignator = ReferenceDesignator (p.placementReferenceDesignator) -- p.placementReferenceDesignator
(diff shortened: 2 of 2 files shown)


shayne-fletcher:
the fix looks good but the test doesn't show it. `OverloadedRecordDot` is not default enabled so in `issue1471 = (x.y)`, `(x.y)` is parsed to a parenthesized function composition and will generate a redundant brackets hint both before your fix and after it.

i suggest these tests from #1471:
```haskell
-- no record dot syntax
referenceDesignator = ReferenceDesignator (p.placementReferenceDesignator)

-- record dot syntax
{-# LANGUAGE OverloadedRecordDot #-} \
referenceDesignator = ReferenceDesignator (p.placementReferenceDesignator) -- p.placementReferenceDesignator
``` 
as reported in the ticket, for the second test, before your fix it won't generate a redundant bracket hint but after your fix it will!

benbellick:
Hey Shayne, thanks for the comments! I have updated the tests. Just to be clear, in the first case (w/o record dot syntax), the expression `p.placementReferenceDesignator` is parsed as `p . placementReferenceDesignator`, i.e. as two functions combined with the composition operator `.`. Is that correct?

shayne-fletcher (approved):
looks good!

shayne-fletcher:
>  Just to be clear, in the first case (w/o record dot syntax), the expression `p.placementReferenceDesignator` is parsed as `p . placementReferenceDesignator`, i.e. as two functions combined with the composition operator `.`. Is that correct? 

yes. if `OverloadedRecordDot` is not in effect the `.` in the expression above is parsed as an `ITdot` token which denotes function composition.

benbellick:
It seems that OverloadedRecordDot was added in version [9.2](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/overloaded_record_dot.html), so version 9.0 is perhaps causing a test failure. Any suggestion on how to deal with this? Perhaps a way to ignore certain tests based off of version?

shayne-fletcher:
@benbellick if the second test were updated to
```haskell
-- record dot syntax
{-# LANGUAGE OverloadedRecordDot #-} \
referenceDesignator = ReferenceDesignator (p.placementReferenceDesignator) -- @Warning p.placementReferenceDesignator @NoRefactor: refactor requires GHC >= 9.2.1
```
i think it will pass the currently failing 9.0 test.

benbellick:
@shayne-fletcher Thanks for the help. I have updated the test!

shayne-fletcher:
Thanks for contributing to hlint!

benbellick:
My pleasure! I love Haskell and hope to keep contributing to the community however I can. Thanks for the help!

PR #1479, opened 2023-03-22
Specialize not-hints

jvoigtlaender:
I'm fairly confident about the hints in the first commit.

The hints in the "Hoist not" commit may be a bit more questionable.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -176,6 +176,8 @@
     - warn: {lhs: foldr (.) id l z, rhs: foldr ($) z l}
     - warn: {lhs: span (not . p), rhs: break p}
     - warn: {lhs: break (not . p), rhs: span p}
+    - warn: {lhs: span (notElem x), rhs: break (elem x), name: Use break}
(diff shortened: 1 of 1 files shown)


data/hlint.yaml
```diff
@@ -222,7 +224,9 @@
     - warn: {lhs: any id, rhs: or}
     - warn: {lhs: all id, rhs: and}
     - warn: {lhs: any (not . f) x, rhs: not (all f x), name: Hoist not}
+    - warn: {lhs: any (notElem y) x, rhs: not (all (elem y) x), name: Hoist not}
```
ndmitchell, line 227: I'd say these are more confusing in the replacement, and no more efficient, so better to leave them as is.

ndmitchell (commented):
Agreed that the hoist not ones don't seem to be better, and potentially worse (although not sure sufficiently worse that I'd encourage the other direction). The rest look good though.

jvoigtlaender:
I've just removed the "Hoist not" commit (and force-pushed).

ndmitchell:
Thanks!

PR #1481, opened 2023-03-24
systematic hints about concat and concatMap

jvoigtlaender:
Most of these avoid allocations or reduce the amount of work `concat`/`concatMap` has to do in traversal.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -155,6 +155,8 @@
     - warn: {lhs: take n (repeat x), rhs: replicate n x}
     - warn: {lhs: map f (replicate n x), rhs: replicate n (f x)}
     - warn: {lhs: map f (repeat x), rhs: repeat (f x)}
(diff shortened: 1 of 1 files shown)


ndmitchell:
CI gives me:

```
No name for the hint catMaybes (concatMap f x) ==> concatMap (catMaybes . f) x


No name for the hint filter f (concatMap g x) ==> concatMap (filter f . g) x


No name for the hint mapMaybe f (concatMap g x) ==> concatMap (mapMaybe f . g) x
```

Those would be good to supply.

jvoigtlaender:
I've pushed an additional commit naming them `Move concatMap out`, in anology to these existing rules: https://github.com/ndmitchell/hlint/blob/6c17bf7468a8ec7deff0b048b0af9778ca927ea3/data/hlint.yaml#L725-L737

Issue #1469, opened 2023-03-08
Support SARIF output


chungyc:
[SARIF](https://sarifweb.azurewebsites.net/) is an interchange format for static analysis results.  HLint analyzes code statically.  It would be nice if the results from HLint can be outputted in SARIF so that automatic tools can interpret the results.  In particular, HLint could be used in a code scanning workflow on GitHub.

Not sure if building SARIF support in HLint itself would be better, or whether it would be better to build a separate application which uses HLint as a library.  Either way, I'm willing to study SARIF and implement the support.

The [sarif](https://hackage.haskell.org/package/sarif) package may be useful here.

PR #1482, opened 2023-03-26
Support SARIF output

chungyc:
This adds the ability to output into SARIF from HLint.  SARIF is an open interchange format for exchanging results for static analyses.  In particular, GitHub uses SARIF for reading results from code scanning tools; this would allow HLint to be used as a code scanning tool in GitHub.  The output is complete enough such that output uploaded from a [test workflow](https://github.com/chungyc/hlint/blob/sarif-code-scanning/.github/workflows/scan.yaml) shows up as a code scanning alert.

The changes adds a `--sarif` flag and a "SARIF" module.  E.g., the following will result in the hints being output in SARIF.

```shell
$ hlint --sarif data/Test.hs
```

There are now quite a few output formats that HLint supports.  I.e., HLint-native JSON, Code Climate, HTML, Refactor, and now SARIF.  I was almost tempted to add a `--format=sarif` flag instead and put the module in "Format.SARIF", but I didn't in order to maintain consistency with existing practice.  However, if you would like such a change made, I would be _very_ happy to make it (obviously, I would keep the existing flags alone).

This change does not use the [sarif](https://hackage.haskell.org/package/sarif) package.  While it would have not added any new dependencies other than itself, I was a little worried by it having an extra module in the "Data.Aeson" namespace.  It is not on Stackage, either, so it would have been a little more work to get it working on stack.

This pull request will resolve https://github.com/ndmitchell/hlint/issues/1469.

---
By raising this pull request, I confirm I am licensing my contribution under all licenses that apply to this project and that I have no patents covering my contribution.

### DIFF
--- a/src/SARIF.hs
+++ b/src/SARIF.hs
@@ -1,0 +1,177 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
+
+{- |
+Description: Formats hlint ideas in the Statis Analysis Results Interchange Format (SARIF).
+License: BSD-3-Clause
+
+Supports the conversion of a list of HLint 'Idea's into SARIF.
+
+SARIF (Static Analysis Results Interchange Format) is an open interchange format
+for storing results from static analyses.
+-}
+module SARIF ( printIdeas
+             , showIdeas
+             , toJSONEncoding
+             -- * See also
+             --
+             -- $references
+             ) where
+
+import Data.Aeson hiding (Error)
+import Data.Aeson.Encoding
+import Data.ByteString.Lazy (ByteString)
+import qualified Data.ByteString.Lazy as B
+import Data.Text.Lazy (Text)
+import Data.Version (showVersion)
+import GHC.Util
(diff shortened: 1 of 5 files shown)


src/SARIF.hs
```diff
@@ -0,0 +1,177 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
+
+{- |
+Description: Formats hlint ideas in the Statis Analysis Results Interchange Format (SARIF).
+License: BSD-3-Clause
```
ndmitchell, line 6: Any reason to add a license to just this one file, given the entire project is BSD3 licensed?

ndmitchell:
Looks good - thanks for the very well structured pull request!

src/SARIF.hs
chungyc: I fill out Haddock module documentation fields out of habit, and the license field was just something that could be filled.  Other than the fact that I wrote these lines, there's no reason this file needs special treatment.

PR #1485, opened 2023-04-02
Functor on tuples

jvoigtlaender:
In the spirit of these existing rules: https://github.com/ndmitchell/hlint/blob/864ef69aa8a0741c0591297f7db2e75f69574aa6/data/hlint.yaml#L958-L998


ndmitchell:
Thanks

PR #1486, opened 2023-04-02
rename three of the hints introduced in #1481

jvoigtlaender:
Sorry, I think the name `Move concatMap out` was not so good after all. For example, it's not like in `catMaybes (concatMap f x) ==> concatMap (catMaybes . f) x` the `concatMap`-call is afterwards "outside" a `catMaybes` call.

The new names proposed now instead emphasize the function that is moving from outside into the higher-order argument of `concatMap`. A bit in the spirit of the name of this existing hint: https://github.com/ndmitchell/hlint/blob/e05030a08b238ff4d196338d01be1db37e891341/data/hlint.yaml#L122

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -876,9 +876,9 @@
     - hint: {lhs: foldr f z (concatMap g x), rhs: foldr (foldr f . g) z x}
     - warn: {lhs: fold (concatMap f x), rhs: foldMap (fold . f) x}
     - warn: {lhs: foldMap f (concatMap g x), rhs: foldMap (foldMap f . g) x}
-    - warn: {lhs: catMaybes (concatMap f x), rhs: concatMap (catMaybes . f) x, name: Move concatMap out}
-    - hint: {lhs: filter f (concatMap g x), rhs: concatMap (filter f . g) x, name: Move concatMap out}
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Also, what do you think of the following three rules?
```yaml
    - warn: {lhs: catMaybes (concat x), rhs: concatMap catMaybes x}
    - hint: {lhs: filter f (concat x), rhs: concatMap (filter f) x}
    - warn: {lhs: mapMaybe f (concat x), rhs: concatMap (mapMaybe f) x}
```
All three should be efficiency improvements (since stuff is potentially filtered away before concatenating, so there is less work to do for `++`, due to shorter lists).

jvoigtlaender:
Also possible would be:
```yaml
    - hint: {lhs: "concat (fromMaybe [] x)", rhs: "maybe [] concat x"}
    - warn: {lhs: fold (concat x), rhs: foldMap fold x}
    - warn: {lhs: foldMap f (concat x), rhs: foldMap (foldMap f) x}
```

ndmitchell:
Sorry for the delay in getting to this. The renamings look good.

As for the other hints, the three moving to use concatMap look great. The one moving to `maybe [] concat x` looks reasonable. I worry about the ones that merge concat with fold are a bit more complex? In particular `foldMap (foldMap f)` seems a relatively complex outcome?

PR #1487, opened 2023-04-02
Parenthesize negations

### LINKED ISSUE #1484: Suggest explicit brackets to mitigate counter-intuitive prefix negation precedence
# Motivation
The operator precedence of "prefix minus" can be surprising to unsuspecting Haskell programmers.

For example, in the REPL:
```
ghci> -1 ^ 2
-1
```
and
```
ghci> -5 + 3
-2
ghci> plus = (+)
ghci> -5 `plus` 3
-8
```
The case that prompted the filing of this issue was:
```
ghci> -5 `mod` 3
-2
```
which led to inferring incorrect semantics of Haskell's `mod` function, illustrated in [this thread](https://github.com/swarm-game/swarm/pull/1181#discussion_r1151156655).  As it happens, unlike some other languages, `mod a b` in Haskell is  non-negative for all positive `b`.

There is an enlightening [thread on the Haskell mailing list](https://mail.haskell.org/pipermail/haskell-prime/2010-July/003229.html) from July 2010 that discusses fixity of the prefix minus and references (defunct?) proposals to make it bind more tightly.  This corroborates the notion that the current precedence of prefix negation can be counter-intuitive.

# Proposal

In contrast with existing `hlint` rules that suggest to remove redundant brackets, I propose that `hlint` suggest to *add* redundant brackets to disambiguate (for human readers) negation precedence.

Such a rule could catch bugs; when an author writes:
```
 -5 `mod` 3
```
`hlint` would suggest instead to write:
```
 -(5 `mod` 3)
```
which would cause the author to realize that their expression is being parsed differently than intended and update their code to:
```
(-5) `mod` 3
```

# Implementation
The rule should fire whenever the operand of "unary minus" is an **infix operation** of **higher precedence**.

For reference, unary negation is fixity 6, the same as binary subtraction.  Multiplication, division, and modulus have fixity 7, and exponentiation has fixity 8.

kostmo:
Closes #1484

### DIFF
--- a/hints.md
+++ b/hints.md
@@ -1090,6 +1090,37 @@ Does not support refactoring.
 </tr>
 </table>
 
+## Builtin Negation
+
+<table>
+<tr>
+<th>Hint Name</th>
+<th>Hint</th>
+<th>Severity</th>
+</tr>
+<tr>
+<td>Parenthesize unary negation</td>
+<td>
+Example: 
+<code>
+yes = -x `mod` y
+</code>
+<br>
+Found:
+<code>
+- x `mod` y
+</code>
+<br>
+Suggestion:
+<code>
+- (x `mod` y)
+</code>
+<br>
+</td>
+<td>Suggestion</td>
+</tr>
+</table>
+
 ## Builtin NewType
 
 <table>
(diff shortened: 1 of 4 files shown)


kostmo:
Hi @ndmitchell, bump :)

kostmo:
@shayne-fletcher, would you mind reviewing this one as well?

shayne-fletcher (approved):
Thanks!

PR #1488, opened 2023-04-02
update resolver

shayne-fletcher:
update stack.yaml resolver to nightly-2023-04-02 as it has cmdargs-0.10.22 and so `stack build` and `stack run -- --test` "just work".


ndmitchell:
Thanks!

PR #1490, opened 2023-04-03
Warn when functions are unnecessarily monadic

### LINKED ISSUE #832: Warn when things are unnecessarily monadic
I occasionally see Haskell newbies write functions like `addFive x = return (x + 5)`. Is there a way we can add a hint that can catch functions that wrap their result in a `pure` or `return` but aren't actually doing anything monadic (or applicative)?

kostmo:
Closes #832

### DIFF
--- a/src/Hint/Monad.hs
+++ b/src/Hint/Monad.hs
@@ -63,6 +63,10 @@ issue978 = do \
    print "x" \
    if False then main else do \
    return ()
+
+foo x y z = return 7 -- Make foo into a pure function
+foo x y z = pure 7 -- Make foo into a pure function
+foo x y z = negate 7
 </TEST>
 -}
 
@@ -79,6 +83,7 @@ import GHC.Types.Name.Reader
 import GHC.Types.Name.Occurrence
 import GHC.Data.Bag
 import GHC.Data.Strict qualified
+import Control.Monad ( guard )
 
 import Language.Haskell.GhclibParserEx.GHC.Hs.Pat
 import Language.Haskell.GhclibParserEx.GHC.Hs.Expr
@@ -100,8 +105,11 @@ unitFuncs :: [String]
 unitFuncs = ["when","unless","void"]
 
 monadHint :: DeclHint
-monadHint _ _ d = concatMap (f Nothing Nothing) $ childrenBi d
+monadHint _ _ d =
+  baseHints <> gratuitousHints
     where
+        baseHints = concatMap (f Nothing Nothing) $ childrenBi d
+        gratuitousHints = concatMap gratuitouslyMonadic $ universeBi d
         decl = declName d
         f parentDo parentExpr x =
             monadExp decl parentDo parentExpr x ++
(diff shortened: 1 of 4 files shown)


src/Hint/Monad.hs
```diff
@@ -110,6 +134,52 @@ monadHint _ _ d = concatMap (f Nothing Nothing) $ childrenBi d
...
+      filter (not . isLetStatement) statements
+    _ -> False
+  where
+    isGratuitouslyMonadicBodyStatement statement = case statement of
+      (L _ (BodyStmt _ x _ _)) -> gratuitouslyMonadicExpr x
+      _ -> False
+    
+    isLetStatement statement = case statement of
+      (L _ (LetStmt _ _)) -> True
+      _ -> False
```
shayne-fletcher, line 179: we should probably lift this function into ghc-lib-parser-ex and reference it from there.

src/Hint/Monad.hs
```diff
@@ -110,6 +134,52 @@ monadHint _ _ d = concatMap (f Nothing Nothing) $ childrenBi d
         isHsDo (L _ HsDo{}) = True
         isHsDo _ = False
```
shayne-fletcher, line 135: and this (lift to ghc-lib-parser-ex)

shayne-fletcher (commented):
very cool. lgtm.

kostmo:
Interestingly, this new hint found two true positives in the `hlint` codebase, one of which I chose to fix (eb70b4e) and the other to ignore (d1c50ed).

googleson78:
How does this do with something that isn't monadic but uses `do`?

e.g. these examples
```haskell
{-# LANGUAGE BlockArguments
x = 
  (+)
    do 3
    do 4
    
y = do
  let z = 5
  z

f bla do
  g x y z
```

I would really like to get warnings on **none** of these, personally, because some of them are quite useful in specific cases.

googleson78:
That's awesome, thanks!

kostmo:
> How does this do with something that isn't monadic but uses `do`?

I am pleased to report that none of these examples emit a warning.  I have added them as test cases (4b8904e).  The rule is pretty conservative, and only triggers when `pure` or `return` are present.

src/Hint/Monad.hs
```diff
@@ -110,6 +147,48 @@ monadHint _ _ d = concatMap (f Nothing Nothing) $ childrenBi d
...
+-- (excluding any Let bindings) of "Body Statements" with
+-- such expressions. This catches at least some real-world
+-- sightings of the phenomenon.
+gratuitouslyMonadicExpr :: LHsExpr GhcPs -> Bool
+gratuitouslyMonadicExpr x =
+  case simplifyExp x of
+    L _ (HsApp _ (L _ (HsVar _ (L _ myFunc))) _) ->
+      occNameString (rdrNameOcc myFunc) `elem` ["pure", "return"]
+    L _ (HsDo _ _ (L _ statements)) -> all isGratuitouslyMonadicBodyStatement $
+      filter (not . isLetStatement) statements
```
shayne-fletcher, line 184: or, `filter (not . isLetStmt . unLoc)` to avoid the need for `isLetStatement`.

src/Hint/Monad.hs
```diff
@@ -110,6 +147,48 @@ monadHint _ _ d = concatMap (f Nothing Nothing) $ childrenBi d
         isHsDo (L _ HsDo{}) = True
```
shayne-fletcher, line 147: this is available from ghc-lib-parser-ex as `Language.Haskell.GhclibParserEx.GHC.Hs.Expr.isDo`

shayne-fletcher, line 147: this is available from ghc-lib-parser-ex as `Language.Haskell.GhclibParserEx.GHC.Hs.Expr.isDo`

shayne-fletcher (approved):
cc @ndmitchell

shayne-fletcher:
on 
```haskell
f x = do
  pure x
```
we get
```
/L.hs:(4,1)-(5,8): Suggestion: Unnecessarily monadic
Found:
  f x = do pure x
Perhaps:
  Demote `f` to a pure function
```
the ideal i suppose would be
```
/L.hs:(4,1)-(5,8): Suggestion: Unnecessarily monadic
Found:
  f x = do pure x
Perhaps:
  f x c = x
```
not a blocker i would suppose but is that within reach?

kostmo:
> not a blocker i would suppose but is that within reach?

The type signature of `f` would also have to change.

shayne-fletcher:
let's land this. we can always work in-situ if neil comes back with suggestions later.

ndmitchell:
Thanks for the patch, and thanks Shayne for merging. I didn't notice this, and have unfortunately not been paying attention for a bit too long 😞 . Afraid I don't think this can work as an HLint hint. Firstly, there is a bug, in that if you have:

```
foo [] = pure 1
foo (_:_) = print 1
```

This reports that `foo` is unnecessarily monadic. But it's not - you have to consider all equations at once.

The other issue is that there isn't a good "fix" for this, as doing so necessarily changes the signature, and thus you have to update all callers. No other HLint hints have that property, and I'm not sure if it's something we want to start or not.

I guess we could have it as an off by default hint, but it still does things that HLint otherwise doesn't ever do, so I'm not sure it fits. Sorry for the effort you two invested in this.

PR #1491, opened 2023-04-03
make stack run -- src pass again

shayne-fletcher:
https://github.com/ndmitchell/hlint/pull/1481 introduced new hints for `concatMap` expressions. this brings hlint/src into conformance.


ndmitchell:
Thanks!

PR #1495, opened 2023-04-05
Update stack resolver to nightly-2023-04-05.

chungyc:
The lower bound for the cmdargs package is 0.10.22, but the set of packages for nightly-2023-03-12 is not quite recent enough for it to be included. (The version of cmdargs is 0.10.21 in nightly-2023-03-12.)


shayne-fletcher:
this is a duplicate of https://github.com/ndmitchell/hlint/pull/1488.

i'm sure you know but in case not, while it hasn't landed you can `stack build --resolver nightly-2023-04-02` to work around

PR #1498, opened 2023-04-21
[Patch+test] Fix bug with postqualified imports and qualifiedStyle=unrestricted

### LINKED ISSUE #1499: Incorrect hint for postpositive qualified syntax
The following rule:
```
# testcase-rule.yaml

- modules:
  - name: Data.Text
    importStyle: qualified
    qualifiedStyle: unrestricted
    asRequired: True
    as: T
```
when applied to the single-line file:
```
import Data.Text qualified as T
```
emits, erroneously, a silly hint:
```
> hlint -h testcase-rule.yaml test.hs
test.hs:1:1-31: Warning: Data.Text should be imported qualified
Found:
  import Data.Text qualified as T
Perhaps:
  import qualified Data.Text as T
Note: may break the code

1 hint
```
#1498 fixes this.

EDIT: For context see [section 6.3.5 of the GHC User Guide](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/import_qualified_post.html#writing-qualified-in-postpositive-position), "Writing qualified in postpositive position". It is enabled by the GHC extension `ImportQualifiedPost` since GHC 8.10.1. It allows writing multiple imports with sane indentation.

ulidtko:
Hi!

I'll be brief. The bug:

```
# testcase-rule.yaml

- modules:
  - name: Data.Text
    importStyle: qualified
    qualifiedStyle: unrestricted
    asRequired: True
    as: T
```
on the single-line file:
```haskell
import Data.Text qualified as T
```

emits, erroneously, a silly hint:
```
> hlint -h testcase-rule.yaml test.hs
test.hs:1:1-31: Warning: Data.Text should be imported qualified
Found:
  import Data.Text qualified as T
Perhaps:
  import qualified Data.Text as T
Note: may break the code

1 hint
```

I submit a direct testcase for the bug, then a fix, then more tests for better regression coverage.

Hoping for easy review & merge; release isn't urgent, batch up at will :raised_hands: Thanks in advance

Fixes #1499

### DIFF
--- a/src/Hint/Restrict.hs
+++ b/src/Hint/Restrict.hs
@@ -31,6 +31,7 @@ import Data.Set qualified as Set
 import Data.Map qualified as Map
 import Data.List.Extra
 import Data.List.NonEmpty (nonEmpty)
+import Data.Either
 import Data.Maybe
 import Data.Monoid
 import Data.Semigroup
@@ -157,6 +158,11 @@ checkPragmas modu flags exts mps =
      , not $ null bad]
    isGood def mp x = maybe def (within modu "" . riWithin) $ Map.lookup x mp
 
+
+-- | Extension to GHC's 'ImportDeclQualifiedStyle', expressing @qualifiedStyle: unrestricted@,
+-- i.e. the preference of "either pre- or post-, but qualified" in a rule.
+data QualifiedPostOrPre = QualifiedPostOrPre deriving Eq
+
 checkImports :: String -> [LImportDecl GhcPs] -> (Bool, Map.Map String RestrictItem) -> [Idea]
 checkImports modu lImportDecls (def, mp) = mapMaybe getImportHint lImportDecls
   where
@@ -190,30 +196,31 @@ checkImports modu lImportDecls (def, mp) = mapMaybe getImportHint lImportDecls
               case fromMaybe ImportStyleUnrestricted $ getAlt riImportStyle of
                 ImportStyleUnrestricted
                   | NotQualified <- ideclQualified -> (Nothing, Nothing)
-                  | otherwise -> (second (<> " or unqualified") <$> expectedQualStyle, Nothing)
-                ImportStyleQualified -> (expectedQualStyleDef, Nothing)
+                  | otherwise -> (Just $ second (<> " or unqualified") expectedQualStyle, Nothing)
+                ImportStyleQualified -> (Just expectedQualStyle, Nothing)
                 ImportStyleExplicitOrQualified
                   | Just (False, _) <- first (== EverythingBut) <$> ideclImportList -> (Nothing, Nothing)
                   | otherwise ->
-                      ( second (<> " or with an explicit import list") <$> expectedQualStyleDef
+                      ( Just $ second (<> " or with an explicit import list") expectedQualStyle
                       , Nothing )
                 ImportStyleExplicit
                   | Just (False, _) <- first (== EverythingBut) <$> ideclImportList -> (Nothing, Nothing)
                   | otherwise ->
-                      ( Just (NotQualified, "unqualified")
+                      ( Just (Right NotQualified, "unqualified")
                       , Just $ Just (Exactly, noLocA []) )
-                ImportStyleUnqualified -> (Just (NotQualified, "unqualified"), Nothing)
-            expectedQualStyleDef = expectedQualStyle <|> Just (QualifiedPre, "qualified")
+                ImportStyleUnqualified -> (Just (Right NotQualified, "unqualified"), Nothing)
             expectedQualStyle =
               case fromMaybe QualifiedStyleUnrestricted $ getAlt riQualifiedStyle of
-                QualifiedStyleUnrestricted -> Nothing
-                QualifiedStylePost -> Just (QualifiedPost, "post-qualified")
-                QualifiedStylePre -> Just (QualifiedPre, "pre-qualified")
+                QualifiedStyleUnrestricted -> (Left QualifiedPostOrPre, "qualified")
+                QualifiedStylePost -> (Right QualifiedPost, "post-qualified")
+                QualifiedStylePre -> (Right QualifiedPre, "pre-qualified")
             qualIdea
-              | Just ideclQualified == (fst <$> expectedQual) = Nothing
+              | Just (Right ideclQualified) == (fst <$> expectedQual) = Nothing
+              | Just (Left QualifiedPostOrPre) == (fst <$> expectedQual)
+                && ideclQualified `elem` [QualifiedPost, QualifiedPre] = Nothing
               | otherwise = expectedQual
         whenJust qualIdea $ \(qual, hint) -> do
-          let i' = noLoc $ (unLoc i){ ideclQualified = qual
+          let i' = noLoc $ (unLoc i){ ideclQualified = fromRight QualifiedPre qual
                                     , ideclImportList = fromMaybe ideclImportList expectedHiding }
               msg = moduleNameString (unLoc ideclName) <> " should be imported " <> hint
           Left $ warn msg (reLoc i) i' []
--- a/src/Hint/Restrict.hs
+++ b/src/Hint/Restrict.hs
@@ -214,12 +214,18 @@ checkImports modu lImportDecls (def, mp) = mapMaybe getImportHint lImportDecls
                 QualifiedStyleUnrestricted -> (Left QualifiedPostOrPre, "qualified")
                 QualifiedStylePost -> (Right QualifiedPost, "post-qualified")
                 QualifiedStylePre -> (Right QualifiedPre, "pre-qualified")
+            -- unless expectedQual is Nothing, it holds the Idea (hint) to ultimately emit,
+            -- except in these cases when the rule's requirements are fulfilled in-source:
             qualIdea
+              -- the rule demands a particular importStyle, and the decl obeys exactly
               | Just (Right ideclQualified) == (fst <$> expectedQual) = Nothing
+              -- the rule demands a QualifiedPostOrPre import, and the decl does either
               | Just (Left QualifiedPostOrPre) == (fst <$> expectedQual)
                 && ideclQualified `elem` [QualifiedPost, QualifiedPre] = Nothing
+              -- otherwise, expectedQual gets converted into a warning below (or is Nothing)
               | otherwise = expectedQual
         whenJust qualIdea $ \(qual, hint) -> do
+          -- convert non-Nothing qualIdea into hlint's refactoring Idea
           let i' = noLoc $ (unLoc i){ ideclQualified = fromRight QualifiedPre qual
                                     , ideclImportList = fromMaybe ideclImportList expectedHiding }
               msg = moduleNameString (unLoc ideclName) <> " should be imported " <> hint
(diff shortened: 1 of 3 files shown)


src/Hint/Restrict.hs
```diff
@@ -190,30 +196,31 @@ checkImports modu lImportDecls (def, mp) = mapMaybe getImportHint lImportDecls
...
+                      ( Just $ second (<> " or with an explicit import list") expectedQualStyle
                       , Nothing )
                 ImportStyleExplicit
                   | Just (False, _) <- first (== EverythingBut) <$> ideclImportList -> (Nothing, Nothing)
                   | otherwise ->
-                      ( Just (NotQualified, "unqualified")
+                      ( Just (Right NotQualified, "unqualified")
                       , Just $ Just (Exactly, noLocA []) )
-                ImportStyleUnqualified -> (Just (NotQualified, "unqualified"), Nothing)
-            expectedQualStyleDef = expectedQualStyle <|> Just (QualifiedPre, "qualified")
```
ulidtko, line 206: My ID of the root cause is this line :eyes:

Right here, `QualifiedStyleUnrestricted` (no pre-/post- preference) was getting mapped into `QualifiedPre`.

ulidtko, line 206: My ID of the root cause is this line :eyes:

Right here, `QualifiedStyleUnrestricted` (no pre-/post- preference) was getting mapped into `QualifiedPre`.

src/Hint/Restrict.hs
```diff
@@ -190,30 +196,31 @@ checkImports modu lImportDecls (def, mp) = mapMaybe getImportHint lImportDecls
...
-                QualifiedStylePost -> Just (QualifiedPost, "post-qualified")
-                QualifiedStylePre -> Just (QualifiedPre, "pre-qualified")
+                QualifiedStyleUnrestricted -> (Left QualifiedPostOrPre, "qualified")
+                QualifiedStylePost -> (Right QualifiedPost, "post-qualified")
+                QualifiedStylePre -> (Right QualifiedPre, "pre-qualified")
             qualIdea
-              | Just ideclQualified == (fst <$> expectedQual) = Nothing
+              | Just (Right ideclQualified) == (fst <$> expectedQual) = Nothing
+              | Just (Left QualifiedPostOrPre) == (fst <$> expectedQual)
+                && ideclQualified `elem` [QualifiedPost, QualifiedPre] = Nothing
```
ulidtko, line 220: This new match arm handles the `QualifiedStyleUnrestricted` case now.

With `QualifiedPostOrPre` carrying the "either one is fine" intent, we emit the `qualIdea` only when `ideclQualified` is `NotQualified` — accepting either style.

src/Hint/Restrict.hs
```diff
@@ -157,6 +158,11 @@ checkPragmas modu flags exts mps =
      , not $ null bad]
    isGood def mp x = maybe def (within modu "" . riWithin) $ Map.lookup x mp
 
+
+-- | Extension to GHC's 'ImportDeclQualifiedStyle', expressing @qualifiedStyle: unrestricted@,
+-- i.e. the preference of "either pre- or post-, but qualified" in a rule.
+data QualifiedPostOrPre = QualifiedPostOrPre deriving Eq
```
ulidtko, line 164: I chose a singleton datatype instead of another Maybe layer, basically to have `Either QualifiedPostOrPre ImportDeclQualifiedStyle` sum type with values for:
 * QualifiedPostOrPre,
 * QualifiedPre, 
 * QualifiedPost,
 * NotQualified.

`checkImports` already uses 2 or 3 layers of Maybe, and I think it was getting confusing.

src/Hint/Restrict.hs
```diff
@@ -190,30 +196,31 @@ checkImports modu lImportDecls (def, mp) = mapMaybe getImportHint lImportDecls
...
+                QualifiedStylePre -> (Right QualifiedPre, "pre-qualified")
             qualIdea
-              | Just ideclQualified == (fst <$> expectedQual) = Nothing
+              | Just (Right ideclQualified) == (fst <$> expectedQual) = Nothing
+              | Just (Left QualifiedPostOrPre) == (fst <$> expectedQual)
+                && ideclQualified `elem` [QualifiedPost, QualifiedPre] = Nothing
               | otherwise = expectedQual
         whenJust qualIdea $ \(qual, hint) -> do
-          let i' = noLoc $ (unLoc i){ ideclQualified = qual
+          let i' = noLoc $ (unLoc i){ ideclQualified = fromRight QualifiedPre qual
```
ulidtko, line 223: The `fromRight QualifiedPre` conservatively falls back to suggesting pre-qualified import in cases like this:

```
tests/importStyle-1.hs:2:1-26: Warning: HypotheticalModule2 should be imported qualified or with an explicit import list
Found:
  import HypotheticalModule2
Perhaps:
  import qualified HypotheticalModule2
```

This avoids breaking existing tests with `qualifiedStyle: unrestricted`.

ulidtko:
Fixes #1499.

ulidtko:
Hey @zliu41 @ndmitchell — I noticed somebody has run CI on the PR; thanks.

I've checked all 4 workflows — none of them failed because of changes in the PR. Timeout; 143 oomkill; exitcode 251 (?) when building yaml haddock on Windows; build oomkilled on MacOS too.

**Tests do pass** :heavy_check_mark: on this branch; I explicitly centered the bugfix around tests. See PR diff.

-------------------

As a sidenote, the CI setup could use a brush-up. Misuse of `actions/cache` caught my eye right away — when used properly, it will dramatically speed up builds, reduce OOMs, and that without build integrity issues (from stale cache), when done properly.

ulidtko:
@zliu41 @ndmitchell ping. Please review.

googleson78:
Also hitting this same problem

src/Hint/Restrict.hs
googleson78: Seems like it might even be better to make a new data type like `data ImportStyle = PostOrPre | ??? ImportDeclQualifiedStyle` if you could figure out a good name for `???`

src/Hint/Restrict.hs
```diff
@@ -190,30 +196,31 @@ checkImports modu lImportDecls (def, mp) = mapMaybe getImportHint lImportDecls
...
               case fromMaybe QualifiedStyleUnrestricted $ getAlt riQualifiedStyle of
-                QualifiedStyleUnrestricted -> Nothing
-                QualifiedStylePost -> Just (QualifiedPost, "post-qualified")
-                QualifiedStylePre -> Just (QualifiedPre, "pre-qualified")
+                QualifiedStyleUnrestricted -> (Left QualifiedPostOrPre, "qualified")
+                QualifiedStylePost -> (Right QualifiedPost, "post-qualified")
+                QualifiedStylePre -> (Right QualifiedPre, "pre-qualified")
             qualIdea
-              | Just ideclQualified == (fst <$> expectedQual) = Nothing
+              | Just (Right ideclQualified) == (fst <$> expectedQual) = Nothing
```
zliu41, line 218: The logic of this function seems quite convoluted. Can you add some comments in this function, for example, what does it mean when `expectedQual` is a `Just` or a `Nothing`? `expectedQual` doesn't seem to represent what its name suggests - the expected qualifying style, because the code for obtaining `expectedQual` also checks the actual qualifying style from `ideclQualified`.

src/Hint/Restrict.hs
zliu41: > With QualifiedPostOrPre carrying the "either one is fine" intent, we emit the qualIdea only when ideclQualified is NotQualified — accepting either style.

Please add the comment in the code.

zliu41 (commented):
I'm inclined to remove the "should be imported post-qualified" hint, because there are tools available (e.g., stylish Haskell) for converting pre-qualified imports to post-qualified imports (or vice versa). If one cares about using pre-qualified or post-qualified imports consistently in their project, they can simply run the tool in the pre-commit hook.

That said, we can still merge this. @ulidtko the CI has been fixed since this PR was opened, can you rebase?

src/Hint/Restrict.hs
ulidtko: Indeed.

I tried to go with a minimal fix, rather then trying to rework the whole function. Focused patches are typically easier to review.

`expectedQual` gets determined by this beefy case-match (which is easy to see, just lexically):

https://github.com/ndmitchell/hlint/blob/0343ec5d68667ce21188cb7871e7ec0d5ac79671/src/Hint/Restrict.hs#L195-L211

It handles the rule's variations in `importStyle` (together with nuances around `asRequired`, `qualifiedStyle`, hiding lists, etc). It also looks into the declaration-under-check indeed, the `ideclQualified` — AFAICS, just to spare having a second case-match over `riImportStyle`.

TL;DR: if `expectedQual` is Just something, it will emit a hint... unless silenced by the 2 checks in `qualIdea`.

I'll document this in-source.

ulidtko (commented):
Hi @zliu41, thanks for review.

> I'm inclined to remove the "should be imported post-qualified" hint [...]

No, please don't. In a large enough codebase, a "stop the world" PR with mass-rewrite of most imports has negligible chances to ever land, due to pervasive conflicts.

That's why an incremental approach can be preferred, based on hlint rules about individual modules.

That's also why `qualifiedStyle=unrestricted` is an important feature: it enables smooth nondisruptive incremental transitions. That can be the only way in some conditions (e.g. big repo with moderate code-churn velocity and strict hlint-compliance) to get `PostQualifiedImports`.

> @ulidtko the CI has been fixed since this PR was opened, can you rebase?

Sure. Done.

I'll going to add clarifying comments in next commit.

shayne-fletcher:
> I'm inclined to remove the "should be imported post-qualified" hint, because there are tools available (e.g., stylish Haskell) for converting pre-qualified imports to post-qualified imports (or vice versa). If one cares about using pre-qualified or post-qualified imports consistently in their project, they can simply run the tool in the pre-commit hook.

Note also that the flag [`-Wprepositive-qualified-module`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-warnings.html#ghc-flag--Wprepositive-qualified-module) (off by default) will warn on any prepositive `qualified` annotations.

ulidtko:
> Note also that the flag [`-Wprepositive-qualified-module`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-warnings.html#ghc-flag--Wprepositive-qualified-module) (off by default) will warn on any prepositive `qualified` annotations.

Good point. With `{-# OPTIONS_GHC -Wprepositive-qualified-module #-}` — is possible to migrate module-by-module (as opposed to, across entire project at once).

To explain a motivating usecase further; my point to keep (and fix) `qualifiedStyle=unrestricted` is with rules like this:
```yaml
# (project-local) .hlint.yaml

- modules:
  # ...
  - name: Core.Effects,
    importStyle: explicitOrQualified,
    asRequired: True,
    as: Effects,
```
— which encodes the intent to allow e.g. `import Core.Effects (WithMetrics)` as well as `import qualified Core.Effects as Effects`... but, forbids `import Core.Effects qualified as Effects`, completely by accident.

zliu41 (approved):
Looks good. CI is somehow broken again, so I'll just merge it.

> No, please don't. In a large enough codebase, a "stop the world" PR with mass-rewrite of most imports has negligible chances to ever land, due to pervasive conflicts.

A pre-commit hook can be made to run a program (hlint or a code formatter) only on the changed or added files. Alternatively, one can run a program (hlint or a code formatter) only on selected files, which allows them to do the migration piece by piece.

So the real question is, whether something is a formatting issue and can be taken care of by a code formatter, or a code smell that can only be detected by hlint.

PR #1500, opened 2023-04-24
add hlint rule for NonEmpty.unzip

mixphix:
For CLC ticket [#86](https://github.com/haskell/core-libraries-committee/issues/86).

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -1122,6 +1122,7 @@
     - warn: {lhs: "[] /= x", rhs: not (null x), name: Use null}
     - warn: {lhs: "\"\" /= x", rhs: not (null x), name: Use null}
     - warn: {lhs: "maybe []", rhs: foldMap}
+    - warn: {lhs: "Data.List.NonEmpty.unzip", rhs: "Data.Functor.unzip", name: "Avoid NonEmpty.unzip", note: "Because the function is being deprecated"}
 
 - group:
     name: generalise-for-conciseness


mixphix:
I wasn't sure if this was the best category to put it in. Let me know!

ndmitchell:
Thanks for the hint - this should go in the base group, as we want it on by default.

ndmitchell:
I'll move it

ndmitchell:
You are of course right this is a generalisation hint, and if we weren't deprecating NonEmpty.unzip it would be exactly the right place for it to go. But given it is, we want it on by default. Moved in 6f84e5ce9efee39c790b40578ba006552c336340

snowleopard:
I just had to silence this warning because it was firing [here](https://github.com/snowleopard/alga/blob/aee5fa79a1ce53861c9a9bb770309dbf68a1bc9a/src/Algebra/Graph/NonEmpty/AdjacencyMap.hs#L181). I'm not sure why it fires there: as far as I can tell, this `unzip` refers to the monomorphic one for lists, not to `Data.List.NonEmpty.unzip` as the HLint rule says. Maybe HLint can't really tell where that `unzip` is coming from? But then maybe the suggestion should be worded differently.

Note that I have `import Data.List.NonEmpty (NonEmpty (..), nonEmpty, toList, reverse)` at the top of the module, so I shouldn't have `Data.List.NonEmpty.unzip` in scope.

Not a big deal (I just disabled the warning) but something felt fishy, hence my report.

PR #1507, opened 2023-05-08
Fix CI

zliu41:
I'm not sure what the `ndmitchell/neil@master` step does but it is consistently failing, so I changed it to just running the hlint tests.


zliu41:
Several people are working on HLint here at ZuriHac, so I'm going to merge this so that CI works for them.
@ndmitchell if you disagree with this I can address it later.

ndmitchell:
Thanks @zliu41 - good call! Will try putting it back now, and make sure to fix it.

PR #1516, opened 2023-06-11
Removed deprecated call for simulated ANSI.

elland:
hlint called  `hSupportsANSIWithoutEmulation`, which has been deprecated. The ansi-terminal package no longer supports legacy versions of windows that required emulation.


eyeinsky:
Just replace the entire do block with `hSupportsANSI stdout` :)

elland:
@eyeinsky I ended up picking up an issue right after that changes the exact same line https://github.com/ndmitchell/hlint/pull/1517 so any changes here will conflict there. I'd take this PR as more of a check if this is a fine change to make and just apply the change to the other PR instead.

zliu41:
Thanks!

PR #1518, opened 2023-06-11
Add ignore-suggestions flag

### LINKED ISSUE #1430: Support hiding warning levels on command line
Can't find an existing issue for this (or existing functionality); feel free to point to existing issues.

It would be great to be able to hide all suggestions when running hlint in a precommit hook (i.e. warnings are displayed + return nonzero exit code, suggestions not displayed and dont change exit code). hlint.yaml should not change so that IDE plugins could still show suggestions.

We don't want suggestions displayed at all in the pre-commit hook (not just not failing the hook) so that warnings dont get hidden underneath all the non-failing suggestions

batkot:
Closes #1430 

Not sure how it should interact with [`--show` flag](https://github.com/ndmitchell/hlint/blob/56f4d63f328a8512de2498d5e34c719f3db59678/src/CmdLine.hs#L106). My assumption was that `--show` takes precedence and `hlint` reports all ideas (it's called showAll after all).

### DIFF
--- a/src/CmdLine.hs
+++ b/src/CmdLine.hs
@@ -147,7 +147,7 @@ mode = cmdArgsMode $ modes
         ,cmdThreads = 1 &= name "threads" &= name "j" &= opt (0 :: Int) &= help "Number of threads to use (-j for all)"
         ,cmdIgnore = nam "ignore" &= typ "HINT" &= help "Ignore a particular hint"
         ,cmdShowAll = nam "show" &= help "Show all ignored ideas"
-        ,cmdIgnoreSuggestions = nam_ "ignore-suggestions" &= help "Ignore suggestions"
+        ,cmdIgnoreSuggestions = nam_ "ignore-suggestions" &= help "Ignore suggestions, only show warnings and errors"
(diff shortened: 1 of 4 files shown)


batkot:
> That sounds fine to me.

Added a test to cover that case :+1:

src/CmdLine.hs
```diff
@@ -146,6 +147,7 @@ mode = cmdArgsMode $ modes
         ,cmdThreads = 1 &= name "threads" &= name "j" &= opt (0 :: Int) &= help "Number of threads to use (-j for all)"
         ,cmdIgnore = nam "ignore" &= typ "HINT" &= help "Ignore a particular hint"
         ,cmdShowAll = nam "show" &= help "Show all ignored ideas"
+        ,cmdIgnoreSuggestions = nam_ "ignore-suggestions" &= help "Ignore suggestions"
```
zliu41, line 150:
```suggestion
        ,cmdIgnoreSuggestions = nam_ "ignore-suggestions" &= help "Ignore suggestions, only show warnings and errors"
```

zliu41 (commented):
Looks great, one minor suggestion.

zliu41:
Thanks!

Issue #1501, opened 2023-04-25
incorrect "redundant bracket" hint


MangoIV:
HLint hints that I should remove the parens here: 
```haskell
.. \(Proxy @a) -> ...
```
this would be a parsing error.

zliu41:
Fixed in https://github.com/ndmitchell/hlint/pull/1519

PR #1519, opened 2023-06-11
Hint to remove parentheses that would result in a parse error

elland:
Addresses https://github.com/ndmitchell/hlint/issues/1501

### DIFF
--- a/src/Hint/Bracket.hs
+++ b/src/Hint/Bracket.hs
@@ -121,8 +121,9 @@ bar' :: ((A a) => ()) -> ()
 data Dict c where Dict :: (c) => Dict c
 data Dict' c a where Dict' :: (c a) => Dict' c a
 
-issue1501: Redundant bracket hint resulted in a parse error
+-- issue1501: Redundant bracket hint resulted in a parse error
 \(Proxy @a) -> True
+x = f $ \(Proxy @a) -> True
 </TEST>
 -}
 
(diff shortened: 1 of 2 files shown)


src/Hint/Bracket.hs
```diff
@@ -120,6 +120,9 @@ foo' :: ((A) => ()) -> ()
 bar' :: ((A a) => ()) -> ()
 data Dict c where Dict :: (c) => Dict c
 data Dict' c a where Dict' :: (c a) => Dict' c a
+
+issue1501: Redundant bracket hint resulted in a parse error
```
zliu41, line 124:
```suggestion
-- issue1501: Redundant bracket hint resulted in a parse error
```

src/Hint/Bracket.hs
```diff
@@ -120,6 +120,9 @@ foo' :: ((A) => ()) -> ()
...
+\(Proxy @a) -> True
```
zliu41, line 125: You may need a binding, as in `x = \(Proxy @a) -> True` - I'm not sure.

zliu41: Actually this would cause HLint to suggest `x (Proxy @a) = True`. You could use `x = f $ \(Proxy @a) -> True`.

elland: Do you mean as an extra case or instead of the one already there? :)

zliu41: I meant instead of. This is not needed given the one below.

zliu41 (approved):
Thanks!

PR #1520, opened 2023-06-11
Remove incorrect foldr/concat rules (#1511)

### LINKED ISSUE #1511: Incorrect foldr hint
## I meant this suggestion

```
<td>Use foldr</td>
<td>
LHS:
<code>
foldr f z (concat x)
</code>
<br>
RHS:
<code>
foldr (foldr f) z x
</code>
```

## Here is my reasoning

- Assume 
    - `foldr` has type `(a -> b -> b) -> b -> [a] -> b` for the sake of simpler reasoning.
    - `concat :: [[a]] -> [a]`
- Given this code `foldr f z (concat x)`, if it compile, we got
    - `x :: [[a]]`
    - `z :: b`
    - `f :: a -> b -> b`
- Assume the RHS `foldr (foldr f) z x` is correct,
    - `foldr f` has to have type `[a] -> b -> b` given `x :: [[a]]` from previous step.
    - but actually `foldr f :: b -> [a] -> b` given `f :: a -> b -> b` from previous step.
- So the RHS could compile if write in this
    - `foldr (flip $ foldr f) z x`
    - However I dont see this is simpler than the LHS.

## Which version report this hint?

- the latest master build but not `v3.5`

gruhn:
Closes #1511

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -876,8 +876,6 @@
     - warn: {lhs: any f (concatMap g x), rhs: any (any f . g) x}
     - warn: {lhs: all f (concat x), rhs: all (all f) x}
     - warn: {lhs: all f (concatMap g x), rhs: all (all f . g) x}
-    - hint: {lhs: foldr f z (concat x), rhs: foldr (foldr f) z x}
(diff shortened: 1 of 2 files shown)


zliu41 (approved):
Thanks!

PR #1521, opened 2023-06-11
Dont suggest const when NamedFieldPuns are used

### LINKED ISSUE #1453: Invalid hint when using NamedFieldPun field updates
When trying out modernizing Haskell I've changed [some code](https://github.com/Peaker/git-mediate/blob/a991d32a5922cc7ac1821db84e69a642115f022e/src/Conflict.hs#L36) to:

```Haskell
-- traversal
bodies :: Applicative f => (Sides [String] -> f (Sides [String])) -> Conflict -> f Conflict
bodies f c = (\cBodies -> c{cBodies}) <$> f c.cBodies
```

To which hlint gives a wrong hint:

```Haskell
Use const
Found:
  \ cBodies -> c {cBodies}
Why not:
  const c {cBodies}
```

Note that `\cBodies -> c {cBodies}` with `NamedFieldPun` is equivalent to `\cBodies -> c {cBodies=cBodies}` so it does use the lambda parameter (not a const).

batkot:
Closes #1453

### DIFF
--- a/src/Hint/Match.hs
+++ b/src/Hint/Match.hs
@@ -201,7 +201,7 @@ checkSide x bind = maybe True bool x
       isType "Compare" x = True -- Just a hint for proof stuff
       isType "Atom" x = isAtom x
       isType "WHNF" x = isWHNF x
-      isType "Wildcard" x = any isFieldPun (universeBi x) || any hasFieldsDotDot (universeBi x)
+      isType "Wildcard" x = any hasFieldsDotDot (universeBi x)
       isType "Nat" (asInt -> Just x) | x >= 0 = True
       isType "Pos" (asInt -> Just x) | x >  0 = True
       isType "Neg" (asInt -> Just x) | x <  0 = True
--- a/src/Hint/Match.hs
+++ b/src/Hint/Match.hs
@@ -201,7 +201,7 @@ checkSide x bind = maybe True bool x
       isType "Compare" x = True -- Just a hint for proof stuff
       isType "Atom" x = isAtom x
       isType "WHNF" x = isWHNF x
-      isType "Wildcard" x = any isFieldPun (universeBi x) || any hasFieldsDotDot (universeBi x) || any isFieldPunUpdate (universeBi x)
+      isType "Wildcard" x = any isFieldPun (universeBi x) || any hasFieldsDotDot (universeBi x)
       isType "Nat" (asInt -> Just x) | x >= 0 = True
       isType "Pos" (asInt -> Just x) | x >  0 = True
       isType "Neg" (asInt -> Just x) | x <  0 = True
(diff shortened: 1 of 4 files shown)


src/Hint/Match.hs
```diff
@@ -201,7 +201,7 @@ checkSide x bind = maybe True bool x
       isType "Compare" x = True -- Just a hint for proof stuff
       isType "Atom" x = isAtom x
       isType "WHNF" x = isWHNF x
-      isType "Wildcard" x = any isFieldPun (universeBi x) || any hasFieldsDotDot (universeBi x)
+      isType "Wildcard" x = any isFieldPun (universeBi x) || any hasFieldsDotDot (universeBi x) || any isFieldPunUpdate (universeBi x)
```
zliu41, line 204: Nice attempt! But I don't think either `any isFieldPun (universeBi x)` or `any isFieldPunUpdate (universeBi x)` is the right thing to do.

The goal is to avoid suggesting "Use const" for either `\a -> Foo{a}` or `\a -> x{a}`. The first one is field pun, and the second one is field pun update. But there's no need to exclude all field puns or field pun updates. We should indeed suggest "Use const" for `\a -> Foo{b}` and `\a -> x{b}`.

I think the correct approach is to reject `\a -> Foo{a}` and `\a -> x{a}` on the basis of introducing free variables: if we turn `\a -> Foo{a}` into `const Foo{a}`, then we'd introduce a new free variable `a`, and that's why we shouldn't do it. The check is [here](https://github.com/ndmitchell/hlint/blob/291caebde337368d6299dec18dc388a59edba26a/src/Hint/Match.hs#L153).

This check does the right thing for field puns, but not for field pun updates, because the [freeVar computation for field pun updates](https://github.com/ndmitchell/hlint/blob/291caebde337368d6299dec18dc388a59edba26a/src/GHC/Util/FreeVars.hs#L176-L177) is incorrect. When `hfbPun==True`, instead of returning `freeVars x`, it should return the OccName contained in the LHS (the argument before `x`) as the free variable. Hope this helps!

tests/hint.test
```diff
@@ -51,6 +51,29 @@ Perhaps:
 
 1 hint
 
+---------------------------------------------------------------------
+RUN tests/use-const.hs
```
zliu41, line 55: These tests can be written directly in the `<TEST>` section in hlint.yaml. The `.test` files are for tests that need custom hint files, which can't be written there.

zliu41 (changes requested):

src/Hint/Match.hs
batkot: Thanks for directions! Hope I got it right.

> Nice attempt! But I don't think either any isFieldPun (universeBi x) or any isFieldPunUpdate (universeBi x) is the right thing to do.

Should I take the opportunity then and get rid of any isFieldPun (universeBi x) from Wildcard? It doesn't break any tests, but I don't really understand what Wildcard is supposed to check sweat_smile

tests/hint.test
batkot: Thanks! I didn't notice `<TEST>` section in `hlint.yaml` :sweat_smile:

zliu41 (approved):
:+1:

PR #1522, opened 2023-06-12
define __GLASGOW_HASKELL__ for cmdCpp

### LINKED ISSUE #476: Define __GLASGOW_HASKELL__
When doing CPP I should define __GLASGOW_HASKELL__ to the version HLint was compiled with (an approximation for what suits the user)

juhp:
~~We define it in EmbeddData since we can't do the CPP expansion in CmdLine~~

This at least compiles though I haven't explicitly tested that it works.

Fixes #476

### DIFF
--- a/tests/cpp.test
+++ b/tests/cpp.test
@@ -129,3 +129,39 @@ import Network.Wai.Internal
 #endif
 OUTPUT
 No hints
+
+---------------------------------------------------------------------
+RUN tests/cpp-file5.hs
+FILE tests/cpp-file5.hs
+{-# LANGUAGE CPP #-}
+main =
+#if __GLASGOW_HASKELL__ > 800
+  print __GLASGOW_HASKELL__
+#else
+  putStrLn $ show __GLASGOW_HASKELL__
+#endif
+OUTPUT
+tests/cpp-file5.hs:1:1-20: Warning: Avoid restricted extensions
+Found:
+  {-# LANGUAGE CPP #-}
+Note: may break the code
+
+1 hint
+
+---------------------------------------------------------------------
+RUN --cpp-define "__GLASGOW_HASKELL__=123" tests/cpp-file6.hs
+FILE tests/cpp-file6.hs
+{-# LANGUAGE CPP #-}
+main =
+#if __GLASGOW_HASKELL__ == 123
+  print __GLASGOW_HASKELL__
+#else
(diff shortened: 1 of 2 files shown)


src/EmbedData.hs
```diff
@@ -23,3 +24,6 @@ defaultYaml = toString $(embedFile "data/default.yaml")
 {-# NOINLINE reportTemplate #-}
 reportTemplate :: String
 reportTemplate = toString $(embedFile "data/report_template.html")
+
+ghcVersion :: String
+ghcVersion = "__GLASGOW_HASKELL__"
```
zliu41, line 29: I think it should be

```suggestion
ghcVersion = show __GLASGOW_HASKELL__
```

zliu41 (changes requested):
Please add a test case in `cpp.test`, say
```
#if __GLASGOW_HASKELL >= 800
...
#else
...
```

and verify that the first branch is taken by HLint.

juhp:
Thanks for reviewing - I added a simple testcase

zliu41 (approved):
Thanks, looks great!

juhp:
Hmm, the test seems to break CI where hlint gives an output warning/hint about the CPP extension

juhp:
Hmm ok I tweaked the testcase, though I don't know why the CI was giving 2 hints.

juhp:
(Also I don't really grok the CPP hint - I mean I get it that CPP is "unsafe"...)

tests/cpp.test
```diff
@@ -129,3 +129,21 @@ import Network.Wai.Internal
 #endif
 OUTPUT
 No hints
+
+---------------------------------------------------------------------
+RUN --cpp-simple tests/cpp-file5.hs
```
zliu41, line 134: Remove `--cpp-simple`. Otherwise, it runs a simple CPP where lines starting with "#" are simply removed, so it is actually running HLint on 
```
main = 
  print __GLASGOW_HASKELL__
  putStrLn $ show __GLASGOW_HASKELL__
```

zliu41 (changes requested):
> Warning: Avoid restricted extensions

I think it's because the tests by default use `.hlint.yaml` as the hint file, which specifies CPP as a restricted extension.

zliu41 (approved):
Looks great, thanks.

Issue #704, opened 2019-07-05
Parse errors are listed as hints


recursion-ninja:
When I run `hlint` on our code base there are 3 parse errors and two hints, however the textual report says "5 hints".

```
hlint --no-exit-code app test ./lib/pcg-evaluation ./lib/pcg-data-unification ./lib/pcg-error-phase ./lib/pcg-language ./lib/pcg-file-parsers ./lib/pcg-file-source ./lib/pcg-data-normalization ./lib/pcg-alphabet ./lib/pcg-utility ./lib/pcg-serialize ./lib/pcg-tcm-memo ./lib/pcg-core ./lib/pcg-validation-transformer ./lib/pcg-exportable 
./lib/pcg-file-parsers/src/File/Format/Fasta/Parser.hs:120:34: Warning: Prefer fold to mconcat
Found:
  mconcat
Perhaps:
  fold

./lib/pcg-file-parsers/src/File/Format/Fasta/Parser.hs:132:20: Warning: Prefer fold to mconcat
Found:
  mconcat
Perhaps:
  fold

./lib/pcg-data-normalization/src/Data/Normalization/Metadata/Internal.hs:73:38: Error: Parse error
Found:
      where
        f  NormalizedContinuousCharacter {}       = mempty
  >     f (NormalizedDiscreteCharacter   static ) = foldMap toList static
        f (NormalizedDynamicCharacter    dynamic) = foldMap (foldMap toList) dynamic
  

./lib/pcg-utility/src/Numeric/Cost.hs:163:21: Error: Parse error
Found:
          case hasCycle xs of
            Nothing                  -> let t@(_,r) = (v*10) `quotRem` den in go r $ xs |> t
  >         Just (static, repeating) -> render static <> addOverline (render repeating)
    
        addOverline = (<>[c]) . intersperse c
  

./lib/pcg-core/pcg-data-structures/src/Bio/Graph/PhylogeneticDAG/DynamicCharacterRerooting.hs:406:34: Error: Parse error
Found:
            deriveMinimalSequenceForDisplayTree = fmap recomputeCost . foldr1 (zipWith minimizeBlock) . fmap createZippableContext
              where
  >             minimizeBlock (static, dynCharVect1) (_, dynCharVect2) = (static, minimizedDynamicCharacterVector)
                  where
                    minimizedDynamicCharacterVector = zipWith minimizeDynamicCharacterRooting dynCharVect1 dynCharVect2
  

5 hints
```

[...]

PR #1524, opened 2023-06-13
Solves #704: Parse errors are counted as hints

thomas-huegel:
Solves #704: Parse errors are counted as hints


zliu41 (approved):
Looks great, thanks.

PR #1535, opened 2023-08-27
Fix false positive unused TH warning when typed expression splice used

jhrcek:
An attempt to fix  https://github.com/ndmitchell/hlint/issues/1531

PR #1536, opened 2023-08-27
[ghc-lib-parser-ex]: bump 9.6.0.2

shayne-fletcher:
references:
- https://github.com/ndmitchell/hlint/pull/1535
- https://github.com/shayne-fletcher/ghc-lib-parser-ex/pull/158

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -90,7 +90,7 @@ library
       build-depends:
           ghc-lib-parser == 9.6.*
     build-depends:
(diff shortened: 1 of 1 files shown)


hseg (commented):
Missing a corresponding bump in `stack.yaml` in the `extra-deps` clause.
Might be worth adding a clause testing that in CI - [a project I recently contributed to](https://github.com/lierdakil/pandoc-crossref/blob/master/.github/workflows/haskell.yml) checked nix builds on top of that, could be a good source of inspiration

PR #1537, opened 2023-08-28
Fix CI warning

jhrcek:
Fixing this warning that pops up in CI

Screenshot from 2023-08-28 14-57-51

More info [here](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)


ndmitchell:
Thanks!

PR #1538, opened 2023-08-31
Fix partial `comments` field accesses

Gabriella439:
I've been working on reviving the `splint` project for newer versions of GHC so that we can create a GHC plugin for `hlint`.

When running `hlint` as a plugin it seems to trip up on the `comments` field of `EpAnn`, which is not always present for the AST supplied to a GHC plugin.

This change fixes that by replacing all uses of the partial `comments` field with an alternative total function.


zliu41:
Thanks!

PR #1540, opened 2023-09-14
Correct Functor law hint

MatthijsBlom:
`<$>` associates to the left, so `f <$> g <$> x` is equivalent to `f . g <$> x` not because of a Functor law, but because for functions `(<$>) == (.)`.

The actual Functor law was missing, so I added it.

The original hint could perhaps be generalized to `f <$> g` &mapsto; `f . g`, but I could not figure out how to express that `g` should be a function.

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -472,11 +472,12 @@
     # FUNCTOR
 
     - warn: {lhs: fmap f (fmap g x), rhs: fmap (f . g) x, name: Functor law}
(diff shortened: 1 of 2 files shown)


googleson78:
> The original hint could perhaps be generalized to f <$> g ↦ f . g, but I could not figure out how to express that g should be a function.

You almost certainly can't in the general case, because `hlint` only does parsing, no typechecking.

ndmitchell:
Thanks!

PR #1543, opened 2023-10-06
Add hints for replacing some `maybe` usage with equivalent `traverse`

googleson78:


ndmitchell:
Why is this desirable? The `maybe` function is more targeted so helps enforce types better.

googleson78:
I personally find it desirable because the `maybe` usage is essentially reimplementing `traverse`, i.e. this is exactly the implementation of `traverse`.

I do agree that type inference is worse with this hint, and potentially (although I'm not sure when), this might cause ambiguity errors. Perhaps some optional group might be better here?

ndmitchell:
Definitely not in base, which in general doesn't do generalization. But further down there is a set of hints called generalised, where maybe to traverse would fit nicely. Happy to take the patch if adjusted to go there.

PR #1555, opened 2024-01-14
concat-counterparts of existing concatMap-rules

jvoigtlaender:
See https://github.com/ndmitchell/hlint/pull/1486#issuecomment-1891014081


ndmitchell:
Thanks!

PR #1565, opened 2024-02-04
generalize zip on empty lists hint

jvoigtlaender:
Will the `notEq x []` in the side condition work as intended, or is extra "escaping" of the list brackets necessary as in `lhs` and `rhs`?


googleson78:
What's `notEq`?

jvoigtlaender:
The hlint-internal side condition combinator that is already used at various other places in the same file.

ndmitchell:
The `notEq` looks fine. However, you have also used `isNonEmpty`. As far as I can tell, that hasn't been in hlint for many years. Yes, there was a use of it. But no, it didn't do anything, so I've deleted it. Can you remove the `isNonEmpty` uses.

jvoigtlaender:
Done.

PR #1571, opened 2024-02-05
Naming of hints

jvoigtlaender:
Suggested various renamings of hints, in ways I would consider more useful/consistent in communicating with students in particular.

The changes are distributed over several smaller commits "grouping" similar suggestions, and with brief justifications in the commit messages. To enable cherry-picking of individual suggestions/commits if need be.


ndmitchell:
Thanks - they all look improvements to me. Great to see the incremental improvements.

Issue #1575, opened 2024-02-27
hlint-3.8 fails to compile with ghc-9.8.1


peti:
I've run into the following build error:

~~~
[   35s] src/Hint/Extensions.hs:430:41: error: [GHC-39999]
[   35s]     • No instance for ‘Data (HsMatchContext RdrName)’
[   35s]         arising from a use of ‘hasS’
[   35s]     • In the second argument of ‘(||^)’, namely ‘hasS isStrictMatch’
[   35s]       In the expression: hasS isPBangPat ||^ hasS isStrictMatch
[   35s]       In an equation for ‘used’:
[   35s]           used BangPatterns = hasS isPBangPat ||^ hasS isStrictMatch
[   35s]     |
[   35s] 430 | used BangPatterns = hasS isPBangPat ||^ hasS isStrictMatch
[   35s]     |                                         ^^^^
~~~

The full log file is at https://build.opensuse.org/public/build/devel:languages:haskell:ghc-9.8.x/Tumbleweed/x86_64/hlint/_log.

Is this a known problem?

shayne-fletcher:
i have been working through some problems with `isStrictMatch`. it looks like i've made things worse. working on it.

shayne-fletcher:
the linked log doesn't exhibit the error?

shayne-fletcher:
my read is this build failure arose against ghc-lib-parser-ex 9.8.0.1 (the type of `isStrictMatch` in that version is wrong). there was like a 1 or 2 day window in which it was the prevailing version. that has since been fixed by the release of ghc-lib-parser-ex-9.8.0.2. i've verified hlint-3.8 against the latest packages from hackage builds and test ok with ghc-9.8.1. i think we're ok now. thanks for letting me know about the failure in any case.

peti:
> the linked log doesn't exhibit the error?

Indeed, the error has disappeared! It seems like an update to one of hlint's dependencies has fixed the issue. hlint itself hasn't changed -- it's the same version that used to fail 2 days ago. Very weird.

shayne-fletcher:
> > the linked log doesn't exhibit the error?
> 
> Indeed, the error has disappeared! It seems like an update to one of hlint's dependencies has fixed the issue. hlint itself hasn't changed -- it's the same version that used to fail 2 days ago. Very weird.

the hlint-3.8 cabal has bounds on ghc-lib-parser-ex like this:
```
    build-depends:
[...]

PR #1578, opened 2024-02-29
set lower bounds ghc-lib-parser-ex-9.8.0.2

shayne-fletcher:


shayne-fletcher:
related issue https://github.com/ndmitchell/hlint/issues/1575

PR #1591, opened 2024-05-08
WIP: Amoral language

### LINKED ISSUE #1584: Consider using less morally-loaded language in hints.
I know I stand a good chance of being ridiculed for this, so please hear me out, but feel free to ignore this issue entirely.

Hints are sometimes phrased as "Perhaps you should remove it." which seems to imply that "a person who does remove it is a better person".
https://github.com/ndmitchell/hlint/blob/6898b4322b1b3c010b9e61b72cad273f5ca3d650/src/Idea.hs#L82

This issue exists just to suggest replacing "should" by "could".

NorfairKing:
Fix #1584

I started by refactoring the relevant code, but in doing so I found some very strange semantics, as you'll see below.
I assume this is a historical artifact of previous refactors but I'd appreciate a double-check first.

I couldn't figure out how to run the tests either, `stack test` doesn't seem to run any tests.

### DIFF
--- a/hints.md
+++ b/hints.md
@@ -249,7 +249,7 @@ Found:
 <br>
 Suggestion:
 <code>
-Perhaps you should remove it.
+You may be able to remove it.
 </code>
 <br>
 </td>
@@ -1050,7 +1050,7 @@ do
 <br>
 Suggestion:
 <code>
-Perhaps you should remove it.
+You may be able to remove it.
 </code>
 <br>
 </td>
@@ -1319,7 +1319,7 @@ where
 <br>
 Suggestion:
 <code>
-Perhaps you should remove it.
+You may be able to remove it.
 </code>
 <br>
 Does not support refactoring.
--- a/src/Idea.hs
+++ b/src/Idea.hs
@@ -75,12 +75,16 @@ showIdeaANSI = showEx hsColourConsole
 showEx :: (String -> String) -> Idea -> String
 showEx tt Idea{..} = unlines $
     [showSrcSpan ideaSpan ++ ": " ++ (if ideaHint == "" then "" else show ideaSeverity ++ ": " ++ ideaHint)] ++
-    f "Found" (Just ideaFrom) ++ f "Perhaps" ideaTo ++
+    f ideaFrom ++ p ideaTo ++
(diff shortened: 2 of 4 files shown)


googleson78:
Regarding running tests - https://github.com/ndmitchell/hlint#how-to-run-tests

NorfairKing:
@googleson78 Aha nice! I just did that and it passed:
```
Tests passed (956)
```

But I'd like to do a bit more refactoring because the current code is definitely strange.

NorfairKing:
@ndmitchell what do you think?

ndmitchell:
I think no-loaded language is always better. I think "You can remove it" might be the clearest of all? At this point we believe it is definitely safe to remove, so I don't think we need to equivocate. But we also don't need to preach. Does `You can remove it` strike the right balance? Telling you what is safe, but leaving it up to you to take action?

NorfairKing:
> I think no-loaded language is always better.

That sentence is so funny to me :P

> I think "You can remove it" might be the clearest of all? At this point we believe it is definitely safe to remove, so I don't think we need to equivocate. But we also don't need to preach. Does `You can remove it` strike the right balance? Telling you what is safe, but leaving it up to you to take action?

I can make that change if you want.
I chose this wording because I've disagreed with hlint in the past and/or found bugs.

ndmitchell:
While anything in HLint can have bugs, I think we broadly have to assume it's right in the messaging, or it just gets way too confusing.

hints.md
```diff
@@ -249,7 +249,7 @@ Found:
 <br>
 Suggestion:
 <code>
-Perhaps you should remove it.
+You may be able to remove it.
```
ulidtko, line 252: Simplifying to `You can remove it` was my immediate thought too, FWIW

ulidtko (approved):

ndmitchell:
Reflecting on this further, I wonder if "Perhaps remove it" is the right balance? The perhaps turns it more into a dialog with the user - a suggestion. It doesn't suggest doubt in the validity of the hint, but it leaves all agency with the user?

Sorry for ratholing on the smallest things here @NorfairKing - genuinely curious what people think? I think this diff is definitely an improvement to the status quo, just wondering if more is even better?

PR #1600, opened 2024-05-27
evaluate sum on binary list

jvoigtlaender:
Seen in student code.

(Of course, similar evaluation rules like `sum [x] -> x` could also be added. But this is the one I have really seen.)


jvoigtlaender:
The new hint is also in the spirit of existing ones like these: https://github.com/ndmitchell/hlint/blob/752baba987f61d3d3778751f2709293521841e0c/data/hlint.yaml#L132-L133

ndmitchell:
Thanks!

PR #1603, opened 2024-06-22
Add hints to use comparing

josephcsible:


josephcsible:
This test is failing: https://github.com/ndmitchell/hlint/blob/e0f532a0b958e23044175706b611e45b5aa06889/data/hlint.yaml#L1581
It wants just this to fire: https://github.com/ndmitchell/hlint/blob/e0f532a0b958e23044175706b611e45b5aa06889/data/hlint.yaml#L117
But now "Use comparing" fires too. I'm not sure what the cleanest way to fix this is. What is `isCompare` doing in that rule exactly?

ndmitchell:
I think we actually had these hints in the past. But it seems not everyone is convinced by comparing - now the only instances are when you are already using comparing, or there is a significant perf advantage to doing so.

josephcsible:
Fair enough.

PR #1608, opened 2024-07-11
Add a hint to avoid capitalisms

bgohla:
I add a module `Hints.NoCapitalisms` that provides a hint to avoid capitalisms in identifiers. E.g. 
```haskell
module SSL.Foo                                                                                  
class FOO a where
class Foo a where getFOO
data Foo = Bar | BAAZ
getSMS
```
are discouraged.

The hint has severity `Ignored` so it is off by default.

* Things that might need attention
 - [ ] potentially factor out some code.
 - [ ] break into several commits.


src/Hint/NoCapitalisms.hs
```diff
@@ -0,0 +1,113 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
+{-
+    Detect uses of capitalisms
+
+    Only allow up to two consecutive capital letters in identifiers.
```
arybczak, line 6: What is the reason for this exception? `ID`? Personally I'd prefer uniformity here.

src/Hint/NoCapitalisms.hs
```diff
@@ -0,0 +1,113 @@
...
+hasUnderscore :: String -> Bool
+hasUnderscore = elem '_'
+
+hasCapitalism :: String -> Bool
+hasCapitalism s = any isAllUpper (trigrams s)
+  where
+    isAllUpper = all isUpper
+    trigrams = \case
+      a:b:c:as -> [a,b,c] : trigrams (c:as)
+      _otherwise -> []
```
tbidne, line 76: Shouldn't `b` also be part of the recursive call?

```diff
-a:b:c:as -> [a,b,c] : trigrams (c:as)
+a:b:c:as -> [a,b,c] : trigrams (b:c:as)
```

Otherwise you're missing triples that exist across the boundary e.g.

```haskell
hasCapitalism "getFOO" === False -- your example, probably not what you want?
hasCapitalism "geFOO" === True   -- should be the same, in any case, right?
```

src/Hint/NoCapitalisms.hs
bgohla: That would be one case.

I'm all for uniformity, by I'm afraid not allowing this exception could make this rule very annoying. Let me do some statistics on you internal code base…

src/Hint/NoCapitalisms.hs
cgohla: indeed. thanks for pointing that out.

it seems i hadn't written the test cases correctly.

src/Hint/NoCapitalisms.hs
bgohla: Actually, prohibiting two consecutive capitals would catch an identifier like `MVar`, I think that would be awkward.

I agree, we should favor the spelling `Id` over `ID`, but I can't think of a good rule here, that doesn't also catch, e.g., `getIData`.

Maybe we could tighten the proposed rule to prohibit two consecutive capitals unless followed by a lower case letter.

arybczak: I thought this PR only complains about local definitions, not all occurrences of such names :thinking: Doesn't it? If it does, then definitions from other packages are not a problem.

tbidne: You are correct, though I am thinking of interop with upstream definitions, where re-using a captialized name is arguably more readable than adhering to this rule. For instance, I imagine quite a few people (most?) would say `runInIO` is better than `runInIo`, even for **new** definitions, because of how familiar `IO` is.

src/Hint/NoCapitalisms.hs
```diff
@@ -0,0 +1,119 @@
...
+runCI
+foreign import ccall _FIREMISSLES :: IO ()
+getSMS :: IO () -- @Ignore
+gFOO = _ -- @Ignore
+geFOO = _ -- @Ignore
+getFOO = _ -- @Ignore
+</TEST>
+-}
+
+module Hint.NoCapitalisms(noCapitalismsHint) where
```
zliu41, line 43: "capitalism"? How about "NoCAPs"?

src/Hint/NoCapitalisms.hs
```diff
@@ -0,0 +1,119 @@
...
+import GHC.Hs.Extension
+import GHC.Hs
+import GHC.Types.SrcLoc
+
+import Language.Haskell.GhclibParserEx.GHC.Hs.Decls
+import Language.Haskell.GhclibParserEx.GHC.Utils.Outputable
+import GHC.Util
+
+noCapitalismsHint :: DeclHint
+noCapitalismsHint _ _ decl = [ remark Ignore "Avoid capitalisms" (reLoc (shorten decl))
```
zliu41, line 64: Add a Note to the hint, explaining what it does?

src/Hint/NoCapitalisms.hs
```diff
@@ -0,0 +1,119 @@
...
+    isAllUpper = all isUpper
+
+trigrams :: String -> [String]
+trigrams = \case
+  a:b:c:as -> [a,b,c] : trigrams (b:c:as)
+  _otherwise -> []
+
+--- these are copied from Hint.Naming ---
+
+shorten :: LHsDecl GhcPs -> LHsDecl GhcPs
```
zliu41, line 86: Please add some comments explaining what this does.

src/Hint/NoCapitalisms.hs
bgohla: > ```yaml
> # .hlint.yaml
> - ignore:
>     name: "No Capitalism"
>     within:
>       # I think this could work for functions.
>       - "Module.readTBQueue"
>       # Does this work for type / data declarations? Or does this assume
>       # module Another/TBQueue.hs?
>       - "Another.TBQueue"
> ```
> 
> This doesn't get you a global rule for ignoring e.g. `IO`, but maybe it's workable enough.

I don't understand the configuration language, and I have not been able to confirm that your suggestion works. OTOH, ANN pragmas might be more effective for defining exceptions.

bgohla: @arybczak I have become convinced that the stricter rule is feasible, i.e., prohibiting all consecutive capitals, exceptions have to be allowed using pragmas.

src/Hint/NoCapitalisms.hs
bgohla: done.

src/Hint/NoCapitalisms.hs
zliu41: I do not see such a meaning for capitalism in the dictionary.

src/Hint/NoCapitalisms.hs
zliu41: That's just a code comment. A Note is attached to a hint and is displayed with the hint. Alternatively, you can make the hint name more descriptive: "Avoid three consecutive capital letters" instead of "Avoid capitalisms".

src/Hint/NoCapitalisms.hs
arybczak: Yeah, "NoCAPs" seems better since it doesn't have political connotations and shouldn't provoke unnecessary discussions, unlike the current name.

PR #1609, opened 2024-08-03
remove stack.yaml

shayne-fletcher:
i no longer have a use for this and propose to remove it.


hseg:
I hadn't seen this -- is it such a burden to maintain `stack.yaml`?
Not that moving over to building with `cabal` is that difficult, but I was surprised.
Though one good thing about this shakeup is that it has shown me that `stack test` never worked here -- the correct way to run tests apparently being `cabal run -- hlint --test` ? Is there any reason these tests are not packaged up in a standard test suite?

Issue #396, opened 2017-10-22
Add a test suite entry


ndmitchell:
As noted by @contivero, stack test should do a test. Probably just add an extra Test.hs which calls the existing test suite.

hseg:
Started working on this, am puzzled by some of the code structure.
For example, judging by its imports, the module `Language.Haskell.HLint` seems to mainly serve to reexport `HLint.hlint`, as well as serving as a kitchensink export for interactive use?

PR #1614, opened 2024-09-04
Add bare-bones test suite

hseg:
We probably want to be more clever and replace the `--test` invocation
with a `detailed-0.9` testsuite. We probably also should integrate the
QuickCheck tests into the testsuite.
That would take a deeper reorganization than I'm willing to commit to
now, though, so I'll leave it for another day.

Closes: #396

### DIFF
--- a/hlint.cabal
+++ b/hlint.cabal
@@ -62,7 +62,7 @@ flag hsyaml
   manual: True
   description: Use HsYAML instead of yaml
 
-library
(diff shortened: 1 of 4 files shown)


hseg:
No idea why the build failed, it works for me (barring the installing `pply_refact` step)

hseg:
Not clear to me why the CI is failing, would appreciate guidance

PR #1617, opened 2024-10-16
Add hints for when 'x <*> y' could be 'y'

ysangkok:
I haven't added all the alternatives with `mempty` and `empty` as hlint doesn't
know about types, and there could be lawless instances out there I suppose. So
it's safer just to deal with concrete types.

```
ghci> :{
ghci| let f :: Applicative m => m (b -> (Int, b))
ghci|     f = (,) <$> pure 5
ghci| :}
ghci> f <*> mempty :: Ap [] (Int, Sum Int)
Ap {getAp = [(5,Sum {getSum = 0})]}
ghci> f <*> empty :: Ap [] (Int, Sum Int)
Ap {getAp = []}
```

Would be nice to detect the last one, but hlint can't know that it's `empty` on `Ap [] (..., Sum ...)`.


ndmitchell:
Thanks!

PR #1618, opened 2024-10-24
Add support for NumericUnderscores extensions from CLI/config

9999years:
Closes https://github.com/ndmitchell/hlint/issues/1434

Supercedes #1435, updated for latest `master`.

### DIFF
--- a/src/GHC/All.hs
+++ b/src/GHC/All.hs
@@ -6,7 +6,9 @@ module GHC.All(
     CppFlags(..), ParseFlags(..), defaultParseFlags,
     parseFlagsAddFixities, parseFlagsSetLanguage,
     ParseError(..), ModuleEx(..),
-    parseModuleEx, createModuleEx, createModuleExWithFixities, ghcComments, modComments, firstDeclComments,
+    parseModuleEx, createModuleEx, createModuleExWithFixities,
+    createModuleExWithFixitiesAndExtensions, ghcComments, modComments,
+    firstDeclComments,
     parseExpGhcWithMode, parseImportDeclGhcWithMode, parseDeclGhcWithMode,
     ) where
 
@@ -86,8 +88,9 @@ data ParseError = ParseError
     }
 
 -- | Result of 'parseModuleEx', representing a parsed module.
-newtype ModuleEx = ModuleEx {
-    ghcModule :: Located (HsModule GhcPs)
+data ModuleEx = ModuleEx {
+    ghcModule :: Located (HsModule GhcPs),
+    configuredExtensions :: [Extension]
 }
 
 -- | Extract a complete list of all the comments in a module.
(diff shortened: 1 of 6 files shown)


src/Hint/NumLiteral.hs
```diff
@@ -43,10 +45,16 @@ numLiteralHint _ modu =
...
+  -- TODO: there's a subtle bug when the module disables `NumericUnderscores`.
+  -- This seems pathological, though, because who would enable it for their
+  -- project but disable it in specific files?
   if NumericUnderscores `elem` exts then
      concatMap suggestUnderscore . universeBi
   else
      const []
+  where
+    moduleExtensions = union (extensions (modComments modu)) (extensions (firstDeclComments modu))
+    activeExtensions = configuredExtensions modu <> toList moduleExtensions
```
zliu41, line 57: `activeExtensions` doesn't seem to be used anywhere?

9999years: Great catch, thanks! One of the rebases must've broken this (I've fixed conflicts in this PR many times since it was originally opened in late 2022).

zliu41 (commented):
I'll defer this to @ndmitchell. I think the intention of passing extensions to HLint is for parsing, not for deciding if a hint should trigger, although it seems reasonable to do the latter.

Even if we decide to do that, this PR will need a few more updates:
- I'm not seeing how you are passing the additional extensions from the Yaml config to the places that need them.
- Since the additional extensions are from the HLint config and not tied to individual modules, it doesn't seem to make sense to associate them with `ModuleEx`.
- Need to add some tests.

src/GHC/All.hs
```diff
@@ -214,7 +223,7 @@ parseModuleEx flags file str = timedIO "Parse" file $ runExceptT $ do
         ExceptT $ parseFailureErr dynFlags str file str $ NE.fromList errs
       else do
         let fixes = fixitiesFromModule a ++ ghcFixitiesFromParseFlags flags
-        pure $ ModuleEx (applyFixities fixes a)
+        pure $ ModuleEx (applyFixities fixes a) (enabledExtensions flags)
```
parsonsmatt, line 226: So the provenance of these extensions is in `ParseFlags`. This is not passed to hints themselves. 

The type of `DeclHint` is:

```haskell
type DeclHint = Scope -> ModuleEx -> LHsDecl GhcPs -> [Idea]
```

If we want to propagate this information, we have essentially three choices:

1. Smuggle it in `ModuleEx`.
2. Smuggle it in `Scope`.
3. Add another function argument here for the whole `hlint` configuration (`ParseFlags` ? and `[Setting]` maybe?)

`Scope` is, already, just a convenience/performance/cache wrapper around information that is extracted directly from the contents of a `ModuleEx`. [Here is definition and logic to create](https://github.com/parsonsmatt/hlint/blob/71cf6e573070e22efa0555590af02b66a8cf8629/src/GHC/Util/Scope.hs?plain=1#L25-L55).

I think one could argue that `Scope` is a more reasonable place for this than `ModuleEx`, but I'd go further: `Scope` should be part of `ModuleEx` directly. After all, `scopeCreate` is called on the `ModuleEx` - which  means that we have already lost the `ParseFlags` information, if it is not present on `ModuleEx`.

IMO, a `ModuleEx` represents "what we know about a parsed module":

```haskell
-- | Result of 'parseModuleEx', representing a parsed module.
newtype ModuleEx = ModuleEx {
    ghcModule :: Located (HsModule GhcPs)
}
```

But we *don't* know - what extensions were enabled while parsing this module? 

I think "rearchitecting the way we pass flags around in `hlint` and breaking the whole API for hints" is probably not appropriate for this PR. If we extend `ModuleEx`, then we pave the way for further additions to `ModuleEx` as "the" place to put information relevant to a module. If we extend `Scope`, then we need to plumb that through anyway, possibly through `ModuleEx` resulting in duplication. But neither option prevent or impede a more serious refactoring in how this code works.

Issue #1604, opened 2024-06-26
Setting Language Disables Other Extensions


Burtannia:
While configuring my `.hlint.yaml` file for my Yesod project I ran into a strange issue. I decided to dump all my `default-extensions` as arguments into hlint just to ensure correct parsing. I'm using GHC2021 so I also wanted to pass `-XGHC2021`. The minimal example I can give is using template haskell:
```
- arguments: [
  --color,
  --cpp-simple,
  -XGHC2021,
  -XTemplateHaskell,
]
```

This should work just fine to my understanding, in fact `-XTemplateHaskell` is probably redundant because I believe hlint automatically enables it. Either way this produces warnings all over the place of the form:
```
src/Handler/Profile.hs:12:10-31: Suggestion: Redundant bracket
Found:
  do setTitle . toHtml $ userIdent user <> "'s User page"
    $ (widgetFile "profile")
Perhaps:
  do setTitle . toHtml $ userIdent user <> "'s User page"
    $ widgetFile "profile"
```

Removing `-XGHC2021` from the arguments list fixes the issue and the warnings no longer show. My guess is that selecting the language version ignores any other extensions. The issue is also present using `-XHaskell2010`.

As an aside to this, it would be nice if hlint could automatically detect and use the `language` version / `default-extensions` from `package.yaml` to avoid the need for these arguments entirely. This seems to happen via HLS in VSCode but does not happen when running `hlint` manually via the command line.

LiamGoodacre:
Just ran into this too, specifically with GHC2021 & TemplateHaskell; (hlint 3.6.1)

LiamGoodacre:
> to get the GHC2021 extensions and TemplateHaskell then -XGHC201 -XTemplateHaskell should do it

This doesn't appear to work (3.6.1) neither as command line arguments nor arguments in the yaml file.

`Example.hs`
```hs
module Example where

example = $(something)
```

### With command args

```bash
∀ hlint -XGHC2021 -XTemplateHaskell Example.hs
Example.hs:3:11: Error: Parse error: on input `$'
Found:
    module Example where

  > example = $(something)


1 error
```

### With config file

`example.hlint.yaml`
```yaml
- arguments:
    - -XGHC2021
    - -XTemplateHaskell
```

```bash
∀ hlint --hint=example.hlint.yaml Example.hs
Example.hs:3:11: Error: Parse error: on input `$'
Found:
    module Example where

  > example = $(something)


1 error
```

[...]

PR #1619, opened 2024-11-01
enable implied extensions

shayne-fletcher:
this is an attempt at a fix for https://github.com/ndmitchell/hlint/issues/1604. the idea is to make use of [`extensionImplications`](https://hackage.haskell.org/package/ghc-lib-parser-ex-9.10.0.0/docs/Language-Haskell-GhclibParserEx-GHC-Driver-Session.html#v:extensionImplications)

T.hs
```haskell
import Language.Haskell.TH

oneC, twoC, plusC  :: Quote m => m Exp
oneC = [| 1 |]
twoC = [| 2 |]
plusC = [| $oneC + $twoC |]

main = pure ()
````

T.hs compiled with `cabal run exe:hlint -- -XGHC2021  -XTemplateHaskell ~/RRR.hs` fails for lack of `HaskellTemplateQuotes` before this fix and succeeds after.

### DIFF
--- a/src/GHC/All.hs
+++ b/src/GHC/All.hs
@@ -15,6 +15,7 @@ import Control.Monad.Trans.Except
 import Control.Monad.IO.Class
 import Util
 import Data.Char
+import Data.List
 import Data.List.NonEmpty qualified as NE
 import Data.List.Extra
 import Timing
@@ -39,6 +40,8 @@ import Data.Generics.Uniplate.DataOnly
 
 import Language.Haskell.GhclibParserEx.GHC.Parser
 import Language.Haskell.GhclibParserEx.Fixity
+import Language.Haskell.GhclibParserEx.GHC.Driver.Session
+
 import GHC.Util
 
 -- | What C pre processor should be used.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/GHC/All.hs [hunk 1]
```diff
@@ -131,7 +134,7 @@ ghcFixitiesFromParseFlags = map toFixity . fixities
 
 parseModeToFlags :: ParseFlags -> DynFlags
 parseModeToFlags parseMode =
-    flip lang_set (baseLanguage parseMode) $ foldl' xopt_unset (foldl' xopt_set baseDynFlags enable) disable
+  flip lang_set (baseLanguage parseMode) $ foldl xopt_unset (foldl' xopt_set baseDynFlags enable) disable
```
zliu41, line 137: Is the change from `foldl'` to foldl` intentional?

zliu41 (approved):
There seems to be some overlap with parts of `src/Extension.hs`.

src/GHC/All.hs
shayne-fletcher: no! good catch. thanks!

PR #1620, opened 2024-11-03
more hints about eliminating Data.Foldable.toList

jvoigtlaender:
Essentially a continuation of the already merged https://github.com/ndmitchell/hlint/pull/1569.

One thing I'm unsure about: Would these hints also catch/apply to `Data.Set.toList` calls (based on the `Foldable` instance for `Set`)? Or would it be necessary to duplicate all these hints (the ones from the previous issue and these ones here) for `Data.Set.toList`?

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -936,6 +936,16 @@
     - warn: {lhs: find f (Data.Foldable.toList x), rhs: find f x}
     - warn: {lhs: concat (Data.Foldable.toList x), rhs: concat x}
     - warn: {lhs: concatMap f (Data.Foldable.toList x), rhs: concatMap f x}
+    - warn: {lhs: foldrM f c (Data.Foldable.toList x), rhs: foldrM f c x}
+    - warn: {lhs: foldlM f c (Data.Foldable.toList x), rhs: foldlM f c x}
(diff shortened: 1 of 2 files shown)


zliu41 (changes requested):
Looks good, but please update `hints.md`. See README for instructions.

>  Would these hints also catch/apply to Data.Set.toList calls

No - HLint does not know `Data.Set.toList` and `Data.Foldable.toList` are the same `toList`.

jvoigtlaender:
I have updated `hints.md`.

About this:
> No - HLint does not know `Data.Set.toList` and `Data.Foldable.toList` are the same `toList`.

Would it then make sense to duplicate all the hints from https://github.com/ndmitchell/hlint/pull/1569 as well as the ones from this PR here, into versions for `Data.Set.toList` in place of `Data.Foldable.toList`?

zliu41 (approved):
Apologies for the delay - I lost track of it.

> Would it then make sense to duplicate all the hints from https://github.com/ndmitchell/hlint/pull/1569 as well as the ones from this PR here, into versions for Data.Set.toList in place of Data.Foldable.toList?

That seems reasonable to me - Data.Set is widely used, so it’s worth providing some hints for it.

PR #1634, opened 2025-02-14
Add brackets if needed for hint "Avoid lambda"

### LINKED ISSUE #1509: `-XBlockArguments`: Avoid Lambda incorrect
- avoid lambda with `BlockArguments` doesn't insert parens resulting in an error when applying the hint
- if BlockArguments are on, it should be checkd if the lambda is in parens and if that's not the case, automatically surround the eta reduced expressions with parens.

zliu41:
Fixes #1509.

The fix is in the `go [] e` case in `niceLambdaR`.

### DIFF
--- a/src/GHC/Util/HsExpr.hs
+++ b/src/GHC/Util/HsExpr.hs
@@ -141,7 +141,7 @@ niceDotApp a b = dotApp a b
 
 -- Generate a lambda expression but prettier if possible.
 niceLambda :: [String] -> LHsExpr GhcPs -> LHsExpr GhcPs
-niceLambda ss e = fst (niceLambdaR ss e)-- We don't support refactorings yet.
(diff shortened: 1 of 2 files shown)


shayne-fletcher (approved):

andreasabel:
Thanks @zliu41 !

ysangkok:
Hi @zliu41 , excuse me for the late review, but it seems there is a problem with this PR: see https://github.com/ndmitchell/hlint/pull/1642#issuecomment-2817319079 . Do you want me to submit a new issue?

ysangkok:
Reported #1645 to track the issue introduced here.

zliu41:
@ysangkok Sorry it took a while - here's the fix: https://github.com/ndmitchell/hlint/pull/1646

Issue #1539, opened 2023-09-06
Refactor flag can only be used with an individual file


chris-martin:
```
> hlint -j . -o "..." --refactor --refactor-options="--inplace"
hlint: Refactor flag can only be used with an individual file
CallStack (from HasCallStack):
  errorIO, called at src/HLint.hs:213:14 in hlint-3.4.1-2AxkzXrwpEoEYKlPa1ZKgm:HLint
```

Why does this limitation exist? Would it be possible to allow refactoring multiple files at a time? It would be really great to be able to automatically apply a replacement rule throughout an entire project.

As a workaround I can do something like this: `fish> for x in (git ls-files '**/*.hs'); hlint -o "..." --refactor --refactor-options="--inplace" $x; end` but I don't understand why it's necessary.

ndmitchell:
I have no idea. Do you @zliu41 ? I think it should be as simple as changing https://github.com/ndmitchell/hlint/blob/14162e89c2fbf5c42a451e4e2368a8b11625f5cb/src/HLint.hs#L212C1-L223 to be:

```
forM_ cmdFiles $ \file -> do
    ...
```

If someone can test it out, and it actually works, I'd be happy to take a PR.

zliu41:
This was introduced in 2016 (https://github.com/ndmitchell/hlint/pull/274) and no reason was given in the PR. It certainly seems douable, though it may require a bit more than `forM_ cmdFiles $ \file -> do`, since it may need to find the `Idea`s corresponding to each file. I agree this is a good first issue so I'll leave it for someone else to fix.

PR #1635, opened 2025-02-14
Allow inplace refactoring of multiple files

alex-mckenna:
Allow multiple files to be refactored at a time, applying ideas to each relevant file. I was bitten by this the other week at work while adding some [`pre-commit`](https://pre-commit.com/) hooks to a repository, so if it's convenient a release after approval would be appreciated :slightly_smiling_face: 

Closes https://github.com/ndmitchell/hlint/issues/1539.


zliu41 (commented):
Looks good. Can you update the README? It currently says

> Because hlint's --refactor option only works when you pass a single file, ...

zliu41 (approved):
Thanks! The release can only be done by @ndmitchell

Issue #1640, opened 2025-04-02
`hlint-3.10-x86_64-osx.tar.gz` is actually for ARM64, not x86_64


RyanGlScott:
The [`hlint-3.10-x86_64-osx.tar.gz`](https://github.com/ndmitchell/hlint/releases/download/v3.10/hlint-3.10-x86_64-osx.tar.gz) binary offered on the Releases page claims that it is for the x86_64 architecture, but it is actually for ARM64 instead:

```
$ tar -xvf hlint-3.10-x86_64-osx.tar.gz
$ file hlint-3.10/hlint
hlint-3.10/hlint: Mach-O 64-bit executable arm64
```

This is a consequence of the fact that the macOS binaries are built in CI using a [`macos-latest`](https://github.com/ndmitchell/hlint/blob/cf89a66595e9c2c8082bee545ebffa1b08097dca/.github/workflows/release.yml#L38-L39) job, which now uses Apple silicon-based CI runners instead of Intel-based ones. I think we should rename the bindist to something like `hlint-3.10-arm64-osx.tar.gz` to avoid confusion.

RyanGlScott:
I've also opened https://github.com/ndmitchell/hlint/pull/1641 with a CI change that will ensure future bindists are labeled correctly.

RyanGlScott:
#1641 was landed. All that remains is to rename the `hlint-3.10-x86_64-osx.tar.gz` binary on the [Releases page](https://github.com/ndmitchell/hlint/releases/tag/v3.10) to `hlint-3.10-arm64-osx.tar.gz`.

PR #1641, opened 2025-04-02
CI: Correctly label macOS bindist as ARM64

RyanGlScott:
This ensures that future macOS bindists will not be inaccurately labeled as x86_64. This should avoid mixups like the one observed in #1640.


zliu41 (approved):
Thanks!

PR #1667, opened 2025-12-04
Add `typeApplications` option for restricted functions

### LINKED ISSUE #1637: Make it possible to forbid "`fromIntegral` without visible type arguments"
`fromIntegral @Int16 @Int32` is refactor safe but `fromIntegral` is not.
I want to allow the former but forbid the latter.

terror:
Resolves: https://github.com/ndmitchell/hlint/issues/1637

This diff adds a `typeApplications` option to function restrictions, allowing users to require or forbid visible type arguments on specific functions. The primary use case is enforcing refactor-safe usage of polymorphic functions like `fromIntegral` and `show`, where implicit type inference can silently produce incorrect behaviour after type changes.

E.g., you're now able to specify:

```yaml
- functions:
  - {name: fromIntegral, typeApplications: required}
  - {name: show, typeApplications: required}
```

Where `fromIntegral x` will be flagged with "Use visible type application", while `fromIntegral @Int @Int64 x` will be accepted. The option also supports forbidden for cases where type applications should be disallowed.

### DIFF
--- a/src/Config/Yaml.hs
+++ b/src/Config/Yaml.hs
@@ -343,7 +343,19 @@ parseRestrict restrictType v = do
         Just def -> do
             b <- parseBool def
             allowFields v ["default"]
-            pure $ Restrict restrictType b [] mempty mempty mempty mempty [] NoRestrictIdents Nothing
+            pure Restrict
+              { restrictType = restrictType
+              , restrictDefault = b
+              , restrictName = []
+              , restrictAs = mempty
+              , restrictAsRequired = mempty
+              , restrictImportStyle = mempty
+              , restrictQualifiedStyle = mempty
+              , restrictTypeApp = Nothing
+              , restrictWithin = []
+              , restrictIdents = NoRestrictIdents
+              , restrictMessage = Nothing
+              }
         Nothing -> do
             restrictName <- parseFieldOpt "name" v >>= maybe (pure []) parseArrayString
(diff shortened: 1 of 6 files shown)


src/Config/Yaml.hs
```diff
@@ -343,7 +343,19 @@ parseRestrict restrictType v = do
...
+              , restrictName = []
+              , restrictAs = mempty
+              , restrictAsRequired = mempty
+              , restrictImportStyle = mempty
+              , restrictQualifiedStyle = mempty
+              , restrictTypeApp = Nothing
+              , restrictWithin = []
+              , restrictIdents = NoRestrictIdents
+              , restrictMessage = Nothing
+              }
```
terror: Aside from adding `restrictTypeApp`, this is strictly a style refactor to make it more apparent as to what's going on. If this doesn't align with how we proceed elsewhere I'm happy to revert!

src/Config/Yaml.hs
```diff
@@ -361,6 +373,10 @@ parseRestrict restrictType v = do
               , ("post"        , QualifiedStylePost)
               , ("unrestricted", QualifiedStyleUnrestricted)
               ]
+            restrictTypeApp <- parseFieldOpt "typeApplications" v >>= maybeParseEnum
+              [ ("required" , TypeAppRequired)
+              , ("forbidden", TypeAppForbidden)
```
terror: I'm questioning the need for a `forbidden` variant here, I don't see a strong use case for it in the original issue 🤔

terror:
Closing as stale :/

PR #1687, opened 2026-04-19
Fix: qualify Data.List.map in generalise group (fixes #1572)

### LINKED ISSUE #1572: hlint incorrectly suggests renaming shadowed "map" when generalise group is on
Given 

```yaml
- group: {name: generalise, enabled: true}
```

hlint will suggest the following illegal transformation:

```diff
- test map = map
+ test map = fmap
```

The fix is to fully qualify the following:

https://github.com/ndmitchell/hlint/blob/8bd116cc7a3fcabf309cb9c1b9aa35016c190b09/data/hlint.yaml#L1125

to 

```yaml
    - warn: {lhs: Data.List.map, rhs: fmap}
```

mgajda:
## Summary

Fixes #1572. The `generalise` group rule `map -> fmap` triggers even when
`map` is a locally-bound parameter (e.g. `test map = map`). The resulting
refactoring `test map = fmap` does not compile.

Qualifying the LHS as `Data.List.map` restricts the match to the actual
library function, leaving shadowed locals alone.

## Test plan
- [ ] `test map = map` no longer triggers the suggestion under `--hint=generalise`
- [ ] `test xs = map f xs` still triggers under `--hint=generalise`

### DIFF
--- a/data/hlint.yaml
+++ b/data/hlint.yaml
@@ -1167,7 +1167,7 @@
     imports:
     - package base
     rules:
-    - warn: {lhs: map, rhs: fmap}
+    - warn: {lhs: Data.List.map, rhs: fmap}
     - warn: {lhs: a ++ b, rhs: a <> b}
     - warn: {lhs: "sequence [a]", rhs: "pure <$> a"}
     - warn: {lhs: "x /= []", rhs: not (null x), name: Use null}


mgajda:
Closing this — the qualification is too strict. Testing revealed that `lhs: Data.List.map` only fires when `Data.List` is explicitly imported, so Prelude's `map` (the common case) no longer triggers the `map -> fmap` suggestion. The regression is worse than the original bug. A proper fix for #1572 likely requires hlint's matcher to avoid matching library rules against locally-bound identifiers, which is out of scope for a YAML-only change.

Issue #1701, opened 2026-07-20
Allow checking for Map.Map occurrences


NorfairKing:
I've seen a lot of nonsense type occurrences like `Text.Text`, `Map.Map`, or `Set.Set` instead of just a second `import Data.Map (Map)`. I'd like to let `hlint` help me check for this automatically.

PR #1702, opened 2026-07-20
#1701, add a "Redundant module qualifier" hint

NorfairKing:
Implements #1701.

Type occurrences like `Map.Map`, `Set.Set` or `Text.Text` stutter. A module may be imported twice, so the qualified import can stay and a second one brings just the type into scope unqualified:

```haskell
import Data.Map (Map)
import Data.Map qualified as Map
```

```
Foo.hs:5:15-21: Suggestion: Redundant module qualifier
Found:
  Map.Map
Perhaps:
  Map
Note: requires importing Data.Map (Map)
```

### Scope

Only a one-component qualifier repeating the type name, so `Map.Map` but not `Data.Map.Map`: per @zliu41's review, a full module name is how you tell `Data.Map`'s `Map` from `Foo.Map`'s.

For the same reason it stays quiet when an import already binds the name unqualified to a different module, since dropping the qualifier there is an ambiguous occurrence. Detecting that needs certainty rather than a guess, hence `importedUnqualified` in `GHC.Util.Scope`, which counts only imports with an explicit list naming the type.

Fires in signatures, `data`/`newtype` declarations, type synonyms, instance heads and expression type annotations. Value-level occurrences are untouched. Severity is `Suggestion`, since whether the extra import is worth it is a matter of taste.

The note names the module when the qualifier resolves to one, falls back to "requires importing Map unqualified" when ambiguous, and says "already in scope unqualified" when only the qualifier needs deleting. It deliberately doesn't say whether the new import joins the qualified one or replaces it: that depends on whether the qualifier is used elsewhere.

### No refactoring

Dropping the qualifier only compiles once the type is in scope unqualified, which HLint can't arrange, so the hint offers no `Refactoring` rather than letting `--refactor` produce code that doesn't build.

### What it finds in HLint itself

20 occurrences across 6 files, deliberately not changed here — happy to add a follow-up commit applying them. CI is unaffected, since the `ndmitchell/neil` step lints with a released HLint.

| File | Count |
| --- | --- |
| `src/Hint/Restrict.hs` | 10 |
| `src/Hint/Extensions.hs` | 5 |
| `src/Hint/Duplicate.hs` | 2 |
| `src/GHC/Util/ApiAnnotation.hs` | 1 |
| `src/Hint/Naming.hs` | 1 |
| `src/Hint/Smell.hs` | 1 |

### Testing

`hlint --test` passes. The `<TEST>` block covers positives, all three note forms, and negatives: `Data.Map.Map`, `Data.Map.Strict.Map`, `Map.Key`, a non-stuttering alias, a load-bearing alias, value-level `Map.Map`, and the import declaration itself.

`hints.md` regenerated.

---
Written by Claude, reviewed and signed-off on by @NorfairKing

### DIFF
--- a/src/Hint/Stutter.hs
+++ b/src/Hint/Stutter.hs
@@ -20,6 +20,9 @@ instance C Text.Text -- Text
 foo = undefined :: Map.Map k v -- Map
 import qualified Data.Map as Map \
 foo :: Map.Map k v -- @Note requires importing Data.Map (Map)
+import qualified Data.Map as Map \
+import qualified Data.IntMap as Map \
+foo :: Map.Map k v -- @Note requires importing Map unqualified
 import qualified Data.Map as M \
 foo :: M.Map k v
 foo :: Map k v
--- a/src/Hint/Stutter.hs
+++ b/src/Hint/Stutter.hs
@@ -1,0 +1,77 @@
+{-# LANGUAGE ImportQualifiedPost #-}
+
+{-
+    Find type occurrences whose module qualifier repeats the type name,
+    such as 'Map.Map', 'Set.Set' or 'Text.Text'. A module may be imported
+    twice, so the qualified import can stay and a second import can bring
+    just the type into scope unqualified:
+
+    > import Data.Map (Map)
+    > import Data.Map qualified as Map
+
+<TEST>
+foo :: Map.Map k v -- @Suggestion Map
+foo :: Data.Map.Map k v -- Map
+type Foo = Set.Set Int -- Set
+data Foo = Foo Text.Text -- Text
+newtype Foo = Foo {unFoo :: Seq.Seq Int} -- Seq
+instance C Text.Text -- Text
+foo = undefined :: Map.Map k v -- Map
+import qualified Data.Map as Map \
+foo :: Map.Map k v -- @Note requires importing Data.Map (Map)
+import qualified Data.Map as M \
+foo :: M.Map k v
+foo :: Map k v
+foo :: Map.Key k
+foo :: Data.Map.Strict.Map k v
+foo = Map.Map
+foo = Map.empty
+import Data.Map as Map
+</TEST>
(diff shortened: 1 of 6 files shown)


src/Hint/Stutter.hs
```diff
@@ -0,0 +1,77 @@
...
+        [Note $ importNote scope lname occ]
+        []
+    | L l ty@(HsTyVar x promoted lname@(L nameLoc name)) <- universeBi decl :: [LHsType GhcPs]
+    , stutters name
+    , let occ = rdrNameOcc name
+    , let unqualified = HsTyVar x promoted (L nameLoc (mkRdrUnqual occ)) :: HsType GhcPs
+    ]
+
+-- | Does the qualifier's final component repeat the name it qualifies, as in
+-- @Map.Map@ or @Data.Map.Map@?
```
zliu41: `Map.Map` may be unpleasant, but `Data.Map.Map` is reasonable to me. Suppose I use two different Maps in the same module, one from `Data.Map` and one from `Foo.Map`. Qualifying them with the full module names seems sensible.

NorfairKing: Agreed, narrowed to a one-component qualifier: `Map.Map` fires, `Data.Map.Map` doesn't. Your two-Maps case is a negative test now.

The point applies a level down too, at the alias. With `import qualified Data.Map as Map` and `import Foo (Map)`, `Map.Map` was flagged and the note told you to add an import that would be an ambiguous occurrence. It now stays quiet when an import already binds the name unqualified to a different module. Detecting that needs certainty rather than a guess, so there's a small `importedUnqualified` in `GHC.Util.Scope` that counts only imports with an explicit list naming the type.

Self-sweep on HLint's own source is 20 hits across 6 files now, down from 31.

NorfairKing:
Gentle ping

zliu41 (approved):

